PR #2, opened 2021-07-16
Adds design doc for bundle CRD

JoshVanL:
Signed-off-by: joshvanl <vleeuwenjoshua@gmail.com>

### DIFF
--- a/design/20210715.bundle-crd.md
+++ b/design/20210715.bundle-crd.md
@@ -107,6 +107,8 @@ expressive enough to meet most needs.
   [upstream](https://github.com/kubernetes/kubernetes/issues/63726) the best
   solution for trust distribution in Kubernetes.
 - Enable _most_ users to effectively distribute trust data in their clusters.
+- Bundle should provide _both_ clients and servers with the trust data they need
+    to communicate.
 
 
 ### Non-Goals
--- a/design/20210715.bundle-crd.md
+++ b/design/20210715.bundle-crd.md
@@ -1,0 +1,272 @@
+---
+title: Bundle CRD
+authors:
+  - "@joshvanl"
+editor: "@joshvanl"
+creation-date: 2021-07-15
+last-updated: 2021-07-15
+status: implemented
+---
+
+# Bundle CRD
+
+## Table of Contents
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+- [Goals](#goals)
+  * [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  * [Bundle](#bundle)
+    + [API](#api)
+    + [Custom Cache](#custom-cache)
+    + [Upstream CertificateBundle](#upstream-certificatebundle)
+  * [Testing](#testing)
+  * [Risks and Mitigations](#risks-and-mitigations)
+  * [Graduation](#graduation)
+      - [Alpha](#alpha)
+      - [Beta](#beta)
+      - [GA](#ga)
+<!-- /toc -->
+
+
+## Summary
+
+[Bundle](https://github.com/cert-manager/trust/blob/44d8fcaeec7232daf6dd14a1be5a6d0791bc23b8/pkg/apis/trust/v1alpha1/types_bundle.go)
+is a cluster scoped CRD designed to distribute trust data across a cluster so
+that it is made available by all services that need it. Bundle is intended to be
+complimentary to [cert-manager](https://github.com/jetstack/cert-manager) by
+allowing access to trust data that is both known to cert-manager, and external
+to any cert-manager [Issuer](https://cert-manager.io/docs/configuration/).
+
+Bundle sources its data from a trusted namespace. This trusted namespace is the
+same namespace that the operator is deployed to, which is typically
+`cert-manager`. The sourced data is then appended together, and propagated
+across the cluster to a target, making it available for consumers.
+
+Bundle is intended to be powerful and flexible enough to meet the use case of
+most users who want some set of trust bundle data available for their services.
+This trust bundle data may or may not be known to the cluster operator before
+runtime.
+
+## Motivation
+
+Kubernetes currently has a limited story for distributing trust within a
+cluster. While API server CA bundle data is available to all pods at a
+well-known
+[file location](https://kubernetes.io/docs/tasks/run-application/access-api-from-pod/#directly-accessing-the-rest-api)
+mounted by the Kubelet at runtime, as well as the
+[ConfigMap in all namespaces `kube-root-ca.crt`](https://github.com/kubernetes/kubernetes/pull/100926/files#diff-60eb175dae1dcc1320f5c426f6d193a67e961e2bd96d574f941cdf09672090f6R44),
+this bundle data can only be modified via
+[API server CLI flags](https://kubernetes.io/docs/setup/best-practices/certificates/).
+Modifying these CLI flags to append arbitrary trust bundles is both impossible
+for managed Kubernetes users, and is a misuse of its intention. CLI flags are
+static, and can't be realistically changed in response to runtime events.
+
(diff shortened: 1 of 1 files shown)


design/20210715.bundle-crd.md
```diff
@@ -0,0 +1,275 @@
...
+
+
+## Summary
+
+[Bundle](https://github.com/cert-manager/trust/blob/44d8fcaeec7232daf6dd14a1be5a6d0791bc23b8/pkg/apis/trust/v1alpha1/types_bundle.go)
+is a cluster scoped CRD designed to distribute trust data across a cluster so
+that it is made available by all services that need it. Bundle is intended to be
+complimentary to [cert-manager](https://github.com/jetstack/cert-manager) by
+allowing access to trust data that is both known to cert-manager, and external
+to any cert-manager [Issuer](https://cert-manager.io/docs/configuration/).
```
SgtCoDFish, line 41: nitpick: I think this reads like trust data needs to be both known and external to cert-manager

```suggestion
allowing access to trust data that is either known to cert-manager or external
to any cert-manager [Issuer](https://cert-manager.io/docs/configuration/).
```

design/20210715.bundle-crd.md
```diff
@@ -0,0 +1,275 @@
...
+lived, and all images that require the CA data are built in house.
+
+Projects such as [Istio](https://istio.io/) (as well as Kubernetes itself) has
+had success with distributing trust roots, at runtime, by using a
+[well-known ConfigMap `istio-ca-root-cert`](https://istio.io/latest/docs/tasks/security/cert-management/plugin-ca-cert/)
+that is available within every namespace in a cluster. Bundle is an extension of
+this idea- to allow users to use the same approach with a resource that is
+expressive enough to meet most needs.
+
+## Goals
```
SgtCoDFish, line 95: question: :warning: I think this is an important one

Something that I don't think is mentioned in this document is updating a trust store. It seems like a hard problem, because I think in practice an update would require a cluster operator to roll every deployment which consumed the updated trust store, but equally it's an important thing to have.

Without the ability to update a trust store, we can't:

- rotate an expiring root
- replace/remove an exposed certificate
- easily update publicly-trusted CAs lists (if supported), which are regularly changed

It feels like we need a story on this; do you have any thoughts on it?

SgtCoDFish (commented):
I've left a bunch of comments for discussion; this is a great starting point and it's an interesting problem space.

(Feel free to ignore the nitpicks, they're not important but I thought it would be worth highlighting them)

design/20210715.bundle-crd.md
JoshVanL: >  Something that I don't think is mentioned in this document is updating a trust store.

This is left up to the cluster operator and shouldn't be something that Bundle is concerned with. Bundle will ensure all targets are up to date at all times, but the cluster operator must use either: 1. use moderns apps that watch for changes to trust data files, or 2. use some custom solution like a sidecar or manual rollout.

SgtCoDFish: That's a fair position to take; it might be worth adding something to that effect into the doc

JoshVanL: Yes, I agree.

design/20210715.bundle-crd.md
```diff
@@ -0,0 +1,275 @@
...
+While sources currently support `ConfigMap`, `Secret` and `InLine`, it could be
+extended to other resources in future, or even remote servers in the aid of
+custom informers.
+
+Similarly, target only supports `ConfigMap` but can be extended to other
+resources or backends in future.
+
+Each source data has spacing trimmed at the prefix and suffix, and is appended
+together with a new line character.
+
```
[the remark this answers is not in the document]
JoshVanL: I'll make that clear.

We _don't_ remove any such comments, and specifically have no requirements on the contents of the data.

JoshVanL:
Thank you for the review @SgtCoDFish! Hopefully have covered all your comments

design/20210715.bundle-crd.md
```diff
@@ -0,0 +1,275 @@
...
+A single resource `Bundle` within the `trust.cert-manager.io` group makes up
+the trust operator. This resource fetches and appends data from a number of
+`sources` in the trust namespace, and distributes them to a `target`.
+
+The trust namespace is the same namespace that the trust operator is deployed
+to, which would normally be the `cert-manager` namespace.
+
+While sources currently support `ConfigMap`, `Secret` and `InLine`, it could be
+extended to other resources in future, or even remote servers in the aid of
+custom informers.
```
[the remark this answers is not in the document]
JoshVanL: In two minds about this.

While I absolutely agree this would be a great addition, I feel we run the risk of;
1. The baked in trust bundles becoming out of date meaning the tool has an expiry data if left unmaintained,
2. Open a flood gate of "also add this default trust store" feature creep.

SgtCoDFish:
Replying to [this](https://github.com/cert-manager/trust/pull/2#discussion_r671249796) here because it'll be easier to track. This is quite long; maybe we should have a call to discuss this.

> While I absolutely agree this would be a great addition, I feel we run the risk of;
>    1. The baked in trust bundles becoming out of date meaning the tool has an expiry data if left unmaintained,
>    2. Open a flood gate of "also add this default trust store" feature creep.

I definitely understand why you'd be in two minds, but I'm not hugely concerned by either of these things.

> 1. The baked in trust bundles becoming out of date meaning the tool has an expiry [date] if left unmaintained,

I think there's a case to be made for allowing the tool to dynamically fetch updated trust stores at runtime, although there are definitely questions which would need to be answered to implement that (where to fetch from, how often, how to surface an updated trust bundle to cluster admins, etc).

Whether or not we implement dynamic updating at runtime, I don't think most cluster operators would be overly concerned by using a baked trust store, since a lot of them are just going to be trusting whatever's provided in their container images, be that alpine, centos, debian, whatever.

By "centralising" on cert-manager/trust, they potentially only have to update one trust store (via a `Bundle`) rather than having to keep track of the disparate trust stores in their running applications on different distros.

Plus if we stopped maintaining cert-manager/trust, I think they'd be fine; it's open source and they could easily create new versions with updated baked trust stores.

> 2. Open a flood gate of "also add this default trust store" feature creep.

In theory I'd be concerned about this, but I don't think in practice there are actually many lists of publicly trusted roots that people would want in the first place. Most Linux distros that I know of just use Mozilla's list, and the only other major list I'd expect to hear about is [Chrome/Chromium's](https://www.chromium.org/Home/chromium-security/root-ca-policy). If in the worst case people start requesting a bunch of stuff we don't want to add, I'm personally quite comfortable saying "no" :wink: 

## UX Without Publicly Trusted Roots

My fear is that if we don't add support for a publicly trusted list of roots, the UX will suffer massively. The ideal UX for TLS clients connecting to servers IMO is:

- Create Bundle resource + mount resulting target ConfigMap into their containers
- Use TLS as normal both with public and private CAs

That is, they could mount the trust store and replace the system trust store for their container - so they'd not need to modify [RootCAs in tls.Config](https://pkg.go.dev/crypto/tls#Config), [`--ca-file` in curl](https://curl.se/docs/sslcerts.html) or whatever the equivalent is for their language / tool.

They _could_ combine the mounted ConfigMap and the system trust store in the container, but that requires them to run something like [this](https://unix.stackexchange.com/a/464495) at container start-up time, which is a pain and is error prone.

Basically, my vision here is that cert-manager/trust could be the one-stop-shop for trust everywhere in a cluster, providing visibility and a central place to update trust stores.

SgtCoDFish:
/lgtm

I don't think holding this design doc PR open is providing any value - we can always discuss potential features in their own issues. Specifically I'll create an issue from [this comment](https://github.com/cert-manager/trust/pull/2#issuecomment-882482107).

In any case, this doc is really good and well written as-is, so let's merge!

PR #3, opened 2021-07-16
Fix CI

JoshVanL:
Signed-off-by: joshvanl <vleeuwenjoshua@gmail.com>

- Set kubebuilder-tools dependency version to 1.21.2 by default
- Properly set REPO_ROOT `/hack/ci` env var
- Set CIDR range of kind cluster creation
- Increase eventually timeout to 10 seconds in tests


JoshVanL:
```
Error: looks like "https://charts.jetstack.io" is not a valid chart repository or cannot be reached: Get "https://charts.jetstack.io/index.yaml": dial tcp: lookup charts.jetstack.io on 10.8.240.10:53: read udp 10.8.1.19:42716->10.8.240.10:53: i/o timeout
```

:disappointed:

jakexks (approved):
/lgtm

PR #18, opened 2022-04-26
Update logo + link

SgtCoDFish:
See https://github.com/cert-manager/csi-driver/pull/80 for justification on logo changes


SgtCoDFish:
(added an LGTM manually because this is so trivial it doesn't seem worth pestering people about)

PR #21, opened 2022-04-30
Makefile: go vet

JoshVanL:
Adds `go vet ./...` as part of the Makefile `lint` target.

Fixes all the `go vet` linting errors.

/assign @SgtCoDFish

PR #25, opened 2022-05-03
Bundle: `namespaceSelector`

### LINKED ISSUE #1: Limiting what namespaces trust bundles are synced to
I notice in the README it states `and syncs them to a target in every namespace.` - is there any capability to limit _which_ namespaces a particular bundle is synced to? e.g. by label, or explicitly by name?

JoshVanL:
**Branched from #21 which should be merged first**

This PR adds a `namespaceSelector` option to the `bundle.spec.target` API.

The namespace selector is a normal `metav1.LabelSelector`.

If the label selector is empty, it matches on all Namespaces (the current behaviour).

Hopefully this is quite self explanatory :slightly_smiling_face: 

A target ConfigMap will now receive an event with a relevant message when the target ConfigMap is _not_ deleted because it is not owned by the trust operator.

/assign @munnerz 
fixes #1


pkg/bundle/util_test.go
```diff
@@ -60,7 +60,7 @@ func Test_bundleHasCondition(t *testing.T) {
 			expectHasCondition: true,
 		},
 		"an existing condition with a different LastTransitionTime should return true still": {
-			existingConditions: []trustapi.BundleCondition{{Reason: "A", ObservedGeneration: bundleGeneration, LastTransitionTime: &metav1.Time{fixedTime.Add(-time.Second)}}},
+			existingConditions: []trustapi.BundleCondition{{Reason: "A", ObservedGeneration: bundleGeneration, LastTransitionTime: &metav1.Time{Time: fixedTime.Add(-time.Second)}}},
```
jakexks, line 63: thought: is this just a linter change?

hack/boilerplate/boilerplate.py
```diff
@@ -202,9 +202,12 @@ def get_regexs():
     regexs["year"] = re.compile('YEAR')
     # dates can be 2014, 2015, 2016, or 2017; company holder names can be anything
     regexs["date"] = re.compile('(2014|2015|2016|2017)')
-    # strip // +build \n\n build constraints
-    regexs["go_build_constraints"] = re.compile(r"^(// \+build.*\n)+\n",
-                                                re.MULTILINE)
+    # strip the following build constraints/tags:
```
jakexks, line 205: comment: this is for new go 1.18 style go build tags.

thought: what do we think about this script in general? It's not 2017 any more!

pkg/bundle/sync.go
```diff
@@ -148,8 +163,19 @@ func (b *bundle) syncTarget(ctx context.Context, log logr.Logger, bundle *trusta
...
 
-	var needsUpdate bool
+	// Here, the config map exists, but the selector doesn't match the namespace.
+	if !matchNamespace {
+		// The ConfigMap is owned by this controller- delete it.
+		if metav1.IsControlledBy(&configMap, bundle) {
+			log.V(2).Info("deleting bundle from Namespace since namespaceSelector does not match")
+			return true, b.client.Delete(ctx, &configMap)
+		}
+		// The ConfigMap isn't owned by us, so we shouldn't delete it. Return that
```
jakexks, line 173: Should there be a warning event here?

jakexks (commented):
Looks good so far!

pkg/bundle/util_test.go
JoshVanL: Yes

hack/boilerplate/boilerplate.py
JoshVanL: Yes, something to resolve but not for this PR.

JoshVanL: I def think we can do better!

pkg/bundle/sync.go
JoshVanL: This has now been implemented by sending an event to the ConfigMap when the ConfigMap is not deleted due to no longer being owner by trust.

PR #27, opened 2022-05-03
Use the trust namespace as the namespace where the role and rolebinding are created in

### LINKED ISSUE #16: Trust namespace must be equals to deployment namespace
Hello,
 I want to install this application to different namespace then `app.trust.namespace` but I got error about not sufficient permission to create leader election lock. My use case is to install `trust` application several times in `cert-manager` namespace but to point to different `trust namespace`.

The problem is caused by https://github.com/cert-manager/trust/blob/main/deploy/charts/trust/templates/role.yaml that creates permission in deployment namespace not in `trust namespace` causing application does not have sufficient permission to create leader election lock and to watch secrets in `trust namespace`

JoshVanL:
Fixes #16 

/assign @SgtCoDFish

### DIFF
--- a/deploy/charts/trust/README.md
+++ b/deploy/charts/trust/README.md
@@ -8,9 +8,9 @@ A Helm chart for cert-manager-trust
 
 ## Maintainers
 
-|           Name           |                   Email                   |          URL            |
-| ------------------------ | ----------------------------------------- | ----------------------- |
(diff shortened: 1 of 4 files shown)


SgtCoDFish (approved):
/lgtm
/approve

Seems like a clear fix :ok_hand:

PR #28, opened 2022-05-13
Adds verify-helm-docs.sh and update-helm-docs.sh scripts. Verify helm-docs during lint 

JoshVanL:
/assign @SgtCoDFish

This PR adds a scripts, run at make lint, which verifies that the helm chart README.md file is kept up to date.

An update script is also added which will update the helm chart README.md.


JoshVanL:
Unrelated
/retest

PR #32, opened 2022-06-09
Remove same bundle target key validation check 

### LINKED ISSUE #29: Support targets with the same key
Hi Folks, 

I played around a bit and ran into an error when using targets with the same key. Unless there's a way, that satisfies the use case, I find it somewhat limiting. 
Let's take the following example: 

I want to distribute the root ca bundle for the webhooks in linkerd: 
```yaml 
kind: Bundle
metadata:
  name: linkerd-webhook-trust-roots
spec:
  sources:
    - secret:
        name: "linkerd-webhook-trust-roots"
        key: "ca.crt"
  target:
    configMap:
      key: "ca-bundle.crt"
```

Also, I want to distribute the root ca bundle for the linkerd trust anchor: 
```yaml 
kind: Bundle
metadata:
  name: linkerd-identity-trust-roots
spec:
  sources:
    - secret:
        name: "linkerd-identity-trust-roots"
        key: "ca.crt"
  target:
    configMap:
      key: "ca-bundle.crt"
```

This fails with `admission webhook "trust.cert-manager.io" denied the request: spec.target.configMap.key: Invalid value: "ca-bundle.crt": cannot use the same target as another Bundle "linkerd-identity-trust-roots"`. 

Is there a specific reason for not supporting that kind of behaviour?

JoshVanL:
fixes #29 

Currently, we have validation in the admission webhook to ensure that Bundle targets cannot be the same as another Bundle target. This makes no sense, since Bundles are cluster scoped resources and resulting targets (today only ConfigMaps) are named, and therefore namespaced, by the Bundle themselves. There can only be a single target per Bundle.

Basically, the current validation makes no sense, and we should remove it.

---

As an aside, we _might_ now be able to push all of these validations into the API schema itself and do away with needing a networked admission webhook all together (:smile:).

/assign @jakexks

### DIFF
--- a/pkg/webhook/validation.go
+++ b/pkg/webhook/validation.go
@@ -25,10 +25,8 @@ import (
 	"sync"
 
 	"github.com/go-logr/logr"
-	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/validation/field"
-	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 
 	"github.com/cert-manager/trust/pkg/apis/trust"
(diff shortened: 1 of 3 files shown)


irbekrm:
So we don't need the validation because the name of the target is the name of the (cluster-scoped) Bundle. Makes sense 👍🏼 

/lgtm
/approve

PR #36, opened 2022-07-07
WIP: Add Mozilla Root certs as a bundle source

### LINKED ISSUE #8: Publicly trusted certs in bundles
This issue is raised from discussion in #2, regarding whether or not to add support for adding publicly trusted roots to cert-manager/trust.

I'm raising the issue now as a point of discussion for this feature in the future, not because I'm pushing for it at the time I'm raising it - in other words, this is a cleanup.

My initial proposal:

I think Bundle would be tremendously improved by supporting the ability to include publicly trusted CAs, such as Mozilla's list. I worry that if we don't support a publicly trusted list, people will hack around it and get themselves into all kinds of trouble by statically including a bundle which will inevitably never get updated.

I don't know if such a publicly trusted list is required for a v1 release, but I think it would be a huge improvement to have it in place; it'd enable Bundle to essentially be the one-stop-shop for all trust needs for the vast majority of workloads.

> While I absolutely agree this would be a great addition, I feel we run the risk of;
>    1. The baked in trust bundles becoming out of date meaning the tool has an expiry data if left unmaintained,
>    2. Open a flood gate of "also add this default trust store" feature creep.

I definitely understand why you'd be in two minds, but I'm not hugely concerned by either of these things.

> 1. The baked in trust bundles becoming out of date meaning the tool has an expiry [date] if left unmaintained,

I think there's a case to be made for allowing the tool to dynamically fetch updated trust stores at runtime, although there are definitely questions which would need to be answered to implement that (where to fetch from, how often, how to surface an updated trust bundle to cluster admins, etc).

Whether or not we implement dynamic updating at runtime, I don't think most cluster operators would be overly concerned by using a baked trust store, since a lot of them are just going to be trusting whatever's provided in their container images, be that alpine, centos, debian, whatever.

By "centralising" on cert-manager/trust, they potentially only have to update one trust store (via a `Bundle`) rather than having to keep track of the disparate trust stores in their running applications on different distros.

Plus if we stopped maintaining cert-manager/trust, I think they'd be fine; it's open source and they could easily create new versions with updated baked trust stores.

> 2. Open a flood gate of "also add this default trust store" feature creep.

In theory I'd be concerned about this, but I don't think in practice there are actually many lists of publicly trusted roots that people would want in the first place. Most Linux distros that I know of just use Mozilla's list, and the only other major list I'd expect to hear about is [Chrome/Chromium's](https://www.chromium.org/Home/chromium-security/root-ca-policy). If in the worst case people start requesting a bunch of stuff we don't want to add, I'm personally quite comfortable saying "no" :wink: 

## UX Without Publicly Trusted Roots

My fear is that if we don't add support for a publicly trusted list of roots, the UX will suffer massively. The ideal UX for TLS clients connecting to servers IMO is:

- Create Bundle resource + mount resulting target ConfigMap into their containers
- Use TLS as normal both with public and private CAs

That is, they could mount the trust store and replace the system trust store for their container - so they'd not need to modify [RootCAs in tls.Config](https://pkg.go.dev/crypto/tls#Config), [`--ca-file` in curl](https://curl.se/docs/sslcerts.html) or whatever the equivalent is for their language / tool.

They _could_ combine the mounted ConfigMap and the system trust store in the container, but that requires them to run something like [this](https://unix.stackexchange.com/a/464495) at container start-up time, which is a pain and is error prone.

Basically, my vision here is that cert-manager/trust could be the one-stop-shop for trust everywhere in a cluster, providing visibility and a central place to update trust stores.

_Originally posted by @SgtCoDFish in https://github.com/cert-manager/trust/issues/2#issuecomment-882482107_

jakexks:
Fixes #8 
/kind feature

This embeds the latest version of the Mozilla Root Certificates that have the web SSL/TLS bit set, using https://github.com/breml/rootcerts

This adds a new bundle source: CCADB, which will add said embedded root CAs to the target.

breml/rootcerts is under consideration to be included in the Go standard library, but until then, keeping it up-to-date should be seen as a vital security concern, perhaps automated with dependabot.

### DIFF
--- a/pkg/apis/trust/v1alpha1/types_bundle.go
+++ b/pkg/apis/trust/v1alpha1/types_bundle.go
@@ -76,6 +76,10 @@ type BundleSource struct {
 	// InLine is a simple string to append as the source data.
 	// +optional
 	InLine *string `json:"inLine,omitempty"`
+
+	// CCADB is an embedded copy of the certificates in Mozilla's root store
+	// that have the websites trust bit enabled.
+	CCADB *bool `json:"ccadb,omitempty"`
 }
 
 // BundleTarget is the target resource that the Bundle will sync all source
--- a/pkg/bundle/sync.go
+++ b/pkg/bundle/sync.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/breml/rootcerts/embedded"
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -54,6 +55,11 @@ func (b *bundle) buildSourceBundle(ctx context.Context, bundle *trustapi.Bundle)
 
 		case source.InLine != nil:
 			sourceData = *source.InLine
+
+		case source.CCADB != nil:
+			if *source.CCADB {
+				sourceData = embedded.MozillaCACertificatesPEM()
+			}
 		}
 
 		if err != nil {
--- a/deploy/charts/trust/templates/trust.cert-manager.io_bundles.yaml
+++ b/deploy/charts/trust/templates/trust.cert-manager.io_bundles.yaml
@@ -2,7 +2,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.6.1
+    controller-gen.kubebuilder.io/version: v0.8.0
   creationTimestamp: null
   name: bundles.trust.cert-manager.io
 spec:
@@ -60,6 +60,9 @@ spec:
                     description: BundleSource is the set of sources whose data will be appended and synced to the BundleTarget in all Namespaces.
                     type: object
                     properties:
+                      ccadb:
+                        description: CCADB is an embedded copy of the certificates in Mozilla's root store that have the websites trust bit enabled.
+                        type: boolean
                       configMap:
                         description: ConfigMap is a reference to a ConfigMap's `data` key, in the trust Namespace.
                         type: object
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module github.com/cert-manager/trust
 go 1.18
 
 require (
+	github.com/breml/rootcerts v0.2.4
 	github.com/go-logr/logr v1.2.3
 	github.com/onsi/ginkgo v1.16.5
 	github.com/onsi/ginkgo/v2 v2.1.3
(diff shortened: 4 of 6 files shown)


pkg/apis/trust/v1alpha1/types_bundle.go
```diff
@@ -76,6 +76,10 @@ type BundleSource struct {
 	// InLine is a simple string to append as the source data.
 	// +optional
 	InLine *string `json:"inLine,omitempty"`
+
+	// CCADB is an embedded copy of the certificates in Mozilla's root store
+	// that have the websites trust bit enabled.
+	CCADB *bool `json:"ccadb,omitempty"`
```
munnerz, line 82: I `*string` typed field with e.g. `publicTrustBundle: CCADB` allows for further extensibility in future - not sure on exact field naming, but the Kubernetes API convention calls out to be careful with bool typed fields (see the final bullet point here https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types)

jakexks: 👍 makes sense

SgtCoDFish: Would it be worth considering a different name to "CCADB"? My instinct is that a user would have no idea what a "CCADB" is and wouldn't find the name very easy to remember.

`mozilla` would match closer to what I'd probably expect, or maybe something like `embedded` to reflect that this is a trust store embedded in the `trust` tool.

pkg/bundle/sync.go
```diff
@@ -54,6 +55,11 @@ func (b *bundle) buildSourceBundle(ctx context.Context, bundle *trustapi.Bundle)
 
 		case source.InLine != nil:
 			sourceData = *source.InLine
+
+		case source.CCADB != nil:
+			if *source.CCADB {
+				sourceData = embedded.MozillaCACertificatesPEM()
+			}
```
SgtCoDFish, line 62: I hadn't really considered the idea of providing a bundle embedded into the trust project itself - it makes sense, especially in airgapped environments.

(The thing I'd envisioned in #8  - which I'd been meaning to write a PR for - was that trust would fetch + use the latest trust bundle available on a regular basis; obviously that's a more complicated thing to do, and doesn't really have to impact this PR)

munnerz: Embedding a set of CAs into this project is something it's not been used for before (up until now it's only ever dealt with moving CAs around, now it is actual a source of CAs). So I think we should consider very carefully whether this is what we want to do (in this way)

SgtCoDFish:
I don't think this is the approach we've agreed to take so I'll close this - if I'm wrong feel free to re-open!

PR #37, opened 2022-07-18
Bundle: `namespaceSelector`

JoshVanL:
This PR adds a namespaceSelector option to the bundle.spec.target API.

The namespace selector currently only contains a `MatchLabels map[string]string` field.

If the label selector is empty, it matches on all Namespaces (the current behaviour).

Hopefully this is quite self explanatory slightly_smiling_face

A target ConfigMap will now receive an event with a relevant message when the target ConfigMap is not deleted because it is not owned by the trust operator.

/assign @munnerz
fixes https://github.com/cert-manager/trust/issues/1
[Docs PR](https://github.com/cert-manager/website/pull/1038)


SgtCoDFish (approved):
/lgtm
/approve
/hold

Looks like a good improvement to me :grin:

I'm a little squeamish about syncing to all namespaces as the default. I'd probably prefer to default only to syncing to the trust namespace; IMO since this project is so new it's reasonable for us to make a breaking change like that.

But that's a tangent from this PR which is good as-is :grin:

JoshVanL:
@SgtCoDFish As discussed on the standup, I think I agree that syncing to all namespaces by _default_ is probably not the way to go, and we should make the namespace selector a required field in future PRs.

/hold cancel

SgtCoDFish:
> As discussed on the standup...

Cool cool; I'll create issues now for that and for the root cert bundle thing we mentioned!

jaygridley:
Hello @SgtCoDFish, a new version (0.1.2) of the Helm chart with support on CRD level was released, but the functionality in the application itself was not released. It there a plan to align this, please?

SgtCoDFish:
> Hello @SgtCoDFish, a new version (0.1.2) of the Helm chart with support on CRD level was released, but the functionality in the application itself was not released. It there a plan to align this, please?

Hi! maybe @JoshVanL can help, I can't I'm afraid

JoshVanL:
@jaygridley I'll go ahead with a minor release with this feature now :slightly_smiling_face:

PR #38, opened 2022-07-20
Add chart building + ensure BINDIR is created when running targets

SgtCoDFish:
1. Adds `make chart` to generate a local helm chart for testing
2. Creates `$(BINDIR)` and subdirectories using order only dependencies (see details [in the cert-manager repo](https://github.com/cert-manager/cert-manager/tree/master/make#target-dependencies--prerequisites))
3. Use `$@` where possible
4. Adds release version as a variable since it's now used in more than one place

I'd probably also advocate changing `$(BINDIR)` to default to `_bin` for the [same reasons we did that in cert-manager](https://github.com/cert-manager/cert-manager/pull/5130) but that seemed maybe a change too far for this PR

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 BINDIR ?= $(CURDIR)/bin
+
 ARCH   ?= $(shell go env GOARCH)
 OS     ?= $(shell go env GOOS)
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -54,8 +57,12 @@ verify: depend test build ## tests and builds trust
 # To actually push an image to the public repo, replace the `--output` flag and
 # arguments to `--push`.
 .PHONY: image
-image: ## build docker image targeting all supported platforms
-	docker buildx build --platform=$(IMAGE_PLATFORMS) -t quay.io/jetstack/cert-manager-trust:v0.1.0 --output type=local,dest=./bin/cert-manager-trust .
+image: | $(BINDIR) ## build docker image targeting all supported platforms
+	docker buildx build --platform=$(IMAGE_PLATFORMS) -t quay.io/jetstack/cert-manager-trust:v0.1.0 --output type=local,dest=./$(BINDIR)/cert-manager-trust .
```
JoshVanL, line 61:
```suggestion
	docker buildx build --platform=$(IMAGE_PLATFORMS) -t quay.io/jetstack/cert-manager-trust:v0.1.0 --output type=local,dest=$(BINDIR)/cert-manager-trust .
```

JoshVanL (changes requested):
Small thing from me :slightly_smiling_face:

PR #42, opened 2022-07-26
WIP: feat: add HTTPS bundle source

jakexks:
This PR adds a "fetch from remote server over HTTPS" bundle source. It's used like this:

```yaml
spec:
  sources:
  - https:
      url: https://foo.bar/baz
      caBundle: |
        -----BEGIN CERTIFICATE-----
        ...
      pollInterval: 60s
```

It uses etags / if-modified-since headers to avoid redownloading bundles every pollInterval.

It's somewhat opinionated - it only allows https:// schemes and will only use the trust roots specied inline in the spec. This is to prevent users doing something unsafe, but I'm open to discussion.

### DIFF
--- a/hack/boilerplate/boilerplate.Dockerfile.txt
+++ b/hack/boilerplate/boilerplate.Dockerfile.txt
@@ -1,4 +1,4 @@
-# Copyright 2021 The cert-manager Authors.
+# Copyright YEAR The cert-manager Authors.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/pkg/apis/trust/v1alpha1/types_bundle.go
+++ b/pkg/apis/trust/v1alpha1/types_bundle.go
@@ -51,7 +51,7 @@ type BundleList struct {
 	Items []Bundle `json:"items"`
 }
 
-// BundleSepc defines the desired state of a Bundle.
+// BundleSpec defines the desired state of a Bundle.
 type BundleSpec struct {
 	// Sources is a set of references to data whose data will sync to the target.
 	Sources []BundleSource `json:"sources"`
@@ -76,6 +76,21 @@ type BundleSource struct {
 	// InLine is a simple string to append as the source data.
 	// +optional
 	InLine *string `json:"inLine,omitempty"`
+
+	// HTTPS downloads a PEM-Formatted trust bundle from an HTTPS server
+	// +optional
+	HTTPS *SourceHTTPS `json:"https,omitempty"`
+}
+
+// SourceHTTPS represents a target HTTPS URL that will serve a PEM-formatted trust bundle.
+type SourceHTTPS struct {
+	// URL is an HTTPS URL.
+	URL string `json:"url"`
+	// CABundle is the CA Bundle required to verify the target server's identity.
+	CABundle string `json:"caBundle"`
+	// PollInterval is the period at which the controller will check for changes on the server.
+	// +optional
+	PollInterval *string `json:"pollInterval,omitempty"`
 }
 
 // BundleTarget is the target resource that the Bundle will sync all source
--- a/pkg/bundle/bundle.go
+++ b/pkg/bundle/bundle.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"time"
 
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
@@ -62,6 +63,10 @@ type bundle struct {
 
 	// clock returns time which can be overwritten for testing.
 	clock clock.Clock
+
+	// etagCache holds cached responses with their date and etag, so HTTPS bundles
+	// do not cause unnecessary re-downloading
+	etagCache *etagCache
 
 	// Options hold options for the Bundle controller.
 	Options
@@ -215,13 +220,30 @@ func (b *bundle) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, 
 		Message: message,
 	}
 
+	// Requeue after the minimum poll interval time of all HTTPS bundles
+	requeueAfter := time.Duration(0)
+	for _, s := range bundle.Spec.Sources {
+		if s.HTTPS != nil {
+			if s.HTTPS.PollInterval != nil {
+				interval, err := time.ParseDuration(*s.HTTPS.PollInterval)
+				if err != nil {
+					// should never happen as we have passed the validating webhook
+					continue
+				}
+				if interval > requeueAfter {
+					requeueAfter = interval
+				}
+			}
+		}
+	}
+
 	if !needsUpdate && bundleHasCondition(&bundle, syncedCondition) {
-		return ctrl.Result{}, nil
+		return ctrl.Result{RequeueAfter: requeueAfter}, nil
 	}
 
 	log.V(2).Info("successfully synced bundle")
 
 	b.setBundleCondition(&bundle, syncedCondition)
 	b.recorder.Eventf(&bundle, corev1.EventTypeNormal, "Synced", message)
-	return ctrl.Result{}, b.client.Status().Update(ctx, &bundle)
+	return ctrl.Result{RequeueAfter: requeueAfter}, b.client.Status().Update(ctx, &bundle)
 }
--- a/pkg/bundle/sync.go
+++ b/pkg/bundle/sync.go
@@ -18,9 +18,15 @@ package bundle
 
 import (
 	"context"
+	"crypto/tls"
+	"crypto/x509"
+	"encoding/pem"
 	"errors"
 	"fmt"
+	"io"
+	"net/http"
 	"strings"
+	"time"
 
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
@@ -55,6 +61,9 @@ func (b *bundle) buildSourceBundle(ctx context.Context, bundle *trustapi.Bundle)
 
 		case source.InLine != nil:
 			sourceData = *source.InLine
+
+		case source.HTTPS != nil:
+			sourceData, err = b.httpsBundle(ctx, source.HTTPS)
 		}
 
 		if err != nil {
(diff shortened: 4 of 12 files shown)


pkg/bundle/bundle.go
```diff
@@ -215,13 +220,30 @@ func (b *bundle) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result,
 		Message: message,
 	}
 
+	// Requeue after the minimum poll interval time of all HTTPS bundles
+	requeueAfter := time.Duration(0)
```
SgtCoDFish, line 224: note: It seems the obvious default now but I had to [check](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.12.3/pkg/reconcile#Result) that a duration of 0 means "don't requeue" and not "requeue immediately" :grin:

hack/boilerplate/boilerplate.Dockerfile.txt
```diff
@@ -1,4 +1,4 @@
-# Copyright 2021 The cert-manager Authors.
+# Copyright YEAR The cert-manager Authors.
```
JoshVanL, line 1: Thank you for doing this! Could you please split out the boilerplate changes into a separate PR to have separation of concerns?

pkg/apis/trust/v1alpha1/types_bundle.go
```diff
@@ -76,6 +76,21 @@ type BundleSource struct {
 	// InLine is a simple string to append as the source data.
 	// +optional
 	InLine *string `json:"inLine,omitempty"`
+
+	// HTTPS downloads a PEM-Formatted trust bundle from an HTTPS server
```
JoshVanL, line 80: We are not _necessarily_ downloading PEM encoded trust bundles. trust is not _currently_ opinionated about the form of the trust bundles- whether they be X.509 CAs, public keys, or just some text.


 ```suggestion
	// HTTPS downloads trust bundles from an HTTPS server
```

File: pkg/bundle/sync.go [hunk 4]
```diff
@@ -113,6 +122,75 @@ func (b *bundle) secretBundle(ctx context.Context, ref *trustapi.SourceObjectKey
...
+		return "", fmt.Errorf("couldn't make http request: %w", err)
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode == http.StatusNotModified {
+		b.etagCache.RLock()
+		bundleData := b.etagCache.cache[ref.URL].data
+		b.etagCache.RUnlock()
+		return string(bundleData), nil
+	}
+	body, err := io.ReadAll(resp.Body)
```
SgtCoDFish, line 175: suggestion: This feels too trusting of the server! The server could have returned anything except StatusNotModified here and we'd cache the response body and use it, even if it was a 404 or 500 or any error.

Would it be useful here if we checked that the response is 2xx and checking that the response looks something like a trust bundle? E.g. if we can't parse at least one PEM encoded certificate in the response, we shouldn't cache the response and use it.

pkg/apis/trust/v1alpha1/types_bundle.go
```diff
@@ -76,6 +76,21 @@ type BundleSource struct {
...
+	// +optional
+	HTTPS *SourceHTTPS `json:"https,omitempty"`
+}
+
+// SourceHTTPS represents a target HTTPS URL that will serve a PEM-formatted trust bundle.
```
JoshVanL, line 85:
```suggestion
// SourceHTTPS represents a source HTTPS URL that will serve trust bundles.
```

pkg/apis/trust/v1alpha1/types_bundle.go
```diff
@@ -76,6 +76,21 @@ type BundleSource struct {
...
+type SourceHTTPS struct {
+	// URL is an HTTPS URL.
+	URL string `json:"url"`
+	// CABundle is the CA Bundle required to verify the target server's identity.
```
JoshVanL, line 89:
```suggestion
	URL string `json:"url"`
	
	// CABundle is the CA Bundle required to verify the target server's identity.
```

JoshVanL, line 89:
```suggestion
  // CABundle is the X.509 PEM encoded CA Bundle required to verify the target
  // server's identity. Will be used when establishing a TLS connection to the
  // remote server.
```

pkg/apis/trust/v1alpha1/types_bundle.go
```diff
@@ -76,6 +76,21 @@ type BundleSource struct {
...
+	CABundle string `json:"caBundle"`
+	// PollInterval is the period at which the controller will check for changes on the server.
```
JoshVanL, line 91:
```suggestion
	CABundle string `json:"caBundle"`
	
	// PollInterval is the period at which the controller will check for changes on the server.
```

SgtCoDFish (commented):
Left a few comments; I'm not sure if this was meant to be "reviewed" as such given you mentioned it might be to spark discussion, so all of this is just discussion stuff really.

I still need to think about this more - I might have to continue writing my WIP "design doc" for this topic to untangle it in my head!

pkg/apis/trust/v1alpha1/types_bundle.go
```diff
@@ -76,6 +76,21 @@ type BundleSource struct {
...
+	// +optional
+	PollInterval *string `json:"pollInterval,omitempty"`
```
JoshVanL, line 93: Should we instead change this to be `PollIntervalSeconds int64`? This is more inline with the Kubernetes field conventions.

https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types
https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#units

We should also have a default here which is managed by OpenAPI schema (kubebuilder headers). https://book.kubebuilder.io/reference/markers/crd-validation.html

pkg/bundle/bundle.go
```diff
@@ -63,6 +64,10 @@ type bundle struct {
 	// clock returns time which can be overwritten for testing.
 	clock clock.Clock
 
+	// etagCache holds cached responses with their date and etag, so HTTPS bundles
+	// do not cause unnecessary re-downloading
```
JoshVanL, line 68:
```suggestion
	// do not cause unnecessary re-downloading.
```

pkg/bundle/bundle.go
```diff
@@ -215,13 +220,30 @@ func (b *bundle) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result,
...
 	}
 
+	// Requeue after the minimum poll interval time of all HTTPS bundles
+	requeueAfter := time.Duration(0)
+	for _, s := range bundle.Spec.Sources {
+		if s.HTTPS != nil {
+			if s.HTTPS.PollInterval != nil {
+				interval, err := time.ParseDuration(*s.HTTPS.PollInterval)
+				if err != nil {
+					// should never happen as we have passed the validating webhook
```
JoshVanL, line 230: We should definitely return an error here, since something has gone so wrong.

pkg/bundle/bundle.go
```diff
@@ -215,13 +220,30 @@ func (b *bundle) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result,
 		Message: message,
 	}
 
+	// Requeue after the minimum poll interval time of all HTTPS bundles
+	requeueAfter := time.Duration(0)
+	for _, s := range bundle.Spec.Sources {
+		if s.HTTPS != nil {
+			if s.HTTPS.PollInterval != nil {
```
JoshVanL, line 227: Should we not require this field?

munnerz:
Hmm, I am pretty hesitant about adding something that is a dynamic/changeable source. This creates issues for trust in general, as well as observability into versioning.

From an outside perspective, I'd expect that all trust bundles defined with our CRDs are fairly immutable, so if a new version of something were to come about then a new bundle would be created and whatever standard process people have in place for migrating from the older trust store to the new one would 'kick in' (all of this can be automated away behind the scenes, but my point here being that there is one and only one way to actually update a trust store, making audit a lot easier).

Why do we want `trust` to deal with interacting with an external server, rather than say, an administrator doing this periodically or a controller/cron/script that periodically runs and updates/creates a `TrustBundle` resource in the apiserver when it deems necessary? IMO doing it that way would be closer to a 'pluggable' model too, as some orgs somewhere may have some kind of proprietary store for trust bundles that they need their own controller for reading from. Fetching from a multitude of sources seems like it risks scope creep similar to how we now have in-built issuers in cert-manager core.

Just my 2c on this really :) it'd be good to hear the use-case that has led to this, and whether there's a more general mechanism for "integrating external trust sources" that we should be discussing

pkg/bundle/bundle.go
JoshVanL, line 224: This looks like to me that requeueAfter will be 0 if there are no HTTPS sources. I _think_ that is fine because a requeueAfter value of 0 is ignored in controller-runtime (?). Can you add a comment to that effect?

JoshVanL (changes requested):
I seem to be a little too slow for the review here by O(minutes) :laughing: 

I'll stop where I am to prevent duplication and pile on.

JoshVanL:
One more thing.. 
I don't think it makes sense to require the protocol in the URI when the block is already HTTPS

```yaml
  - https:
      url: https://foo.bar/baz
```

pkg/bundle/sync.go
```diff
@@ -113,6 +122,75 @@ func (b *bundle) secretBundle(ctx context.Context, ref *trustapi.SourceObjectKey
 	return string(data), nil
 }
 
+func (b *bundle) httpsBundle(ctx context.Context, ref *trustapi.SourceHTTPS) (string, error) {
+	// Only use CAs from the SourceHTTPS inline bundle:
+	cas := x509.NewCertPool()
```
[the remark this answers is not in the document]
JoshVanL: If we intend to include a set of pre-loaded trust bundles (mozilla), then we could allow a reference to one of those.

munnerz:
> Hmm, I am pretty hesitant about adding something that is a dynamic/changeable source. This creates issues for trust in general, as well as observability into versioning.
> 
> 
> 
> From an outside perspective, I'd expect that all trust bundles defined with our CRDs are fairly immutable, so if a new version of something were to come about then a new bundle would be created and whatever standard process people have in place for migrating from the older trust store to the new one would 'kick in' (all of this can be automated away behind the scenes, but my point here being that there is one and only one way to actually update a trust store, making audit a lot easier).
> 
> 
> 
> Why do we want `trust` to deal with interacting with an external server, rather than say, an administrator doing this periodically or a controller/cron/script that periodically runs and updates/creates a `TrustBundle` resource in the apiserver when it deems necessary? IMO doing it that way would be closer to a 'pluggable' model too, as some orgs somewhere may have some kind of proprietary store for trust bundles that they need their own controller for reading from. Fetching from a multitude of sources seems like it risks scope creep similar to how we now have in-built issuers in cert-manager core.
> 
> 
> 
> Just my 2c on this really :) it'd be good to hear the use-case that has led to this, and whether there's a more general mechanism for "integrating external trust sources" that we should be discussing

Please also don't take my comment as binding if there is general consensus already that we want this 😊 just trying to think of different ways this problem could be approached.

pkg/apis/trust/v1alpha1/types_bundle.go
SgtCoDFish: This comment is ringing major alarm bells for me! I think we _desperately_ need to be aligned on this going forwards. You're of course right about the current state of things.

There are two things here; the formats we accept as inputs and the formats we support as outputs. In either case I don't think it's sustainable  for `cert-manager/trust` to be a controller which essentially implements the unix `cat` command for text files.

I think in an ideal world we should be explicit about what each source accepts as an input, and _always_ validate that those inputs are in a format we expect.

We might want to allow different output formats too. Ideally we should also have exactly one "intermediate" format, to which all inputs can be converted, and from which all outputs can be generated.

If any of that is controversial I think we need to get on a call and make a plan for this or else we need to sit down and agree on a design document on how this would be handled. There's a huge risk of making a mess here I think.

jakexks:
No worries, there's no consensus on anything yet. My 2 PRs were intended to spark discussion ahead of the community meeting, and to try to solutioneer to feed back into forthcoming design docs.

To be more explicit I've stuck the WIP labels on my PRs 😄

SgtCoDFish:
As with #36 I don't think this is the approach we've agreed to take so I'll close this - if I'm wrong feel free to re-open!

PR #43, opened 2022-07-27
Add a design for public trust bundles

SgtCoDFish:
This proposal introduces a design by which bundles are stored in sidecar containers and consumed from a shared path by the controller container.

### DIFF
--- a/design/20220722-publicbundle.md
+++ b/design/20220722-publicbundle.md
@@ -1,0 +1,284 @@
+# Design: Static Public Trust Bundles
+
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [Bundle Container Server](#bundle-container-server)
+  - [Behaviour After Consumption](#behaviour-after-consumption)
+  - [Test Plan](#test-plan)
+  - [Graduation Criteria](#graduation-criteria)
+  - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+  - [Supported Versions](#supported-versions)
+- [Production Readiness](#production-readiness)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+
+## Release Signoff Checklist
+
+This checklist contains actions which must be completed before a PR implementing this design can be
+merged.
+
+- [ ] This design doc has been discussed and approved
+- [ ] Test plan has been agreed upon and the tests implemented
+- [ ] Feature gate status has been agreed upon (whether the new functionality will be placed behind a feature gate or not)
+- [ ] Graduation criteria is in place if required (if the new functionality is placed behind a feature gate, how will it graduate between stages)
+- [ ] User-facing documentation has been PR-ed against the release branch in [cert-manager/website]
+
+## Summary
+
+Most programs, programming languages and libraries consume their trust bundles from a standard
+location on-disk. On Linux, the standard bundle is typically provided by the distribution. The same
+applies to containers, in which the base image used is often the provider of a standard TLS trust bundle.
+
+Currently, `trust-manager` makes it easy to construct a bundle in a mechanical sense, but doesn't really
+help in choosing / collecting the various parts which actually go into a bundle. For private or
+organization-specific CAs, there's only a limited amount which trust-manager can do, but for public
+bundles - such as those mentioned above used by Linux distros - there's a role that trust-manager can play
+by providing a bundle for people to consume.
+
+We propose that we should start helping in what is likely the most common case: to add an option
+for a "public" trust bundle as a source, which should be as easy to consume as possible.
+
+## Motivation
+
+The ideal situation for the vast majority of applications is likely to be that they don't have to
+explicitly configure their application's trust bundle.
+
+By way of example, a Linux Golang application would [currently](https://github.com/golang/go/blob/2d655fb15a50036547a6bf8f77608aab9fb31368/src/crypto/x509/root_linux.go#L8-L15)
+look in the following locations by default:
+
+```go
+var certFiles = []string{
+	"/etc/ssl/certs/ca-certificates.crt",                // Debian/Ubuntu/Gentoo etc.
+	"/etc/pki/tls/certs/ca-bundle.crt",                  // Fedora/RHEL 6
+	"/etc/ssl/ca-bundle.pem",                            // OpenSUSE
+	"/etc/pki/tls/cacert.pem",                           // OpenELEC
+	"/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem", // CentOS/RHEL 7
+	"/etc/ssl/cert.pem",                                 // Alpine Linux
+}
+```
+
+It's a chore to manually specify a location for a trust store, and it's unfamiliar and error-prone
+for many users. However, if a user instead overwrites any of these files with their own
+trust-manager-generated certificate trust bundle they'll currently overwrite their public trust bundle
+provided by their OS / container base image.
+
+Unless the user's application only talks to services which the user also controls, what users are
+likely to really want in most cases is a bundle containing both the publicly trusted certs _and_
+the user's own roots.
+
+It's possible in most distros to append to the cert bundle, but that usually involves running some
+command to generate the final bundle, which would change how the user invokes their application
+inside their container.
+
+Ideally, the whole bundle a user requires would simply be available at one of the above locations.
+
+In Kubernetes, that today likely means overwriting the system trust bundle with an entire bundle
+at the point when the container is created. That implies that the Target which trust-manager creates
+must be completed before the container starts, and that in turn implies a need for the public
+trust bundles users rely on to be available as sources for trust-manager.
+
+### Goals
+
+- A user configuring a `Bundle` can add a public trust bundle at least as easily as they can add certs from a `ConfigMap` or `Secret`
+- It's clear to users exactly what version of the public bundle is used at any given time, for auditing purposes
+- A user whose application must use both public and private CAs can confidently overwrite their "system" trust store in a container using a trust `Bundle`
+- A user can update any bundle in use with any version of the trust-manager controller and be confident that it will work
+
+### Non-Goals
+
+- Editing/pruning the public trust bundle (out of scope for the initial feature but might be desirable later)
+- Supporting [OS-specific](https://superuser.com/questions/437330/how-do-you-add-a-certificate-authority-ca-to-ubuntu) certificate update mechanisms
+- Supporting a plethora of options for sources of public trust bundles; we might want this in the future but we should start with one
+- Dynamically updating the public bundle (users will have to explicitly choose to upgrade, at least for now)
+
+## Proposal
+
+We propose that the cert-manager project becomes a distributor of its own public trust bundle, which at
+least to start will be a carbon copy of the Debian `ca-certificates` package.
+
+That bundle in will in turn be packaged into a container which will be used as an init container in the
+trust-manager controller's pod, with the ability to present its bundle to the controller container.
+
+To disambiguate between "public trust bundles" and the trust-manager "Bundle" resource, we'll call
+the cert-manager project's public trust bundles "packages" instead.
+
+In the initial implementation of this feature, users will pass the location of a single JSON-formatted
+package to the trust-manager controller at startup. If given, this package must load successfully
+and have a nonzero number of valid certificates in it. For most users who use Helm, they'll not have
+to worry about this since the default trust package will be configured entirely in Helm.
+
+This single package then becomes the "default CA package", and it remains static for the lifetime of the
+trust-manager controller pod.
+
+In order to refer to the default package in a Bundle, we'll also add a new boolean source for
+requesting the package.
+
+```yaml
+apiVersion: trust.cert-manager.io/v1alpha1
+kind: Bundle
+metadata:
+  name: my-org.com
+spec:
+  sources:
+  - configMap:
+      name: "my-root-cm"
+      key: "root-cert.pem"
+  - useDefaultCAs: true
+  target:
+    ...
+```
+
+### Workflow Example
+
+1. The trust-manager pod starts with an init container: the `cert-manager-package-debian` container, which simply copies its JSON file to a predefined location and then exits.
+2. The copied package JSON looks like this:
+
+    ```json
+    {
+      "name": "cert-manager-debian",
+      "bundle": "-----BEGIN CERTIFICATE-----\n...\n-----END CERTIFICATE-----\n...",
+      "version": "<VERSION>.0"
+    }
+    ```
+
+3. The trust-manager controller starts, this time with a new argument: `--default-package-location=/path/to/package-cert-manager-debian.json`
+4. During startup, the controller validates the package and if succesful loads its `bundle` into memory for later use as the "default package".
+5. When a `Bundle` resource requests to `useDefaultCAs`, the `bundle` field from the default package is inserted into the target.
+
+### User Stories
+
+#### Story 1: Standard Platform Deployment
+
+Alice is a platform engineer using trust to ensure that the same trust bundle is present across all
+of the pods in her cluster.
+
+She upgrades to a new version of trust which supports the bundle container. The Helm chart points
+to a default version of the bundle which was set at the time the version of trust was released, and
+she's happy with the defaults.
+
+After rolling out the upgrade she edits her `Bundle` resources to `useDefaultCAs`. She kicks off
+a fresh deployment of all pods and now they all have all the certificates she needs to talk to both
+public services and services using her organization's private CA.
+
+#### Story 2: Security Incident
+
+Bob is a security engineer responding to a major security incident involving a root certificate
+whose private key was exposed on the internet.
+
+For each cluster in which trust-manager is deployed Bob updates the Helm `defaultPackageImage.tag` variable
+to point to the latest version of the bundle container. He then proceeds to roll out the bundle much
+as Alice did above - by rolling the deployments of each service. The vulnerability is patched safely.
+
+#### Story 3: Security Incident Without Packages
+
+(This story is intended to illustrate the problems an engineer would have today if trying to deal with
+a root CA private key being exposed)
+
+Charles is a security engineer responding to the same incident as in Story 2, but without already using
+trust-manager inside his clusters.
+
+He has to update the base image of every container the organisation builds to remove the compromised root,
+but that task requires several steps:
+
+1. Ensuring he has a list of every base image used across the organization. If he misses a container,
+   it'll be vulnerable to MitM attacks until it's updated.
+2. Making sure that each distinct base image used across the organization (Alpine / Debian / Distroless / etc)
+   has actually received the updated trust store. There's no point updating if there's nothing to update to!
+3. Rebuilding every single container which has an embedded trust bundle across the organization. In practice,
+   this is nearly all of them.
+4. Testing the newly built containers in CI to ensure that the updated base images haven't changed any other
+   functionality. If some containers were running an out-of-support base image, they won't get the updated
+   ca-certificates package, and Charles might need to bump to a new minor version which could be incompatible
+   with his applications.
+5. After rebuilding, he has to roll all deployments of all pods to make sure they're using the new images.
+
+### Risks and Mitigations
+
+#### Bundle Container Updates
+
+This proposal makes it absolutely critical that we have a smooth and rapid release process for the
+"package" container. A security incident involving a root certificate would be front-page news and a
+major incident for basically every organisation in the world and we'd have to treat it the same,
+to the point of waking engineers up in the middle of the night to cut a new release if required.
+
+Likewise, it's important that the package is validated during the build, to ensure it'll be usable.
+
+#### Bundle Container Attack Surface
+
+Since this proposal includes a need for some kind of process by which the controller container can
+query the bundle from the bundle container, that necessitates an increase in the attack surface of
+trust-manager as an overall application.
+
+The communication channel between the controller and bundle containers is absolutely security critical;
+if a malicious actor were able to intercept and modify the bundle in-flight it would allow them to insert
+their own root certificate and MitM every connection between any container using the resulting bundle.
+
+This is mitigated by the use of two containers in a Pod - the Kubernetes security model should allow
+for this to be done securely without interference. Ultimately any design is likely to come down to
+sharing files, and if we can't securely share files between our containers then any reasonable threat
+model would break down anyway.
+
+## Design Details
+
+### Test Plan
+
+End-to-end tests should be written which use public bundles in a trust object, and ensure that the
+bundle is correctly inserted into the output.
+
+### Upgrade / Downgrade Strategy
+
+For an implementation of this design to be acceptable, it should ensure that it's _always_ trivial to
+upgrade and downgrade between versions of the bundle container. That implies that the API between
+the bundle container and trust container should be absolutely ironclad with regards to backwards
+compatibility.
+
+In this design, the bundle is kept as simple as possible by having it be a field ("bundle") in a JSON
+file. As long as future versions of trust-manager are able to read JSON files, the data should be available.
+
+## Alternatives
+
+### Bundling Packages Into trust-manager
+
+The obvious alternative here is that the public trust bundle be embedded into trust-manager itself.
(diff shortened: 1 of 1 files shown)


design/20220722-publicbundle.md
```diff
@@ -0,0 +1,289 @@
...
+apiVersion: trust.cert-manager.io/v1alpha1
+kind: Bundle
+metadata:
+  name: my-org.com
+spec:
+  sources:
+  - configMap:
+      name: "my-root-cm"
+      key: "root-cert.pem"
+  - publicBundle:
```
JoshVanL, line 119: It may be the case that in future that it is not only publicly trusted bundles which are in the sidecar, but other ones as well. For that reason, I think it makes sense to rename this field to something like `sideCar`, `wellKnown` or `imported` (just some ideas).

design/20220722-publicbundle.md
```diff
@@ -0,0 +1,289 @@
...
+      source: "cert-manager" # i.e. use the cert-manager bundle
```
JoshVanL, line 120: I'd rather we be more specific, and use `mozilla` or `cert-manager/mozilla`.

JoshVanL, line 120:
```suggestion
      name: "cert-manager" # i.e. use the cert-manager bundle
```

`source` is a stutter from the `sources` block.

design/20220722-publicbundle.md
```diff
@@ -0,0 +1,289 @@
...
+	"pemBundle": "<pem-encoded-trust-bundle",
+	"bundleVersion": "some version number",
+	"bundleType": "static|dynamic"
+}
+```
+
+`pemBundle` is the bundle itself in PEM format.
+
+`bundleVersion` is the version of the bundle application which served the bundle, which might be
+useful for logging.
```
JoshVanL, line 221: The version will also be essential for writing to the `status` of the `Bundle` object so that a human can verify that trust has propagated a just-released version of a particular bundle.

JoshVanL (dismissed):
Thanks for writing this up @SgtCoDFish  :slightly_smiling_face: 

Left some comments from my side, but it looking good.

design/20220722-publicbundle.md
```diff
@@ -0,0 +1,289 @@
...
+
+## Design Details
+
+### Bundle Container "Server"
+
+The initial version of the bundle container would simply embed the trust bundle into a small Go
+application, similar to what [this PR](https://github.com/cert-manager/trust/pull/36) does.
+
+This Go application would then expose the certificate to the controller pod by writing the bundle,
+to a known location on [a shared volume](https://kubernetes.io/docs/tasks/access-application-cluster/communicate-containers-same-pod-shared-volume/).
```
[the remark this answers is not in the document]
JoshVanL: With unix socket- I wonder if there are some extra authentication checks we could do with things like PID numbers.. thinking out load :slightly_smiling_face:

design/20220722-publicbundle.md
```diff
@@ -0,0 +1,289 @@
...
+remove or change the type of existing fields, the same blob should be usable with any client.
+
+The initial blob might look like this:
+
+```json
+{
+	"pemBundle": "<pem-encoded-trust-bundle",
+	"bundleVersion": "some version number",
+	"bundleType": "static|dynamic"
+}
```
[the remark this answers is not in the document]
SgtCoDFish: Assuming "the name of the bundle" would vary, this significantly complicates the UX of reading this JSON file.

Maybe I'm missing what you're aiming at here, but when I imagine the actual Go code to parse a JSON file which looks like this, it seems hugely error prone and complicated versus simply parsing a struct like this:

```go
type Foo struct {
    PEMBundle string `...`
    BundleVersion string `...`
    BundleType string `...`
```

design/20220722-publicbundle.md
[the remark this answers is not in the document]
SgtCoDFish: > I would rather the two containers talk over a unix socket. This prevents the middle-man file which can be of course hijacked, but also means we can have instance bi-directional communication if either of them need a watch style stream.

How would you share the unix socket without making the socket itself a "middle-man file"? Would a unix socket not normally be presented as a file, itself?

> means we can have instance bi-directional communication if either of them need a watch style stream.

I think we should absolutely rule out any bi-directional communication.

That significantly complicates the design of both the bundle and the controller container, and therefore makes it _much_ harder to ensure that any version of the bundle container would work with any version of the controller.

In addition, it adds a tonne of failure modes and a lot of attack surface for the bundle container since it needs to be able to listen, parse input and respond.

Currently in this design the bundle is simply writing a JSON file to disk, which is trivial to reason about and trivial for the controller to consume. If the controller wants to watch that file for updates that's easy too. Bidirectional systems complicate things a lot, and I'm not sure what we gain from that.

design/20220722-publicbundle.md
SgtCoDFish: > We could also optionally allow users to select the version of this named bundle to use. The default would be the named bundle with the largest version according to SEMVAR.

I think this adds significant complications. The idea presented in this design is that to update the bundle version the entire bundle container would be changed out.

SgtCoDFish: Yeah I'm easy about the name really, it's not important!

design/20220722-publicbundle.md
SgtCoDFish: I think you're maybe envisioning a much bigger and more complex role for the sidecar than I've designed here.

What other bundles would you expect to ship? If we wanted other bundles, could we not ship other sidecars rather than creating one big sidecar?

I think `publicBundle` or something that explicitly references the fact that the bundle is publicly trusted is a big UX win. The action the end user wants is to include a bundle of publicly trusted certificates in their trust bundle - i.e. I think that in a `Bundle` resource they should be specifying what they want, not where it comes from.

design/20220722-publicbundle.md
```diff
@@ -0,0 +1,289 @@
...
+```
+
+`pemBundle` is the bundle itself in PEM format.
+
+`bundleVersion` is the version of the bundle application which served the bundle, which might be
+useful for logging.
+
+`bundleType` would indicate whether the bundle is expected to ever change or whether it should
+be rechecked periodically by the trust controller.
```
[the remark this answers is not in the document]
SgtCoDFish: Streaming would require that we maintain long-lived connections and long-running applications, consuming resources. If we simply write a file we can spend most of our idle time asleep - and the _vast_ majority of the time the bundle container will do absolutely nothing.

design/20220722-publicbundle.md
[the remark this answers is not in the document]
JoshVanL: The struct would end up looking like this. I don’t think it makes things particularly more troublesome

```
type Set struct {
    Name string `…`
    Bundle []struct {
      Version string `...`
      Type string `...`
      PEM string `...`
    } `…`
}
```

design/20220722-publicbundle.md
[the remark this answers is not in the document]
JoshVanL: > How would you share the unix socket without making the socket itself a "middle-man file"? Would a unix socket not normally be presented as a file, itself?

In this context I mean there is no way to write to the file. The trust manager can be sure it’s only talking to one PID.

—-

Bidirectional communication would be required if we want to have “dynamic” side car stores. FS notify is no more or less trivial than listening on a unix socket.

I’m not necessarily is favour of having bidirectionalness either, but the design talks about the source being dynamic, so that was where my head was going.

—-

Regardless, I do think using a unix socket is going to be far less flaky and more reliable to make changes immediately. I also believe it to be less fallible than files on disk since socket connections objectively have more information about the client/server. 

The CSI mechanism in kube is some prior art here which is in the same domain and also uses unix sockets.

I don’t think the API will be any different for either approach.

design/20220722-publicbundle.md
JoshVanL: We should still be able to select which version through the Bundle resourced to allow roll backs or pinning.

design/20220722-publicbundle.md
JoshVanL: Yes, regardless of what the end implementation we choose or the scope, if we intend people to bring their own sidecars containing their corps CA or “public certs but not those signed by the wicked witch”, `publicBundle` is the wrong name for that source, and we should not have to update the trust.cert-manager.io API every time a user wants a specific home grown bundle.

> i.e. I think that in a Bundle resource they should be specifying what they want, not where it comes from.

I disagree here- the source type _is_ the where (or domain) you want to get it from. The keys or names referenced on the type is the what.

design/20220722-publicbundle.md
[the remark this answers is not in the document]
JoshVanL: The sidecar is going to need to be lived long anyway, and a open tcp socket doesn’t really cost anything

design/20220722-publicbundle.md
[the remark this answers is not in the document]
SgtCoDFish: > Bidirectional communication would be required if we want to have “dynamic” side car stores. FS notify is no more or less trivial than listening on a unix socket.

Bidirectional comms explicitly _wouldn't_ be required under the proposed design; there's absolutely no need for the controller container to send any message or data to the bundle container under any circumstances. The controller container is a consumer of whatever file is written by the bundle container.

> I’m not necessarily is favour of having bidirectionalness either, but the design talks about the source being dynamic, so that was where my head was going.

By a dynamic source I mean that the bundle container would write the public bundle with updates after the first time; there's no need for bidirectional comms there since the controller can either use fsnotify or just read the file every X seconds. I think I could make this clearer in the design, which I'll do.

> In this context I mean there is no way to write to the file. The trust manager can be sure it’s only talking to one PID.

As far as I understand, it should be possible with a shared volume to have the bundle container run with (say) UID 2000 and to have the controller container run with (say) UID 1000.  If the bundle container writes the bundle with permissions `0o664`, then anyone can read but only the bundle container would be able to write.

Do you have any links which show how you'd expect the unix socket to be shared between the two containers?

SgtCoDFish:
/test pull-cert-manager-trust-verify

I've updated the guide with proposed workflows for both static and dynamic bundles :+1:

design/20220722-publicbundle.md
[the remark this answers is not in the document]
JoshVanL: As discussed in the standup, here is a very toy example of the PID stuff I was talking about. You can see the general behavior by running `go run .` under `/controller` and `/sidecar` and re-running the sidecar program. 


[sandbox.zip](https://github.com/cert-manager/trust/files/9241127/sandbox.zip)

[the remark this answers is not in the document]
SgtCoDFish: That pretty much matches what I expected it to look like, yeah; I've used unix sockets like that before, locally.

```go
// sidecar/main.go
conn, err := net.Dial("unix", "/tmp/trust.sock")

// controller/main.go
listener, err := net.Listen("unix", "/tmp/trust.sock")            
```

The problem is how we share `trust.sock` between two containers in a pod. This design proposes using a shared volume to share the bundle JSON file. If we replace that JSON file with a unix socket called `trust.sock`, then we'd presumably use a shared volume to contain that socket file.

If we're replacing the JSON with a socket file, we _must_ have **at least** the same security concerns as we'd have with just the JSON blob. If we can't secure the JSON blob we can't secure the socket, because they're both just files.

Is there some other way to share the file between the containers?

I think I'm also missing why we'd actually want to assert the same PID using `SO_PEERCRED`. Apart from PID 1 (which isn't relevant here) I don't know of a reason to treat one PID differently to another, and I don't see any value in asserting that the same PID is used. Why would the controller care how the sidecar operates regarding its PIDs? What threat model are we using where we require protection against the sidecar changing its PID?

(I do see a reason for asserting the same UID is used, but we can let the filesystem do that for us by marking the file `0o664` or `0o644`)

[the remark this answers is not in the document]
SgtCoDFish: Ugggh this is all kinda moot :sob:  k8s `Deployment`s only allow a pod restartPolicy of `Always` so you can't have a container which exits with a success code in a Pod which is part of a Deployment; they _have_ to stay running and "hang" forever even if there's no more work to be done.

Given that the static bundle won't ever update in the lifecycle of a pod, I'd much prefer to not take up any resources after our work is done... but I guess Kubernetes will _force_ us to consume more resources here. At that point, a unix socket which just always returns the same content makes sense, and would at least allow the client to be simpler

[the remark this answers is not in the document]
SgtCoDFish: Actually I think this works reasonably well with minimal resource usage and would respect the pod being shut down by k8s

```go
package main

import (
	"os"
	"os/signal"
	"syscall"
)

func main() {
	// write file

	sigs := make(chan os.Signal, 1)

	signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)
	<-sigs
}
```

design/20220722-publicbundle.md
SgtCoDFish: I'm open to changing; I'd not really thought about people wanting to embed their own sidecars. We _could_ support that, but I'm not really sure why someone would want to do that instead of using a ConfigMap or Secret like we already support. A public trust bundle is a special case IMO.

I went with "packagedBundle" in #46 and kinda half-assumed that people might package their own sidecars.

design/20220722-publicbundle.md
SgtCoDFish: I _really_ don't think we should allow either rollbacks or pinning in a `Bundle`. A roll back should be the person replacing the bundle container with an old version.

If I specify this:

```yaml
kind: Bundle
metadata:
  ...
spec:
  sources:
  - packagedBundle:
      name: "cert-manager"
      version: "2"
  target: ...
```

And version 2 doesn't exist (e.g. we have a static bundle which is at version "3"), what can we meaningfully do except block creation of the target?

For a static bundle the only meaningful operation is to consume the bundle as-is (because there's only one bundle ever).

For a dynamic bundle the only thing I'd want is to consume the latest (because if the user doesn't want that behaviour, why opt into receiving updates in the first place? just use a static bundle and update by hand, surely?)

design/20220722-publicbundle.md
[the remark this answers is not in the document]
SgtCoDFish: I think I'm missing your point here; what do you expect a typical value for "the name of the bundle" to be?

SgtCoDFish:
I've raised #46 as a very rough proof-of-concept for this!

design/20220722-publicbundle.md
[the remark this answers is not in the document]
JoshVanL: Some examples might be “cert-manager.io/Mozilla” or “my.corp.com/prod”.

design/20220722-publicbundle.md
JoshVanL: Yes, we should be blocking a sync of the target when the source doesn’t exist.

Having more than one bundle is important for users who wish to  propagate a trust bundle and don’t want the sources to be accessible via the ConfigMap or Secrets API (and therefore editable).

I never just want to consume the latest- one can imagine scenarios such as CA rotation. A user can safely upgrade all bundle side cars, and then programmatically control the rollout via the API (monitoring the target syncing by acting on the Bundle statuses). A bad rollout can be rolled back in a controlled way by changing the Bundle spec, and not the sidecar container.


There is also the case of multi-tenanted scenarios where teams want control of their destiny- version pinning solves this.

SgtCoDFish: > Yes, we should be blocking a sync of the target when the source doesn’t exist.
> Having more than one bundle is important for users who wish to propagate a trust bundle and don’t want the sources to be accessible via the ConfigMap or Secrets API (and therefore editable).

Yeah I'm convinced by supporting the addition of user-added sidecars down the road as additional sources and I agree about the blocking.

I think the initial version of the implementation of this design should take care to _enable_ the use of additional sidecars but we should keep a focus on providing a first-party public trust bundle sidecar first.

> I never just want to...

As far as I understand it, everything you've described is already solved by the proposal as written without the need for a bundle to support providing multiple versions within itself.

The whole reason the bundles are in a separate container in the first place is so they can be individually versioned separately to the version of the controller.

If I want to do a CA rotation these are the steps with the current design:

1. I start with bundle sidecar v1 with just `oldRoot`
2. I roll out bundle sidecar v2 with both `oldRoot` and `newRoot`; my target configmap is updated automatically
3. I redeploy every pod to ensure it picked up the change
4. I roll out bundle sidecar v3 with just `newRoot`; my target configmap is updated automatically
5. I redeploy every pod to ensure it picked up the change

The hard part at each step is to ensure that every pod is consuming the latest bundle. If there's a problem with sidecar v2 or v3, I roll back to any older version and redeploy every pod.

Adding versions inside the container presumably looks like this (?):

1. I start with bundle sidecar v1 which contains bundle version `A`. All `Bundle` resources set `A` as their target version [1]
2. I roll out bundle sidecar v2 which contains bundle versions `A`, `B` and `C`. Nothing changes yet.
3. I change all my `Bundle` resources to set their target version as `B`. [2]
4. I redeploy every pod to make sure it picked up the change
5. I change all my `Bundle` resources to set their target version as `C`.
6. I redeploy every pod to make sure it picked up the change
7. I roll out bundle sidecar v3 which contains bundle version `C` only
8. I redeploy every pod to make sure it picked up the change

If I detect a problem before step 7 I can roll back by changing all my bundles to point to `A` again, sure, and that's maybe a little easier than rolling back the sidecar container version in some scenarios, but I don't think that gain is worth the huge amount of added complexity and additional failure modes.

> There is also the case of multi-tenanted scenarios where teams want control of their destiny- version pinning solves this.

I don't see any issues here which couldn't be solved by the cluster admin of a multi-tenanted cluster adding additional new sidecars. Would need to see specific examples of a situation where this could go wrong and why version pinning would be required as a solution.

[1] I don't know how we make the fact that it's called `A` discoverable, either; I suspect it ends up being some doc that people have to consult which is outside of k8s, or else they'll need to check the logs of the trust controller or something

[2] There's a huge risk here that Bundles go directly from `A` -> `C` and cause an outage for themselves. That can be gotten around by making sidecar v2 just contain `A` and `B` and then v3 contain `B` and `C`, and then v4 contain just `C`, but that's really quite a brutal upgrade path and it complicates rollbacks too.

design/20220722-publicbundle.md
munnerz: On the surface this sounds to me like we are creating a 'filesystem' source and basically shipping a container containing a trust bundle. I don't see why this sidecar mechanism is specifically related to 'public' bundles. If we are fixing the list of CAs included in here/not making it extensible, I also think it makes more sense to store them in the same container as the binary anyway (what happens if someone runs the controller _without_ the sidecar?).

What's the difference between this and us creating a ConfigMap containing the public CAs that we want to distribute and configuring it by default as a source?

design/20220722-publicbundle.md
```diff
@@ -0,0 +1,326 @@
...
+}
+```
+
+It's a chore to manually specify a location for a trust store, and it's unfamiliar and error-prone
+for many users. However, if a user instead overwrites any of these files with their own
+`trust`-generated certificate trust bundle they'll currently overwrite their public trust bundle
+provided by their OS / container base image.
+
+Unless the user's application only talks to services which the user also controls, what users are
+likely to really want in most cases is a bundle containing both the publicly trusted certs _and_
```
munnerz, line 74: What if I want to talk to two different infra providers services and they both have their own distinct roots? I don't think 'public' should be treated as a special case myself as the reality of PKIs in different environments are more nuanced than a single 'internal' root and then 'public' (cross-company communication is another example)

design/20220722-publicbundle.md
```diff
@@ -0,0 +1,326 @@
...
+
+### Non-Goals
+
+- Editing/pruning the public trust bundle (out of scope for the initial feature but might be desirable later)
+- Supporting [OS-specific](https://superuser.com/questions/437330/how-do-you-add-a-certificate-authority-ca-to-ubuntu) certificate update mechanisms
+- Supporting a plethora of options for sources of public trust bundles; we might want this in the future but we should start with one
+
+## Proposal
+
+We propose that the cert-manager project becomes a distributor of its own public trust bundle, which at
```
munnerz, line 98: I am a very strong -1 on this goal. I don't think we have staffing nor the desire to take on the burden of ensuring this is up to date, notifying users of changes, etc etc. cert-manager has always been a "middle man" and this puts us in an extremely privileged position in many people's infrastructure.

design/20220722-publicbundle.md
SgtCoDFish: > What if I want to talk to two different infra providers services and they both have their own distinct roots?

Then you put the roots of the two providers into configmaps and add them as sources in your `Bundle` resource?

design/20220722-publicbundle.md
SgtCoDFish: >  I also think it makes more sense to store them in the same container as the binary anyway

The reason not to do this is talked about at some length in the design document already.

>  (what happens if someone runs the controller without the sidecar?).

Either we refuse to start the controller or they simply don't get to use a sidecar bundle.

> What's the difference between this and us creating a ConfigMap containing the public CAs that we want to distribute and configuring it by default as a source?

The main difference really is the configmap was my original suggestion which was rejected in the biweekly meeting I presented it at, and this is the compromise that was agreed upon.

design/20220722-publicbundle.md
SgtCoDFish: 🤷 We end up effectively being a distributor of a public trust bundle in basically every proposed design so far; I'm just acknowledging it upfront.

I think we can very easily automate all of this work

>  this puts us in an extremely privileged position in many people's infrastructure.

We're in an extremely privileged position in every single k8s cluster that `trust` or cert-manager is deployed into already.

SgtCoDFish:
Pushed an updated version of the design which more closely matches the version as implemented currently in #46 

Attempts to laser-focus on the problem which I'd expect most users to want to solve, which is to add a single public trust 'package' and to easily refer to it in Bundle resources.

design/20220722-publicbundle.md
```diff
@@ -0,0 +1,279 @@
...
+
+- Editing/pruning the public trust bundle (out of scope for the initial feature but might be desirable later)
+- Supporting [OS-specific](https://superuser.com/questions/437330/how-do-you-add-a-certificate-authority-ca-to-ubuntu) certificate update mechanisms
+- Supporting a plethora of options for sources of public trust bundles; we might want this in the future but we should start with one
+- Dynamically updating the public bundle (users will have to explicitly choose to upgrade, at least for now)
+
+## Proposal
+
+We propose that the cert-manager project becomes a distributor of its own public trust bundle, which at
+least to start will be a carbon copy of the Debian `ca-certificates` package.
```
mt-inside, line 104: It would be great to have blocklisting over this - the issue with the distro trust bundles is they include too much, including the ones that are clearly iffy. I'd like to be able to say "debian, minus foo and bar"

mt-inside, line 104: In future, allow-listing would also be great, eg cert-manager holds all the root certs separately. I can say "defaultPackage" to get all of them, a la Debian, or I can name individual ones - that way I can version-control that config and audit why each root was added.

design/20220722-publicbundle.md
```diff
@@ -0,0 +1,279 @@
...
+
+That bundle in will in turn be packaged into a container which will be used as an init container in the
```
mt-inside, line 106: I think it would be a killer feature to have this dynamically update. For me, a big draw of trust-manager is that it provides more up-to-date certs than an old container image based on an old OS release. Trust-manager could pull a new root cert bundle every day, and cycle Pods that have those certs mounted if there's any changes?

design/20220722-publicbundle.md
SgtCoDFish: Thanks for the comment, really appreciate it!

I fully agree that this would be great functionality for us to add down the road. I'm personally super invested in making trust-manager handle these situations!

SgtCoDFish: Agreed again, good suggestion!

design/20220722-publicbundle.md
SgtCoDFish: Again, I agree that this would be _immensely_ valuable. There are major design questions that need to be solved here though - a big one is "what server does trust-manager query for updated roots from?"

There are tonnes of public places to get that information - debian apt repos, public mozilla bundles, etc - but we'd need to think about how we'd do it safely and without putting an unfair load on the servers. Once every 24h isn't too crazy obviously - just saying this would need further engineering effort.

SgtCoDFish:

Obviously flaking given none of the tests are changed in this PR!

design/20220722-publicbundle.md
mt-inside: Yeah there's a lot of implementation challenges with this, but it'd be amazing to get it working. Maybe something like TUF to solve the trust bootstrap problem? Also agree re the load on the servers hosting the roots, but I'm sure that can be solved - having a little agent that uploads them to a CDN bucket seems like a good start.

PR #45, opened 2022-08-01
Fix copy paste errors, add warning to readme

SgtCoDFish:
See individual commit messages for more context!

Also updates the boilerplate checker.


hack/tools/go.mod
```diff
@@ -0,0 +1,37 @@
+module github.com/cert-manager/csi-driver/hack/tools
```
JoshVanL, line 1:
```suggestion
module github.com/cert-manager/trust/hack/tools
```

JoshVanL (changes requested):
One thing

/assign @SgtCoDFish

hack/tools/go.mod
SgtCoDFish: Ah good spot, cheers! Fixed now :grin:

PR #46, opened 2022-08-02
Bundles with a sidecar container

SgtCoDFish:
This PR implements the current iteration of the proposed bundle sidecar concept with a single default trust package.

Example file which uses the bundle:

```yaml
kind: Bundle
metadata:
  name: trust-bundle
  namespace: default
spec:
  sources:
  - useDefaultCAs: true
  target:
    configMap:
      key: trust.pem
```


SgtCoDFish:
(tests are broken by this but that doesn't really matter because I've not written any new ones! obviously an actual implementation would be more careful and would include a bunch of new tests)

pkg/fspkg/bundle_package.go
```diff
@@ -0,0 +1,89 @@
+/*
```
JoshVanL, line 1: Not a fan of `_` in Go file names. Maybe `parser.go` would be better?

pkg/fspkg/bundle_package.go
```diff
@@ -0,0 +1,89 @@
...
+// of the trust-manager container and subsequently used in a Bundle resource
+type Package struct {
+	// Name contains a friendly name for the bundle
+	Name string `json:"name"`
+
+	// Bundle contains the PEM-formatted certificates which are provided by this bundle.
+	Bundle string `json:"bundle"`
+
+	// Version identifies the bundle's version, to distinguish updated bundles from older counterparts
+	Version string `json:"version"`
```
JoshVanL, line 46: I don't think this is currently being exposed on the `Bundle` status when it has successfully synced. I think that is quite important.

pkg/fspkg/bundle_package.go
```diff
@@ -0,0 +1,89 @@
...
+	if err != nil {
+		return Package{}, fmt.Errorf("failed to read filesystem package '%s': %w", path, err)
+	}
+
+	var pkg Package
+
+	err = json.Unmarshal(data, &pkg)
+	if err != nil {
+		return Package{}, fmt.Errorf("failed to parse package JSON at path '%s': %w", path, err)
+	}
```
JoshVanL, line 70: nit:

```suggestion
		f, err := os.Open(path)
	if err != nil {
		return Package{}, fmt.Errorf("failed to read filesystem package '%s': %w", path, err)
	}

	var pkg Package
	if err := json.NewDecoder(f).Decode(&pkg); err != nil {
		return Package{}, fmt.Errorf("failed to decode filesystem package '%s': %w", path, err)
	}
```

JoshVanL (dismissed):

deploy/charts/trust-manager/templates/deployment.yaml
```diff
@@ -15,6 +15,27 @@ spec:
         app: {{ include "trust-manager.name" . }}
     spec:
       serviceAccountName: {{ include "trust-manager.name" . }}
+      initContainers:
+      - name: cert-manager-bundle-debian
```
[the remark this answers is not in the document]
SgtCoDFish: Yeah good shout - would be better to rename the image as well, while we can!

deploy/charts/trust-manager/README.md
```diff
@@ -33,6 +33,9 @@ A Helm chart for trust-manager
 | app.webhook.port | int | `6443` | Port that the webhook listens on. |
 | app.webhook.service | object | `{"type":"ClusterIP"}` | Type of Kubernetes Service used by the Webhook |
 | app.webhook.timeoutSeconds | int | `5` | Timeout of webhook HTTP request. |
+| bundleImage.pullPolicy | string | `"IfNotPresent"` | Kubernetes imagePullPolicy on Deployment. |
+| bundleImage.repository | string | `"quay.io/jetstack/cert-manager-bundle-debian"` | Repository for bundle image; this might vary based on how you want to handle getting updates for bundles |
+| bundleImage.tag | string | `"20210119.0"` | Tag for bundle image; the format of the tag will vary based on how you want to handle getting updates for bundles |
```
SgtCoDFish, line 38: note: To self: rename "bundleImage" to "defaultPackageImage"

pkg/apis/trust/v1alpha1/types_bundle.go
```diff
@@ -76,6 +76,14 @@ type BundleSource struct {
...
 	// +optional
 	InLine *string `json:"inLine,omitempty"`
+
+	// DefaultPackage, when true, requests the default bundle package to be used as a source.
+	// The default bundle package is available if trust-manager was installed via Helm
+	// or was otherwise set up to include a package-injecting init container.
+	// If the default package was not configured at start-up, a request to use the default
+	// package will fail.
+	// +optional
+	DefaultPackage bool `json:"defaultPackage,omitempty"`
```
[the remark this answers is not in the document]
SgtCoDFish: > What do you think the evolution of the API will be to allow a cluster operator to do that?

I think that would be an entirely different (new) field, added later. At the start of this work I was using something like `FilesystemPackage string` which allowed a user to point at a named package on the filesystem... but it's incredibly awkward to use that because you have to inspect the deployment of trust-manager to work out what container might've written a package, and maybe check logs to see what was written and what it's called. Much worse when it's something that's configured by default.

I think I'm moving away from arbitrary filesystem packages being something that's generally desirable, at least until I see a valid use case where it's an improvement. We're really only using a container to supply a default package because it's easy to change the package while running the same version of trust-manager.

For other non-default "package" types, I can't currently see why I'd want to run a container as a sidecar to trust-manager. I think I'd want a separate pod which writes/maintains ConfigMaps that trust-manager can consume. We can't really do that for default packages but I think that would be preferable for other types. Just IMO.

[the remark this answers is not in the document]
SgtCoDFish: > Fairly sure this need to be a pointer as it is going to default to false and not be omitted in marshelling.

Bool pointers make my heart hurt but I see the reasoning, will change :+1:

pkg/bundle/bundle_test.go
```diff
@@ -624,6 +624,29 @@ func Test_Reconcile(t *testing.T) {
 			),
 			expEvent: "",
 		},
+		"if Bundle references the default package but it wasn't configured at startup, update with error": {
```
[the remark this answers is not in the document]
SgtCoDFish: We can; there's one in sync_test.go! Is there a reason to have another here?

pkg/fspkg/bundle_package.go
SgtCoDFish: I don't like `parser.go` and I don't personally have a problem using `_` in Go file names... but I think `package.go` would be clearer and would remove the underscore, so I'll go for that.

pkg/fspkg/bundle_package.go
SgtCoDFish: Yeah that's a useful review comment, thanks. I'll look into it!

pkg/fspkg/bundle_package.go
```diff
@@ -0,0 +1,89 @@
...
+)
+
+// Package specifies the structure of JSON packages which can be read from a the filesystem
+// of the trust-manager container and subsequently used in a Bundle resource
+type Package struct {
+	// Name contains a friendly name for the bundle
+	Name string `json:"name"`
+
+	// Bundle contains the PEM-formatted certificates which are provided by this bundle.
+	Bundle string `json:"bundle"`
```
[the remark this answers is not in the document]
SgtCoDFish: I think string is a reasonable type for PEM bundles in Go.

I could see the case for why we'd use `[]byte`, but if we wanted to use []byte here we should be using it for the [`InLine` Bundle source](https://github.com/cert-manager/trust-manager/blob/21c839ff1128990e049eaf23000a9a8d6716c89e/pkg/apis/trust/v1alpha1/types_bundle.go#L78) too, for example.

The PEM bundle could contain a comment which is not valid UTF-8, but a string type here will still parse that successfully. Stripping potentially dodgy stuff like that is part of the purpose of https://github.com/cert-manager/trust-manager/pull/75 and I added a test for ValidateAndSanitizePEMBundle loading dodgy 'strings' [too](https://github.com/cert-manager/trust-manager/pull/75/files#diff-d8f68201dac2ff467d25b31d17b81af55029bcf7bb440973f16c229a1b93d391R29).

Main thing for me is consistency. If InLine is string then so should this be. They both expect the same input - PEM bundles.

pkg/webhook/validation.go
```diff
@@ -133,8 +133,14 @@ func (v *validator) validateBundle(ctx context.Context, bundle *trustapi.Bundle)
 				unionCount++
 			}
 
+			if source.DefaultPackage {
+				unionCount++
+			}
+
```
SgtCoDFish, line 139: note to self: courtesy of @inteon - this will change to if `source.DefaultPackage != nil`. What if the user sets DefaultPackage to true in one source and false in another? Consider rejecting in that case.

SgtCoDFish, line 139: note to self: courtesy of @inteon : Consider `useDefaultCAs` instead of `defaultPackage` to try to make it clearer at the point of use what's going on.

See also [this comment](https://github.com/cert-manager/trust-manager/pull/77#issuecomment-1355126726)

pkg/apis/trust/v1alpha1/types_bundle.go
```diff
@@ -76,6 +76,14 @@ type BundleSource struct {
 	// InLine is a simple string to append as the source data.
 	// +optional
 	InLine *string `json:"inLine,omitempty"`
+
+	// DefaultPackage, when true, requests the default bundle package to be used as a source.
```
irbekrm, line 80: I feel like the naming ('DefaultPackage') suggests that these CA certs will be included _by default_ in any Bundle even when this is not specified, which is not the case.
I also think that 'package' is not sufficiently descriptive.

I think that naming this source type in a way that conveys where it actually comes from (CAs compiled by cert-manager in this case) would be more valuable. Perhaps the source type could be named just 'CertManager'? It would also make it easier to refer to it in other contexts by calling it 'cert manager trust source' which is in my opinion more descriptive than 'default package'.

SgtCoDFish: What do you think to the names Tim and I discussed on Friday? E.g. https://github.com/cert-manager/trust-manager/pull/46#discussion_r1050906535

irbekrm: I would be against using a verb like 'include' for the same reason that the 'secret' source is not called 'includeFromSecret'.

But I also think that any version of these source names should be okay given that this is `v1alpha1` API version and we could rename it later if the users complain.

SgtCoDFish: I see your point re verbs, but I think I'm actually kinda OK with using a verb for this specific field because it's a boolean. It wouldn't make sense for `secret` or `configMap`, I agree.

`useDefaultCAs` feels fine though because it's a yes / no (true / false) question. "Do you want to use the default CAs?". It also gets around the idea that the CA certs will be included by default I think.

I'm a little loathe to use something like `certManager` because anyone could configure this to point to any bundle by setting `--default-package-location=/my/bundle.json` - in which case it wouldn't necc. have anything to do with cert-manager. 

- "What are the default CAs if I installed with Helm and didn't change anything?" - cert-manager debian bundle
- "What are the default CAs if I installed with Helm and changed the file it points to?" - Whatever you pointed it at
- "What are the default CAs if I installed manually?" Not available unless you set `--default-package-location`

irbekrm: > because anyone could configure this to point to any bundle by setting --default-package-location=/my/bundle.json

I didn't think about this, thanks 👍🏼 

In that case I do prefer `useDefaultCAs` to `DefaultPackage` because it does not introduce a potentially unneccesary extra term (`package`)

SgtCoDFish: Yeah I think we agree then! I think package makes sense on the backend - the physical JSON bundle can be called a "package" - but at the point of consumption I think "package" is more confusing!

pkg/webhook/validation.go
SgtCoDFish: I think this isn't worth checking for. We already [check](https://github.com/cert-manager/trust-manager/blob/db71205da2310f2d389e18e74dcec9526c9ffa7f/pkg/webhook/validation.go#L106) if multiple fields are set.

If someone chose to explicitly write `defaultPackage: false` alongside a different source, I think we should treat that the same as we would if DefaultPackage were nil.

pkg/bundle/bundle_test.go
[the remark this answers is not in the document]
SgtCoDFish: There is a reason to have another here in a world where we're setting the status to reflect the version of the default package. Will add!

SgtCoDFish:
Still to do:

- rename field in CRD to reflect above discussion
- [nit](https://github.com/cert-manager/trust-manager/pull/46#discussion_r1049759230)

pkg/webhook/validation.go
```diff
@@ -133,6 +134,10 @@ func (v *validator) validateBundle(ctx context.Context, bundle *trustapi.Bundle)
 				unionCount++
 			}
 
+			if source.UseDefaultCAs != nil && *source.UseDefaultCAs {
+				unionCount++
+			}
+
 			if unionCount != 1 {
 				el = append(el, field.Forbidden(
 					path, fmt.Sprintf("must define exactly one source type for each item but found %d defined types", unionCount),
```
inteon, line 143: Does this line mean that we cannot combine eg. Secret + Defaults?

SgtCoDFish: It does in one entry - but you can have multiple entries, and they're all combined at the end.

So you can't have:

```text
- configMap: ...
  useDefaultCAs: true
```

but you can have:

```text
- configMap: ...
- useDefaultCAs: true
```

inteon: ah makes sense

pkg/apis/trust/v1alpha1/types_bundle.go
```diff
@@ -127,6 +138,12 @@ type BundleStatus struct {
 	// Known condition types are `Bundle`.
 	// +optional
 	Conditions []BundleCondition `json:"conditions,omitempty"`
+
+	// DefaultCAPackageVersion, if set and non-empty, indicates the version information
+	// which was retrieved when the set of default CAs was requested in the bundle
+	// source. This should only be set if useDefaultCAs was set to "true" on a source,
+	// and will be the same for the same version of a bundle with identical certificates.
+	DefaultCAPackageVersion *string `json:"defaultCAVersion,omitempty"`
```
inteon, line 146: I'm not sure about reporting this version here, we also don't do this for secret/ configmap sources; would it make sense to just remove this status field?

SgtCoDFish: No, this is absolutely required IMO and was added in response to @JoshVanL 's previous review [here](https://github.com/cert-manager/trust-manager/pull/46#discussion_r1049753380).

The question this answers is "what version of the source was used when creating the target?"

For a ConfigMap / Secret, you can always look at the configmap or secret to find out the contents. In fact, trust-manager can't help with that - it doesn't know what the contents are until it reads for itself and doesn't store the contents once the target is written.

For InLine you look at the Bundle resource `spec` directly.

For DefaultCAs there's no easy way to answer that question without this status field. The other source would either be trust-manager logs or by inspecting the way trust-manager was set up (e.g. by looking at values which were set when the helm chart was installed).

We need a single place where someone with access to read a Bundle resource can look to determine whether that Bundle's DefaultCAs are out of date because it's critical that when the defaultCAs are out of date, they can be updated quickly. I think this status field is the best place for that.

pkg/webhook/validation.go
```diff
@@ -133,6 +134,10 @@ func (v *validator) validateBundle(ctx context.Context, bundle *trustapi.Bundle)
 				unionCount++
 			}
 
+			if source.UseDefaultCAs != nil && *source.UseDefaultCAs {
+				unionCount++
+			}
```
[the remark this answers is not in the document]
SgtCoDFish: We can; I suppose without that constraint a user could request the default CAs be included twice and it would be respected. I can't think why anyone would want that though, so it makes sense to limit and we can always relax later if needed.

Will change!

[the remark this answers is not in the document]
SgtCoDFish: This is now implemented along with a couple of extra validation tests!

pkg/apis/trust/v1alpha1/types_bundle.go
inteon: I mean we could include the generation of the secret configmap and the resource itself in the status.

inteon: Also, out of curiosity: when would there be a situation in which the secret defaultCAs are not up-to-date? Why would the defaultCAs be out-of-date and the secret/ configmap not?

SgtCoDFish: > I mean we could include the Kubernetes Resource Generation of the secret configmap and the resource itself in the status.

We could, but it's significantly less valuable than adding the defaultCAs version here. I can see how it'd help diagnose if a Secret/CM was updated but the Bundle which references that Secret/CM didn't pick up the changed source. But that would be a feature for a different PR in any case.

> Also, out of curiosity: when would there be a situation in which the secret defaultCAs are not up-to-date? Why would the defaultCAs be out-of-date and the secret/ configmap not?

Secrets/CMs/InLine are all user-controller and user-specified. They're out of date if the user says they're out of date. We can't make that call - a Bundle uses whatever it's pointed at. If the user updates a Secret/CM/InLine, then the Bundle controller should update ASAP with the latest available update.

DefaultCAs are different - they're configured at the time when trust-manager is installed (likely via helm's `defaultPackageImage.tag`). They're out of date if there's a new version of the default package available at https://quay.io/repository/jetstack/cert-manager-package-debian and there are only new versions published there if Debian publish new versions of their `ca-certificates` package.

Put another way: The update cycle is different. For Secret/CM/InLine, users update some Secret/CM/InLine in their cluster. For defaultCAs, users update trust-manager to point at a new default CA package.

inteon: @SgtCoDFish can you consider this alternative approach and let me know what you think:
https://github.com/SgtCoDFish/cert-manager-trust-manager/pull/1
(I would prefer not using a new struct field specifically for reporting this one version)

SgtCoDFish:
After discussing with @inteon this morning, we agreed to remove the Type field from fspkg, which I'll do now

inteon:
@SgtCoDFish Thank you for the change, this small new feature will bring a lot of value to our `trust-manager` users.
It is a very minimal solution that should solve most use cases.
/lgtm
/approve

SgtCoDFish:
Thank you so much @inteon and others for the reviews!

NB: We shouldn't release a new version of trust-manager until I've made a few more changes. I want to ensure that trust-package containers we release are tested more carefully, and I want to remove the `type` field from https://quay.io/repository/jetstack/cert-manager-package-debian?tab=tags&tag=latest before anyone starts using it!

PR #47, opened 2022-08-03
add restrictive securitycontext for controller pod

SgtCoDFish:
Currently if we run `make e2e-setup-certmanager` in a cert-manager checkout, we'll install kyverno which will insist on a restrictive security context. That makes a lot of sense, and there's no reason we couldn't use the same restrictive context here.

This also adds `readOnlyRootFilesystem: true` which seems to work fine locally and seems a worthwhile thing to set early if we possibly can!


JoshVanL:
@SgtCoDFish probably worth updating the Chart version, and publishing a release.

SgtCoDFish:
> probably worth updating the Chart version, and publishing a release.

I think that could be a separate PR, but I don't really see a reason to release just for this!

PR #49, opened 2022-08-08
WIP: Move tests to ginkgo v2

SgtCoDFish:
This is a solid base to build on but I've not finished it completely or got confident about it. I think it does the same things as v1, but the config could use some work

- Fixes structural errors relating to `Context` which are warned on by ginkgo v2 (`Context` apparently shouldn't appear inside an `It`)


SgtCoDFish:
WIP because software is terrible!

PR #50, opened 2022-08-09
Update go to v1.19

JoshVanL:
Signed-off-by: joshvanl <me@joshvanl.dev>

Updates go modules and Docker image golang version to v1.19

/hold
Until [jetstack/testing](https://github.com/jetstack/testing/tree/master/config/jobs) has been updated

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 # Build the trust binary
-FROM docker.io/library/golang:1.18 as builder
+FROM docker.io/library/golang:1.19 as builder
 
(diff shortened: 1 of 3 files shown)


irbekrm (approved):
I guess it feels quite early to adopt Go 1.19 as it will probably have some bugs that will get discovered later, see i.e how Kubernetes had to put off release when a major bug was discovered in Go 1.18 that they had adopted very soon after it was released https://groups.google.com/a/kubernetes.io/g/dev/c/dYklbmFPZcw

But trust is experimental, so it feels less risky 👍🏼 

/lgtm

PR #51, opened 2022-08-09
Use `FROM scratch` for base image

JoshVanL:
Signed-off-by: joshvanl <me@joshvanl.dev>

/assign @SgtCoDFish

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,9 +30,7 @@ RUN go mod download
 # Build
 RUN make build
 
-# Use distroless as minimal base image to package the manager binary
-# Refer to https://github.com/GoogleContainerTools/distroless for more details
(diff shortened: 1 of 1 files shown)


SgtCoDFish:
I don't have the bandwidth to be assigned to things, sorry

Dockerfile
```diff
@@ -30,9 +30,7 @@ RUN go mod download
 # Build
 RUN make build
 
-# Use distroless as minimal base image to package the manager binary
-# Refer to https://github.com/GoogleContainerTools/distroless for more details
-FROM gcr.io/distroless/static@sha256:bca3c203cdb36f5914ab8568e4c25165643ea9b711b41a8a58b42c80a51ed609
+FROM scratch
```
irbekrm, line 33: Question: do I understand it correctly that this is to avoid inclusion of CA certs as we'll be consuming them from another source? I've probably missed some discussions around this

irbekrm: makes sense 👍🏼

PR #57, opened 2022-09-22
Update Eventually timeout to be 10s

JoshVanL:
The current timeout for this Eventually check is 1s. This is very prohibitive and often causes false negatives in CI runs which is a resource constrained environment.


inteon (approved):
/lgtm

PR #61, opened 2022-10-17
#14 Updating the apiVersion to v2

joscabmar:
This should solve #14 

Signed-off-by: Jose Antonio <jacm.legend@gmail.com>

PR #67, opened 2022-11-03
Release v0.3.0

inteon:
Update all versions (Helm chart version & app version) to v0.3.0.


JoshVanL:
@inteon I think it will be cleaner for #61 to remain separate

PR #71, opened 2022-11-23
Add a static bundle sidecar container build 

SgtCoDFish:
This only adds the build process for the bundle image so we can start generating these images regularly, ahead of creating a trust-manager release which actually uses the images.

The intention here is that we'll make a PR to https://github.com/jetstack/testing which will add a regular periodic to check if the latest image version available is the same as the latest `ca-certificates` version available in Debian.

That job will run `make ci-update-debian-bundle`, but this PR adds several targets relating to image building jobs we might need.

It's also proposed that if we're happy with `make/image-bundle-debian.mk` as a method of building, we could use that sort of approach for building and releasing trust-manager and other "sideprojects" where there's no defined repeatable build process.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,8 +20,11 @@ WORKDIR /workspace
 COPY go.mod go.mod
 COPY go.sum go.sum
 
-# Copy the go source files
+# Copy build scripts
 COPY Makefile Makefile
+COPY make/ make/
+
+# Copy the go source files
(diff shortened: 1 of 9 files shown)


SgtCoDFish:
/hold

Working on better ways of building this.

SgtCoDFish:
Closing for now since I might take a different approach

PR #75, opened 2022-12-07
Add PEM verification function

SgtCoDFish:
This function is intended to validate PEM inputs to ensure that the only contain certificates and that all certificates within are valid.

Also updates tests to use actual dummy PEM certificates rather than junk data. I think hardcoding certs here is fine - for now at least, we don't care if the certs expire in the future - we just care that they're valid PEM certificates.

### DIFF
--- a/pkg/bundle/bundle.go
+++ b/pkg/bundle/bundle.go
@@ -49,10 +49,10 @@ type bundle struct {
 	// client is a Kubernetes client that makes calls to the API for every
 	// request.
 	// Should be used for updating, deleting, and when requesting data from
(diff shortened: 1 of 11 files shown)


JoshVanL:
I agree
/lgtm

SgtCoDFish:
Thanks for the review! Looks like I totally missed the smoke test so I'll go back and cure that tomorrow :D

SgtCoDFish:
@JoshVanL looks like the smoke test is sorted - would it be possible for you to give another lgtm? :grin:

PR #76, opened 2022-12-12
Update bundle output and add wait option to copyandpause

SgtCoDFish:
- Bundles are now written without the version in their name.

- Bundles have their names added as fields in the JSON

- copyandpause is now copyandmaybepause, with pausing being an optional choice. This is to enable its use in init containers.

### DIFF
--- a/bundles/debian/Containerfile
+++ b/bundles/debian/Containerfile
@@ -31,7 +31,7 @@ COPY go.mod go.mod
 COPY go.sum go.sum
 COPY main.go main.go
 
-RUN CGO_ENABLED=0 go build -o copyandpause main.go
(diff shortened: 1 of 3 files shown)


inteon:
Thanks for the PR. Do you think that we might remove the --wait flag in the future?
/approve
/lgtm

SgtCoDFish:
> Thanks for the PR. Do you think that we might remove the --wait flag in the future?

I only added it because I can see the same code being used for dynamic bundles in the future, but the image could be _much_ simpler now we don't need to worry about pausing. I think it'll be a judgement call - if we end up not doing anything with dynamic images, we can simplify the code.

Thanks for the review! :heart:

PR #77, opened 2022-12-16
Rename debian 'bundle' to 'package' / 'trust-package'

SgtCoDFish:
Also add some README files explaining what trust-packages actually are.

This was sparked by work on #46 and should be merged before that PR.

This will also necessitate changes to the builder config in GCP. Before I raised this PR I paused the builder, and I'll modify the builder after this merges.


Makefile
```diff
@@ -78,10 +78,10 @@ image: | $(BINDIR) ## build docker image targeting all supported platforms
 
 .PHONY: kind-load
 kind-load: local-images | $(BINDIR)/kind
-	$(BINDIR)/kind load docker-image $(CONTAINER_REGISTRY)/trust-manager:$(RELEASE_VERSION) $(CONTAINER_REGISTRY)/cert-manager-bundle-debian:latest
+	$(BINDIR)/kind load docker-image $(CONTAINER_REGISTRY)/trust-manager:$(RELEASE_VERSION) $(CONTAINER_REGISTRY)/cert-manager-package-debian:latest$(DEBIAN_TRUST_PACKAGE_SUFFIX)
```
SgtCoDFish, line 81: note: Adding `$(DEBIAN_TRUST_PACKAGE_SUFFIX)` to the end here was something I'd missed previously so I snuck it in here.

inteon:
Thank you for the change, it makes it clear that a trust-manager Bundle and a package are two different things.
In terms of naming, I think we concluded the following:
- Bundle is our CRD and can be used to generate configmaps that contain CAs from different sources
- a package is a set of CA certs that can be used as the default set of CAs by the trust-manager controller
- using the `useDefaultCAs: true` source, we can include this default set of CAs in the configmaps generated for a Bundle
/lgtm
/approve

PR #78, opened 2022-12-18
fix broken link and old name references

hdeadman:
Signed-off-by: Hal Deadman <hdeadman@users.noreply.github.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,12 +15,12 @@ trust-manager is designed to complement
 trust X.509 certificates signed by Issuers, as well as external CAs which may
 not be known to cert-manager at all.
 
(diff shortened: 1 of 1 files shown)


SgtCoDFish (approved):
/lgtm
/approve
/ok-to-test

Thank you for this, always great to keep this stuff up to date @hdeadman :tada: 

Note for future reviewers: The link wasn't actually _broken_ as far as I can tell - it was out of date. I'm noting this because we have [redirect rules in place](https://github.com/cert-manager/website/blob/dc77e81b56523812040befcc487ff5a009a114a2/public/_redirects#L186-L187) for the old URL and if the link had been broken it would have been a sign of some wider breakage!

SgtCoDFish:

Obvious flake!

PR #79, opened 2022-12-19
Pass GOPROXY to build environment

SgtCoDFish:
This will enable devs with a custom GOPROXY to leverage it when building + testing trust-manager!

NOTE: I couldn't see a way to do this without build args using `docker buildx`. Maybe I missed the equivalent of `docker build -e GOPROXY` but in any case I can see why they maybe wouldn't want that for buildx.

### DIFF
--- a/trust-packages/debian/Containerfile
+++ b/trust-packages/debian/Containerfile
@@ -25,13 +25,15 @@ RUN /work/build.sh $EXPECTED_VERSION $VERSION_SUFFIX /work/package.json
 
 FROM docker.io/library/golang:1.19 as gobuild
 
+ARG GOPROXY
+
 WORKDIR /work
 
 COPY go.mod go.mod
 COPY go.sum go.sum
 COPY main.go main.go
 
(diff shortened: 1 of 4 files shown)


trust-packages/debian/Containerfile
```diff
@@ -25,13 +25,15 @@ RUN /work/build.sh $EXPECTED_VERSION $VERSION_SUFFIX /work/package.json
...
+ARG GOPROXY
+
 WORKDIR /work
 
 COPY go.mod go.mod
 COPY go.sum go.sum
 COPY main.go main.go
 
-RUN CGO_ENABLED=0 go build -o copyandmaybepause main.go
+RUN GOPROXY=$GOPROXY CGO_ENABLED=0 go build -o copyandmaybepause main.go
```
SgtCoDFish, line 36: note: Currently this is redundant because there aren't any deps for `copyandmaybepause`, but I figured I'd throw it in there anyway.

PR #82, opened 2022-12-21
Various makefile updates / improvements

SgtCoDFish:
- Always verify helm docs in lint
- Improve helm install method to be more "make"-y, and bump helm version to latest
- Bump kubebuilder tools to latest
- Move tests with external dependencies out of `./pkg` (as we did in cert-manager), so that `go test ./pkg/...` works on a clean checkout
- Unifies all of the build steps for trust-manager as we did for the Debian trust package. This also adds a build target for pushing images


irbekrm (approved):
Thanks @SgtCoDFish , I've read through this and it all makes sense

I guess it would also be good at some point to document the process for releasing as well as for building this locally/running tests, but that doesn't have to be this PR

/lgtm

SgtCoDFish:
> I guess it would also be good at some point to document the process for releasing as well as for building this locally/running tests, but that doesn't have to be this PR

You've read my mind :wink:  Gonna write those docs and chuck them on the cert-manager website's contributing section I think!

PR #84, opened 2022-12-22
Fix logic error in trust-manager tests

SgtCoDFish:
Before this change, the function was checking if a bundle had synced to the first namespace returned and not checking if all namespaces had synced the bundle

Spotted while refactoring tests more widely but this seems too important to leave!


JoshVanL (approved):
Whoopa!

/lgtm

PR #85, opened 2022-12-22
Upgrade to ginkgo v2 + refactor several aspects of integration tests

SgtCoDFish:
Also see commit messages on individual commits for more details!

- Upgrade to ginkgo v2, so we only have one version in go.mod
  Also improves the install of ginkgo so it's easier to update to different versions

- Context is not valid inside If, so remove its use
  We can use code comments or a descriptive message in Should() instead

- Remove use of Eventually except in ginkgo tests / specific functions
  We can't easily catch errors from Eventually and format them nicely,
  so it's helpful to use standard Golang errors where possible and then
  only have Eventually at the top level

- Add CheckBundleSynced in favour of BundleHasSynced
  Golang errors allow us to provide more context of failures at the
  actual point of failure than a boolean would allow.

- Don't log "expected / got" for PEM certificates in HasSynced funcs
  The HasSynced functions would log output when certificates didn't
  match, which was very noisy and was difficult to compare in any
  case. Prefer to log the last state when the Eventually actually fails,
  since we'd expect at least a few calls to fail during an Eventually

- Temp: Don't dump PEM data when Eventually...HasSynced... fails
  Related to the above; output for "expected / got" contained PEM
  certificates which are difficult to compare by eye. The above
  removed the spamminess of that output, but it would still be helpful
  to be able to compare the last observed state of "expected / got" when
  the call to Eventually actually fails. For now, we don't do that since
  it's very difficult to compare by eye; we'll add a function to parse PEM
  output and dump it usefully, later.

- Disable colors in CI
  Colors aren't visible in prow output so there's no reason to use them.
  We can check that we're in CI by checking the CI variable.

- Disable verbose output when not in CI
  Verbose output causes every log line for every test to be printed, and
  the controller-runtime-managed test environment can be very noisy.
  With verbose output disabled, log lines are printed for a test only if
  that test actually failed.
  We use verbose logs in CI to aid debugging if something goes wrong.
  It's easy to enable verbose logs either in the ginkgo CLI or by
  setting the CI environment variable locally

- Use a constant for Eventually poll interval
  This makes it clearer what "100ms" refers to

- Use Makefile to set up smoke test environment
  Doesn't use a second docker build in the cluster setup script, doesn't
  recreate the cluster each time or reinstall cert-manager each time.
  Instead, always reinstalls trust-manager, and uses the kind-load
  target to create and load the images.


SgtCoDFish:
At the time of writing, this pushes the duration of the test from ~2-3m to ~25m, because this now locally builds all images for all platforms to check that they can be built successfully.

I think that's unacceptably large for an increase. Since the smoke tests build local images, they are at least built for one platform. I've reverted to a local `go build` in the `verify` target

Makefile
```diff
@@ -56,6 +63,9 @@ integration-test: depend  ## runs integration tests, defined as tests which requ
 .PHONY: lint
 lint: vet verify-boilerplate verify-helm-docs
 
+.PHONY: verify
+verify: depend build test ## tests and builds trust-manager
```
inteon, line 67: should CI not run `verify-boilerplate` and `verify-helm-docs` too?

SgtCoDFish: It does, I think! `verify` depends on `test` which depends on `lint`, which runs both `verify-boilerplate` and `verify-helm-docs`

inteon: Ah! yes thank you

test/env/data.go
```diff
@@ -179,3 +184,79 @@ func BundleHasSyncedAllNamespaces(ctx context.Context, cl client.Client, name, e
...
+			continue
+		}
+
+		if err := CheckBundleSynced(ctx, cl, name, namespace.Name, expectedData); err != nil {
+			errs = append(errs, fmt.Errorf("namespace %q: %w", namespace.Name, err).Error())
+		}
+	}
+
+	if len(errs) > 0 {
+		return fmt.Errorf("bundle has not synced all namespaces. namespaces with errors: %s", strings.Join(errs, " | "))
```
inteon, line 247: instead of joining the strings, can we use `utilerrors "k8s.io/apimachinery/pkg/util/errors"` here to create a combined error & wrap that error using `fmt.Errorf`?

test/env/data.go
```diff
@@ -179,3 +184,79 @@ func BundleHasSyncedAllNamespaces(ctx context.Context, cl client.Client, name, e
...
+	var errs []string
+
+	for _, namespace := range namespaceList.Items {
+		// Skip terminating namespaces since Bundle won't be synced there
+		if namespace.Status.Phase == corev1.NamespaceTerminating {
+			continue
+		}
+
+		if err := CheckBundleSynced(ctx, cl, name, namespace.Name, expectedData); err != nil {
+			errs = append(errs, fmt.Errorf("namespace %q: %w", namespace.Name, err).Error())
```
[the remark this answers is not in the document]
SgtCoDFish: I went that route first, but I think it's worse; makes it a pain to use `strings.Join(errs, " | ")` below.

test/env/data.go
```diff
@@ -179,3 +184,79 @@ func BundleHasSyncedAllNamespaces(ctx context.Context, cl client.Client, name, e
 
 	return true
 }
```
inteon, line 186: Can we remove BundleHasSyncedAllNamespaces and use EventuallyBundleHasSyncedAllNamespaces instead everywhere?

test/env/data.go
SgtCoDFish: I don't think that's functionally different to the current implementation. It might be a little more "future proof" but I'm not really anticipating that this function would need massively changing in the future.

I don't know what we'd really gain from wrapping these errors, because we're not _realistically_ going to be digging into their context, IMO.

I think I disagree with the suggestion, but I'll do it anyway!

test/env/data.go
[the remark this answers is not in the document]
SgtCoDFish: Resolving in favour of this discussion: https://github.com/cert-manager/trust-manager/pull/85#discussion_r1061529715

test/env/data.go
SgtCoDFish: Yeah, makes sense to do that here!

Makefile
```diff
@@ -116,12 +124,47 @@ clean: ## clean up created files
 		_artifacts
 
 .PHONY: demo
-demo: depend ## create cluster and deploy trust
-	REPO_ROOT=$(shell pwd) ./hack/ci/create-cluster.sh
+demo: ensure-kind kind-load ensure-cert-manager ensure-trust-manager $(BINDIR)/kubeconfig.yaml  ## ensure a cluster ready for a smoke test or local testing
```
inteon, line 127: I like that the demo environment is used in our smoke tests.

test/env/data.go
```diff
@@ -179,3 +184,79 @@ func BundleHasSyncedAllNamespaces(ctx context.Context, cl client.Client, name, e
...
+
+// CheckBundleSynced returns nil if the given Bundle has synced the expected data to the given
+// namespace, or else returns a descriptive error if that's not the case.
+// - Skips Namespaces that are Terminating since targets are not synced there.
+// - Ensures the Bundle status has been updated with the appropriate target.
+// - Ensures the Bundle has the correct status condition with the same ObservedGeneration as the current Generation.
+func CheckBundleSynced(ctx context.Context, cl client.Client, bundleName string, namespace string, expectedData string) error {
```
inteon, line 193: This function looks a LOT like `BundleHasSynced` are there any differences, can we merge the two functions?

SgtCoDFish: Difference is that it returns an error so that failures can be passed to callers in the normal way. Means that if the `CheckBundleSynced` fails when called inside an `Eventually` and it times out, you see where it went wrong.

Totally fair that it's very similar; I'll remove `BundleHasSynced` and any uses of it.

inteon (changes requested):
I added some suggestions for possible improvements

inteon:
Wow, thank you for improving these integration tests.
/approve
/lgtm

SgtCoDFish:
> Wow, thank you for improving these integration tests.

Thank you for reviewing, much appreciated :heart:

test/env/data.go
SgtCoDFish: Interestingly, something related to this is coming in go 1.20!

https://tip.golang.org/doc/go1.20#errors

PR #86, opened 2023-01-05
Integration test tweaks

SgtCoDFish:
See commit messages for more details.

1. Disables leader election (as discussed in standup 2023-01-05)
2. Adds a channel to wait on when shutting down integration test env
3. Uses ginkgo logger explicitly

1 and 2 inspired by: https://github.com/cert-manager/approver-policy/pull/166

### DIFF
--- a/test/integration/bundle/suite.go
+++ b/test/integration/bundle/suite.go
@@ -23,10 +23,12 @@ import (
 	. "github.com/onsi/gomega"
 	. "github.com/onsi/gomega/gstruct"
 
+	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
 	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/klog/v2/ktesting"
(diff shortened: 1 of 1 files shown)


SgtCoDFish:
In draft because I can still trigger flakes locally with this PR

inteon:
I think the leader election is only useful when running ginkgo with > 1 node; since that would result in multiple tests using the same envtest cluster in parallel.

inteon:
I added an additional check to make sure that no two controllers are running at the same time.

inteon:
I made some minor changes to make sure that we can safely disable the leader election.
/approve
/lgtm

Makefile
```diff
@@ -128,7 +128,7 @@ demo: ensure-kind kind-load ensure-cert-manager ensure-trust-manager $(BINDIR)/k
 
 .PHONY: smoke
 smoke: demo  ## ensure cluster, deploy trust-manager and run smoke tests
-	${BINDIR}/ginkgo -nodes 1 test/smoke/ -- --kubeconfig-path ${BINDIR}/kubeconfig.yaml
+	${BINDIR}/ginkgo -procs 1 test/smoke/ -- --kubeconfig-path ${BINDIR}/kubeconfig.yaml
```
SgtCoDFish, line 131: note: these are aliases

inteon: thanks, for the info

PR #87, opened 2023-01-05
Experimental: Use different envtest Environment for each test

SgtCoDFish:
This is #86 but also uses a new environment for every test.

### DIFF
--- a/test/integration/bundle/integration.go
+++ b/test/integration/bundle/integration.go
@@ -29,12 +29,11 @@ import (
 )
 
 var (
-	env *envtest.Environment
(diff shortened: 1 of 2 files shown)


SgtCoDFish:
For reference:

This branch _still_ has flakes of the same kind as we see in #86 which would seem to imply that the issue is not related to long-running Managers.

There's a performance penalty from doing it this way, too - it increases runtime for the integration tests from ~12s to ~85s.

SgtCoDFish:
I still saw flakes using this, and it was significantly slower. Closing this since it's such a performance hit.

PR #88, opened 2023-01-06
Remove type from trust packages + add validation on package build

SgtCoDFish:
Also adds a validator for trust packages, which we'll use to ensure we don't publish packages which trust-manager won't be able to load.

This PR doesn't actually _use_ the validator yet - that'll come in a separate PR to solve the chicken and egg problem (we want the build for the trust-package to happen in a container, but don't want to share anything from the local environment with that container. So we'll `go install github.com/cert-manager/trust-manager/cmd/validate-trust-package@latest` during the build of the trust package, but to do that we need to publish the build for that first)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -166,6 +166,12 @@ ifneq ($(strip $(CI)),)
 	sleep 2
 endif
 
+.PHONY: build-validate-trust-package
+build-validate-trust-package: $(BINDIR)/validate-trust-package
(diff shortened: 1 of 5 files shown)


wallrj:
/lgtm

Thanks @SgtCoDFish  for explaining the motivation for this change during the cert-manager standup meeting.

PR #89, opened 2023-01-06
Use SSA

inteon:
This PR introduces SSA to update the bundles and the configmaps.

The tests are a bit hard to fix since the fakeclient does not support faking SSA.

### DIFF
--- a/cmd/trust-manager/app/app.go
+++ b/cmd/trust-manager/app/app.go
@@ -23,6 +23,8 @@ import (
 
 	"github.com/spf13/cobra"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/selection"
 	"k8s.io/client-go/kubernetes"
 	clientv1 "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/tools/record"
@@ -113,6 +115,28 @@ func NewCommand() *cobra.Command {
 				return fmt.Errorf("failed to create manager: %w", err)
 			}
 
+			targetCache, err := cache.New(mgr.GetConfig(), cache.Options{
+				HTTPClient:                  mgr.GetHTTPClient(),
+				Scheme:                      mgr.GetScheme(),
+				Mapper:                      mgr.GetRESTMapper(),
+				ReaderFailOnMissingInformer: true,
+				DefaultLabelSelector: func() labels.Selector {
+					targetRequirement, err := labels.NewRequirement(trustapi.BundleLabelKey, selection.Exists, nil)
+					if err != nil {
+						panic(fmt.Errorf("failed to create target label requirement: %w", err))
+					}
+
+					return labels.NewSelector().Add(*targetRequirement)
+				}(),
+			})
+			if err != nil {
+				return fmt.Errorf("failed to create target cache: %w", err)
+			}
+
+			if err := mgr.Add(targetCache); err != nil {
+				return fmt.Errorf("failed to add target cache to manager: %w", err)
+			}
+
 			// Add readiness check that the manager's informers have been synced.
 			mgr.AddReadyzCheck("informers_synced", func(req *http.Request) error {
 				if mgr.GetCache().WaitForCacheSync(req.Context()) {
@@ -124,7 +148,7 @@ func NewCommand() *cobra.Command {
 			ctx := ctrl.SetupSignalHandler()
 
 			// Add Bundle controller to manager.
-			if err := bundle.AddBundleController(ctx, mgr, opts.Bundle); err != nil {
+			if err := bundle.AddBundleController(ctx, mgr, opts.Bundle, targetCache); err != nil {
 				return fmt.Errorf("failed to register Bundle controller: %w", err)
 			}
 
--- a/pkg/bundle/bundle.go
+++ b/pkg/bundle/bundle.go
@@ -27,12 +27,16 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/types"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/utils/clock"
+	"k8s.io/utils/ptr"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	trustapi "github.com/cert-manager/trust-manager/pkg/apis/trust/v1alpha1"
+	"github.com/cert-manager/trust-manager/pkg/bundle/internal/ssa_client"
 	"github.com/cert-manager/trust-manager/pkg/fspkg"
 )
 
@@ -57,6 +61,10 @@ type bundle struct {
 	// a cache-backed Kubernetes client
 	client client.Client
 
+	// targetCache is a cache.Cache that holds cached ConfigMap
+	// resources that are used as targets for Bundles.
+	targetCache client.Reader
+
 	// defaultPackage holds the loaded 'default' certificate package, if one was specified
 	// at startup.
 	defaultPackage *fspkg.Package
@@ -69,12 +77,39 @@ type bundle struct {
 
 	// Options holds options for the Bundle controller.
 	Options
+
+	// patchResourceOverwrite allows use to override the patchResource function
+	// it is used for testing purposes
+	patchResourceOverwrite func(ctx context.Context, obj interface{}) error
 }
 
 // Reconcile is the top level function for reconciling over synced Bundles.
 // Reconcile will be called whenever a Bundle event happens, or whenever any
 // related resource event to that bundle occurs.
 func (b *bundle) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	result, statusPatch, resultErr := b.reconcileBundle(ctx, req)
+	if statusPatch != nil {
+		con, patch, err := ssa_client.GenerateBundleStatusPatch(req.Name, req.Namespace, statusPatch)
+		if err != nil {
+			err = fmt.Errorf("failed to generate bundle patch: %w", err)
+			return ctrl.Result{}, utilerrors.NewAggregate([]error{resultErr, err})
+		}
+
+		if err := b.client.Status().Patch(ctx, con, patch, &client.SubResourcePatchOptions{
+			PatchOptions: client.PatchOptions{
+				FieldManager: fieldManager,
+				Force:        ptr.To(true),
+			},
+		}); err != nil {
+			err = fmt.Errorf("failed to apply bundle patch: %w", err)
+			return ctrl.Result{}, utilerrors.NewAggregate([]error{resultErr, err})
+		}
+	}
+
+	return result, resultErr
+}
+
+func (b *bundle) reconcileBundle(ctx context.Context, req ctrl.Request) (result ctrl.Result, statusPatch *trustapi.BundleStatus, returnedErr error) {
 	log := b.Log.WithValues("bundle", req.NamespacedName.Name)
 	log.V(2).Info("syncing bundle")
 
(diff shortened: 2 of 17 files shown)


pkg/bundle/internal/cache.go
```diff
@@ -126,26 +135,64 @@ func (b *multiScopedCache) WaitForCacheSync(ctx context.Context) bool {
...
+		return err
+	}
+	return b.cacheFromGVK(obj.GetObjectKind().GroupVersionKind()).Get(ctx, key, obj, opts...)
 }
 
 // List returns the underlying cache's List based on resource type.
 func (b *multiScopedCache) List(ctx context.Context, list client.ObjectList, opts ...client.ListOption) error {
+	if err := setGroupVersionKind(b.scheme, list); err != nil {
+		return err
+	}
```
inteon, line 156: Let us merge these cache fixes first (in another PR)

inteon: see https://github.com/cert-manager/trust-manager/pull/90

irbekrm:
Thanks for working on this Tim, overall it makes sense to me to use SSA.

I haven't read through the code thoroughly, but I experimented a little bit it and noticed that (as you already suggested in an earlier standup) it is ineed the case that when
1) a version of trust-manager without SSA update bundle status with some optional field such  as namespace selector
2) a version of trust-manager without SSA applies a patch to bundle status that does not include this field because it should actually be removed
the field ends up not being removed
```
irbe@trust-manager$ k get bundle  my-org.com -oyaml --show-managed-fields
kind: Bundle
metadata:
...
  managedFields:
  - apiVersion: trust.cert-manager.io/v1alpha1
    fieldsType: FieldsV1
    fieldsV1:
      f:status:
        f:conditions: {}
        f:target:
          f:configMap:
            f:key: {}
    manager: trust-manager
    operation: Apply
    subresource: status
    time: "2023-01-11T08:42:32Z"
  - apiVersion: trust.cert-manager.io/v1alpha1
    fieldsType: FieldsV1
    fieldsV1:
      f:status:
        .: {}
        f:target:
          .: {}
          f:configMap:
            .: {}
            f:key: {}
          f:namespaceSelector:
            .: {}
            f:matchLabels:
              .: {}
              f:foo: {}
    manager: trust-manager
    operation: Update
    subresource: status
    time: "2023-01-11T08:35:51Z"
...
spec:
  sources:
  - configMap:
      key: ca.crt
      name: source-ca
  target:
    configMap:
      key: ca.crt
status:
...
  target:
    configMap:
      key: ca.crt
    namespaceSelector:
      matchLabels:
        foo: bar
```

I've now reminded myself that this is indeed the behaviour of SSA https://github.com/kubernetes/kubernetes/issues/99003#issuecomment-1012554047

I see that there now is  some client-go code to take over ownership of managed fields https://github.com/kubernetes/client-go/tree/master/util/csaupgrade. I haven't thought about this in depth, but am wondering whether we could use this to take over ownership of managed fields with SSA. (If it works for us, worth also looking at how expensive it might be to run)

inteon:
@irbekrm thank you for the remark, we can add some migration logic to take ownership of the fields that were created by "manager: trust-manager operation: Update" or just make sure that we are owner of the key defined in the resource status target.

irbekrm:
> we can add some migration logic to take ownership of the fields that were created by "manager: trust-manager operation: Update" or just make sure that we are owner of the key defined in the resource status target.

It feels like the right state to be in for a version of trust-manager that implements SSA would be if `manager: trust-manager operation: Update` was the owner of the whole of status.

pkg/bundle/sync.go
```diff
@@ -145,89 +150,163 @@ func (b *bundle) secretBundle(ctx context.Context, ref *trustapi.SourceObjectKey
...
+	}
+
+	return nil
+}
+
+// MIGRATION: This is a migration function that migrates the ownership of
+// fields from the Update operation to the Apply operation. This is required
+// to ensure that the apply operations will also remove fields that were
+// created by the Update operation.
+func (b *bundle) migrateToApply(ctx context.Context, obj client.Object) error {
```
inteon, line 298: TODO: this function still needs a test

inteon: @irbekrm I added a fix here, could you test if this works?
I haven't added a test yet, if you or someone else want to try to add one, feel free to do so.

irbekrm: Thanks @inteon I tested it this morning with the same strategy as in https://github.com/cert-manager/trust-manager/pull/89#issuecomment-1378452644 - I still see the namespace selector remain on Bundle status after it has been removed from the spec.

FWIW, I think you cannot test it by deploying trust-manager using the make commands as they uninstall every time before reinstalling.
I also had to add 'update' for configmaps to trust's clusterrole to get it working.

pkg/bundle/bundle.go
```diff
@@ -199,7 +199,11 @@ func (b *bundle) reconcileStatusPatch(ctx context.Context, req ctrl.Request) (re
 				Kind:       "ConfigMap",
 			},
 		}
-		err := b.directClient.List(ctx, configMapList, &client.ListOptions{})
+		err := b.directClient.List(ctx, configMapList, &client.ListOptions{
```
irbekrm, line 202: Why do we need to list directly from kube apiserver? We know which configmaps we're interested in in each namespace as the configmap name is static. I think we should get them by name from cache. Else if there is a reason for wanting to list all that are labelled, then probably better to cache all the labelled ones in full (rather than just partial metadata).

inteon: I reworked the caching logic in the latest commit.

pkg/bundle/sync.go
inteon: @irbekrm I was only migrating the ConfigMap resource. I added a migration for the bundle resource too now.

pkg/bundle/bundle.go
```diff
@@ -69,99 +77,82 @@ type bundle struct {
...
 
 // Reconcile is the top level function for reconciling over synced Bundles.
 // Reconcile will be called whenever a Bundle event happens, or whenever any
 // related resource event to that bundle occurs.
 func (b *bundle) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	result, statusPatch, resultErr := b.reconcileStatusPatch(ctx, req)
+	if statusPatch != nil {
+		con, patch, err := ssa_client.GenerateBundleStatusPatch(req.Name, req.Namespace, statusPatch)
+		if err != nil {
+			err = fmt.Errorf("failed to generate connection patch: %w", err)
```
erikgb, line 94:
```suggestion
			err = fmt.Errorf("failed to generate bundle patch: %w", err)
```

pkg/bundle/bundle.go
```diff
@@ -69,99 +77,82 @@ type bundle struct {
...
+			return ctrl.Result{}, utilerrors.NewAggregate([]error{resultErr, err})
+		}
+
+		if err := b.client.Status().Patch(ctx, con, patch, &client.SubResourcePatchOptions{
+			PatchOptions: client.PatchOptions{
+				FieldManager: "trust-manager",
```
erikgb, line 100: Should we have a const for the field manager value? Or use the one in `sync.go`? (reviewing from top to bottom) 😉

pkg/bundle/bundle.go
```diff
@@ -69,99 +77,82 @@ type bundle struct {
...
+				Force:        ptr.To(true),
+			},
+		}); err != nil {
+			err = fmt.Errorf("failed to apply connection patch: %w", err)
```
erikgb, line 104:
```suggestion
			err = fmt.Errorf("failed to apply bundle patch: %w", err)
```

pkg/bundle/bundle.go
```diff
@@ -69,99 +77,82 @@ type bundle struct {
...
+			return ctrl.Result{}, utilerrors.NewAggregate([]error{resultErr, err})
+		}
+	}
+
+	return result, resultErr
+}
+
+func (b *bundle) reconcileStatusPatch(ctx context.Context, req ctrl.Request) (result ctrl.Result, bundlePatch *trustapi.BundleStatus, returnedErr error) {
```
erikgb, line 112:
```suggestion
func (b *bundle) reconcileBundle(ctx context.Context, req ctrl.Request) (result ctrl.Result, statusPatch *trustapi.BundleStatus, returnedErr error) {
```

WDYT? It does a bit more than reconciling status patch?

erikgb (commented):
First review-pass. Still pretty big! Is it possible to do some pre-work to make the test changes easier to review?

pkg/bundle/sync.go
```diff
@@ -259,137 +262,265 @@ func (e pkcs12Encoder) encode(trustBundle string) ([]byte, error) {
...
+	target := bundle.Spec.Target
 	if target.ConfigMap == nil {
 		return false, errors.New("target not defined")
 	}
 
-	matchNamespace := namespaceSelector.Matches(labels.Set(namespace.Labels))
-
-	var configMap corev1.ConfigMap
-	err := b.client.Get(ctx, client.ObjectKey{Namespace: namespace.Name, Name: bundle.Name}, &configMap)
+	dataHash := fmt.Sprintf("%x", sha256.Sum256([]byte(data)))
```
erikgb, line 320: You have maybe considered this already, but working on the PKCS12 feature, I noted this comment - which is deleted in this PR:

> Generated JKS is not deterministic - best we can do here is update if the pem cert has changed (hence not checking if JKS matches)

If this holds, I think we might have a potential reconcile loop form the logic using a hash.

pkg/bundle/sync.go
```diff
@@ -259,137 +262,265 @@ func (e pkcs12Encoder) encode(trustBundle string) ([]byte, error) {
...
+		// prefix which appears on managed field keys.
+		configmapData.Iterate(func(path fieldpath.Path) {
+			properties.Insert(strings.TrimPrefix(path.String(), "."))
+		})
 	}
+
+	return properties, nil
+}
+
+func (b *bundle) patchResource(ctx context.Context, obj interface{}) error {
```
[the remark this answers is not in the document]
inteon: I'm preparing for a word where we also support SecretApplyConfiguration patches.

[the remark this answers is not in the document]
inteon: I'm preparing for a world where we also support SecretApplyConfiguration patches.

pkg/bundle/sync.go
```diff
@@ -259,137 +262,265 @@ func (e pkcs12Encoder) encode(trustBundle string) ([]byte, error) {
...
+		for _, mf := range configMap.ManagedFields {
+			if mf.Manager == fieldManager && mf.Operation == metav1.ManagedFieldsOperationUpdate {
+				needsMigration = true
+				needsPatch = true
+			}
 		}
 
-		needsUpdate = true
+		if needsMigration {
+			if err := b.migrateToApply(ctx, configMap, ptr.To("")); err != nil {
```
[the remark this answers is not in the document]
inteon: I removed the nil case.

pkg/bundle/sync.go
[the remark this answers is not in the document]
erikgb: Nice! SGTM. What about using generics? 😉

pkg/bundle/sync.go
inteon: The hash should only hash the data PEM string.
We only use the hash to determine if the "source of truth" PEM bundle changed.
Any modification to existing PEM contents/ JKS contents/ ... we notice by making sure we are still field owner of the key in the map.

erikgb: Right, `data` contains the PEM only. The `binaryData` field will container JKS/PKCS12. My bad.

cmd/trust-manager/app/app.go
```diff
@@ -113,6 +115,28 @@ func NewCommand() *cobra.Command {
 				return fmt.Errorf("failed to create manager: %w", err)
 			}
 
+			targetCache, err := cache.New(mgr.GetConfig(), cache.Options{
```
inteon, line 118: We need a target cache because we only want to store metadata for the target resources.

pkg/bundle/sync.go
erikgb:
```suggestion
        // Generated JKS is not deterministic - best we can do here is update if the pem cert has
	// changed (hence not checking if JKS matches)
	dataHash := fmt.Sprintf("%x", sha256.Sum256([]byte(data)))
```

I think it's worth keeping this comment for future reference. WDYT?

erikgb (approved):
/lgtm
/hold

Amazing job, thanks! And thanks for the patience! It's been a learning experience for me to see how SSA can be implemented without first-class support in c/r!

Since I don't feel 100% confident in this area, it would be nice if someone else also could review this. But feel free to unhold at wish. 😄

pkg/bundle/bundle_test.go
```diff
@@ -95,15 +111,17 @@ func Test_Reconcile(t *testing.T) {
...
+		baseBundleOwnerRef = []metav1.OwnerReference{*metav1.NewControllerRef(baseBundle, trustapi.SchemeGroupVersion.WithKind(trustapi.BundleKind))}
 
 		namespaces = []client.Object{
 			&corev1.Namespace{TypeMeta: metav1.TypeMeta{Kind: "Namespace", APIVersion: "v1"}, ObjectMeta: metav1.ObjectMeta{Name: trustNamespace}},
 			&corev1.Namespace{TypeMeta: metav1.TypeMeta{Kind: "Namespace", APIVersion: "v1"}, ObjectMeta: metav1.ObjectMeta{Name: "ns-1"}},
 			&corev1.Namespace{TypeMeta: metav1.TypeMeta{Kind: "Namespace", APIVersion: "v1"}, ObjectMeta: metav1.ObjectMeta{Name: "ns-2"}},
 		}
 
-		fixedTime     = time.Date(2021, 01, 01, 01, 0, 0, 0, time.UTC)
+		fixedTime     = time.Date(2021, 01, 01, 01, 0, 0, 0, time.Local)
```
SgtCoDFish, line 124: question: Why change to Local? Does this not introduce the possibility of daylight savings weirdness?

I'd think that everything should be UTC unless there's a really good reason for it not to be!

SgtCoDFish (changes requested):
One question, and there's also the comment I made in standup about whether this should change our supported k8s versions: https://github.com/cert-manager/trust-manager/blob/907b043b845c23a8633eddc34787e37a15368be2/deploy/charts/trust-manager/Chart.yaml#L9

pkg/bundle/bundle_test.go
inteon: If you unmarshall metav1.Time from json, it automatically creates a Local time. (see https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/time.go#L115)
That is why I have to use Local time here, so I can compare this value after the bundel is marshalled and unmashalled in the test.

inteon: Previously, we were comparing json. Now we are comparing go objects (using `assert.Equal`)

SgtCoDFish: Interesting; I wonder if tests should be run with the env var TZ set to ""? Hadn't really considered doing that before... https://pkg.go.dev/time#UTC

In any case, I can see why it's changed to `time.Local` here; not a blocker 😁

SgtCoDFish (approved):

SgtCoDFish:
/lgtm
/hold

We discussed this morning that this should probably merge becuase if this breaks anything it's a great excuse to write more tests.

This looks like it took a lot of work; thank you for your dedication @erikgb and @inteon (and @irbekrm !)

PR #90, opened 2023-01-08
Fix the MultiScopedCache

inteon:
This PR remoes `ClientDisableCacheFor` and fixes the `MultiScopedCache` by populating GVK values when missing based on the type of the provided `client.Object`.

Currently, `ClientDisableCacheFor` disables caching for Secret resources; I reenabled caching for these resources. I also fixed the `MultiScopedCache`, now it can handle *corev1.Secret resources that don't have GVK values set.

### DIFF
--- a/test/integration/bundle/cache_test.go
+++ b/test/integration/bundle/cache_test.go
@@ -1,0 +1,153 @@
+/*
+Copyright 2021 The cert-manager Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package test
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/rest"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	multiscopedcache "github.com/cert-manager/trust-manager/pkg/bundle/cache"
+)
+
+var _ = Describe("Integration test cache", func() {
+	It("should be possible to Get a resource without having cluster-wide List & Watch permissions", func() {
+		ctx, cancel := context.WithTimeout(context.Background(), 20*time.Second)
+		defer cancel()
+
+		namespace := "test-namespace"
+
+		// Create a service account that can only retrieve secrets in a single namespace.
+		var cacheRestConfig *rest.Config
+		{
+			godClient, err := client.New(env.Config, client.Options{})
+			Expect(err).NotTo(HaveOccurred())
+
+			ns := &corev1.Namespace{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: namespace,
+				},
+			}
+			err = godClient.Create(ctx, ns)
+			Expect(err).NotTo(HaveOccurred())
+
+			sa := &corev1.ServiceAccount{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "cache-sa",
+					Namespace: namespace,
+				},
+			}
+			err = godClient.Create(ctx, sa)
+			Expect(err).NotTo(HaveOccurred())
+
+			role := &rbacv1.Role{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "cache-role",
+					Namespace: namespace,
+				},
+				Rules: []rbacv1.PolicyRule{
+					{
+						Verbs:     []string{"list", "watch"},
+						APIGroups: []string{""},
+						Resources: []string{"secrets"},
+					},
+				},
+			}
+			err = godClient.Create(ctx, role)
+			Expect(err).NotTo(HaveOccurred())
+
+			rolebinding := rbacv1.RoleBinding{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "cache-rolebinding",
+					Namespace: namespace,
+				},
+				RoleRef: rbacv1.RoleRef{
+					APIGroup: "rbac.authorization.k8s.io",
+					Kind:     "Role",
+					Name:     "cache-role",
+				},
+				Subjects: []rbacv1.Subject{
+					{
+						Kind:      "ServiceAccount",
+						Name:      "cache-sa",
+						Namespace: namespace,
+					},
+				},
+			}
+			err = godClient.Create(ctx, &rolebinding)
+			Expect(err).NotTo(HaveOccurred())
+
+			// Create a config that uses the service account.
+			cacheRestConfig = rest.CopyConfig(env.Config)
+			cacheRestConfig.Impersonate.UserName = fmt.Sprintf("system:serviceaccount:%s:%s", namespace, "cache-sa")
+			cacheRestConfig.Impersonate.UID = string(sa.UID)
+
+			// Create a secret that the service account can access.
+			secret := &corev1.Secret{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "test-secret",
+					Namespace: namespace,
+				},
+				Data: map[string][]byte{
+					"test": []byte("test"),
+				},
+			}
+			err = godClient.Create(ctx, secret)
+			Expect(err).NotTo(HaveOccurred())
+		}
+
+		newCache := multiscopedcache.NewMultiScopedCache(namespace, []schema.GroupKind{{Group: "", Kind: "Secret"}})
+
+		scheme := runtime.NewScheme()
+		Expect(corev1.AddToScheme(scheme)).NotTo(HaveOccurred())
+		cache, err := newCache(cacheRestConfig, cache.Options{
+			Scheme: scheme,
+		})
+		Expect(err).NotTo(HaveOccurred())
+
+		done := make(chan error)
+		go func() {
+			done <- cache.Start(ctx)
+		}()
+		defer func() {
+			Expect(<-done).NotTo(HaveOccurred())
+		}()
+
+		Expect(cache.WaitForCacheSync(ctx)).To(BeTrue())
+
+		secret := &corev1.Secret{}
+		err = cache.Get(ctx, client.ObjectKey{
+			Namespace: namespace,
+			Name:      "test-secret",
+		}, secret)
+		Expect(err).NotTo(HaveOccurred())
+
+		Expect(secret.Data["test"]).To(Equal([]byte("test")))
+	})
+})
--- a/cmd/app/app.go
+++ b/cmd/app/app.go
@@ -60,13 +60,12 @@ func NewCommand() *cobra.Command {
 			eventBroadcaster.StartLogging(func(format string, args ...interface{}) { mlog.V(3).Info(fmt.Sprintf(format, args...)) })
 			eventBroadcaster.StartRecordingToSink(&clientv1.EventSinkImpl{Interface: cl.CoreV1().Events("")})
 
-			scheme := trustapi.GlobalScheme
 			mgr, err := ctrl.NewManager(opts.RestConfig, ctrl.Options{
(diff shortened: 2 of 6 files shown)


SgtCoDFish (commented):
I'd not really thought much about caching in trust-manager yet, so thank you for this thought provoking PR!

Looking at it, this makes sense and seems to be an improvement in the case that we want caching. But I'm not actually sure caching is worth it right now.

As far as I can see, the absolute worst case scenario in which our cache would help is in the case where we have _n_ different `Bundle` resources, all of which refer to the same `Secret`. 

That would mean that without a cache an update to the `Secret` would trigger _n_ API calls, whereas a cache would reduce that to 1 API call in the best case.

That's an improvement of course but I see some issues:

1. Given that Secrets probably aren't the best type of source for a bundle (since trusted CA data isn't a "secret") the question is 'what about ConfigMaps'. We don't have caching for ConfigMaps (I think?) and they're likely to be more common.
2. Large values of _n_ seem really unlikely to me. I can imagine a multi-tenanted situation where a single trusted CA is shared by many bundles.  But for _n_ to grow large enough that we cause performance problems seems like it would need an _extremely_ large cluster - and I'd imagine that in such a large cluster, the kube API server would be pretty beefy and capable of handling a decent amount of load.
3. If we were to add `Secret`s as targets in the future, having to work around a cache would introduce complications. Nothing we couldn't manage, but it'd be something we'd have to think about.

Caches are complicated things and introduce a lot of possibilities for bugs. I'm thinking it might be easier to just avoid all of those bugs by explicitly skipping caching entirely and only adding it later if someone actually comes along with a concrete example of a performance issue (in which case, we'd have to look at ConfigMaps too). 

As a bonus, skipping caching will slightly reduce runtime memory usage and eliminate the possibility of us caching any Secrets we definitely don't want to cache, reducing our attack surface (slightly).

What do you think?

JoshVanL:
> 

@SgtCoDFish it is not entirely clear to me what this PR is trying to fix however the reason why we have this "multi-namespace cache" is that we want to inform on Secrets for only a single Namespace (the trust Namespace). Other resources we want to inform at the cluster level; either because they are cluster scoped resources (`Bundle`), or because we read/write to more than just the trust Namespace (`ConfigMap`s for both source and target). controller-runtime doesn't give this functionality by default, hence the custom cache implementation which is mostly just a wrapper with a bit of logic to return the right underlying cache based on GVK. Without this, we would have to give RBAC to trust-manager to be able to read secrets in all namespaces (as well as having the overhead of metadata watch on Secrets in all Namespaces rather than just one).

> Caches are complicated things and introduce a lot of possibilities for bugs. I'm thinking it might be easier to just avoid all of those bugs by explicitly skipping caching entirely and only adding it later if someone actually comes along with a concrete example of a performance issue (in which case, we'd have to look at ConfigMaps too).

We can't get away from using caches as they are a requirement for the informer. This should ideally be hidden away from us with the controller-runtime SDK, but at least at the time of writing this code, multi-scope namespacing per gvk wasn't possible.

SgtCoDFish:
> Currently, ClientDisableCacheFor disables caching for Secret resources; I reenabled caching for these resources.

> We can't get away from using caches as they are a requirement for the informer.

As far as I can tell, only one of these can be true.

JoshVanL:
@SgtCoDFish This is the point of spinning up our own Namespaced cache which is wrapped. If you disable the `ClientDisableCacheFor` which makes controller-runtime spin up cache at the cluster level then you will get the following errors. 

```
diff --git a/cmd/app/app.go b/cmd/app/app.go
index 900de4e..7967d0a 100644
--- a/cmd/app/app.go
+++ b/cmd/app/app.go
@@ -61,12 +61,12 @@ func NewCommand() *cobra.Command {
                        eventBroadcaster.StartRecordingToSink(&clientv1.EventSinkImpl{Interface: cl.CoreV1().Events("")})

                        mgr, err := ctrl.NewManager(opts.RestConfig, ctrl.Options{
-                               Scheme:                        trustapi.GlobalScheme,
-                               EventBroadcaster:              eventBroadcaster,
-                               LeaderElection:                true,
-                               LeaderElectionNamespace:       opts.Bundle.Namespace,
-                               NewCache:                      bundle.NewCacheFunc(opts.Bundle),
-                               ClientDisableCacheFor:         bundle.ClientDisableCacheFor(),
+                               Scheme:                  trustapi.GlobalScheme,
+                               EventBroadcaster:        eventBroadcaster,
+                               LeaderElection:          true,
+                               LeaderElectionNamespace: opts.Bundle.Namespace,
+                               NewCache:                bundle.NewCacheFunc(opts.Bundle),
+                               //ClientDisableCacheFor:         bundle.ClientDisableCacheFor(),
                                LeaderElectionID:              "trust-manager-leader-election",
                                LeaderElectionReleaseOnCancel: true,
                                ReadinessEndpointName:         opts.ReadyzPath,
```

```
E0109 12:17:12.222330       1 reflector.go:138] pkg/mod/k8s.io/client-go@v0.23.6/tools/cache/reflector.go:167: Failed to watch *v1.Secret: failed to list *v1.Secret: secrets is forbidden: User "system:serviceaccount:cert-manager:trust-manager" cannot list resource "secrets" in API group "" at the cluster scope
```

inteon:
> @SgtCoDFish This is the point of spinning up our own Namespaced cache which is wrapped. If you disable the `ClientDisableCacheFor` which makes controller-runtime spin up cache at the cluster level then you will get the following errors.
> 
> ```
> diff --git a/cmd/app/app.go b/cmd/app/app.go
> index 900de4e..7967d0a 100644
> --- a/cmd/app/app.go
> +++ b/cmd/app/app.go
> @@ -61,12 +61,12 @@ func NewCommand() *cobra.Command {
>                         eventBroadcaster.StartRecordingToSink(&clientv1.EventSinkImpl{Interface: cl.CoreV1().Events("")})
> 
>                         mgr, err := ctrl.NewManager(opts.RestConfig, ctrl.Options{
> -                               Scheme:                        trustapi.GlobalScheme,
> -                               EventBroadcaster:              eventBroadcaster,
> -                               LeaderElection:                true,
> -                               LeaderElectionNamespace:       opts.Bundle.Namespace,
> -                               NewCache:                      bundle.NewCacheFunc(opts.Bundle),
> -                               ClientDisableCacheFor:         bundle.ClientDisableCacheFor(),
> +                               Scheme:                  trustapi.GlobalScheme,
> +                               EventBroadcaster:        eventBroadcaster,
> +                               LeaderElection:          true,
> +                               LeaderElectionNamespace: opts.Bundle.Namespace,
> +                               NewCache:                bundle.NewCacheFunc(opts.Bundle),
> +                               //ClientDisableCacheFor:         bundle.ClientDisableCacheFor(),
>                                 LeaderElectionID:              "trust-manager-leader-election",
>                                 LeaderElectionReleaseOnCancel: true,
>                                 ReadinessEndpointName:         opts.ReadyzPath,
> ```
> 
> ```
> E0109 12:17:12.222330       1 reflector.go:138] pkg/mod/k8s.io/client-go@v0.23.6/tools/cache/reflector.go:167: Failed to watch *v1.Secret: failed to list *v1.Secret: secrets is forbidden: User "system:serviceaccount:cert-manager:trust-manager" cannot list resource "secrets" in API group "" at the cluster scope
> ```

@JoshVanL that is what this PR fixes, now controller-runtime does not try to spin up a cache at the cluster level, instead it uses the MultiScopedCache to limit the scope of the cache. It was not working before because the *corev1.Secret objects don't have a GVK value set.

JoshVanL:
@inteon could you please add a test showing what is being fixed here? Still struggling to see what we are improving. Is it the fact that ClientDisableCacheFor is now nil?

inteon:
@JoshVanL see https://github.com/cert-manager/trust-manager/pull/90/commits/a57220e69157fdfd673dd0ba0ae6ba6253013aa6 for a test, this test should fail for main

pkg/bundle/multi_scoped_cache/cache.go
```diff
@@ -14,12 +14,14 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package internal
+package multi_scoped_cache
```
JoshVanL, line 17: We want to avoid underscores in package names as much as we can. Changing the name to `cache` would be better.

https://go.dev/blog/package-names

pkg/bundle/multi_scoped_cache/cache.go
```diff
@@ -51,23 +56,23 @@ type multiScopedCache struct {
 // resources in the given namespace. namespacedInformers is the set of resource
 // types which should only be watched in the given namespace.
 // namespacedInformers expects Namespaced resource types.
-func NewMultiScopedCache(namespace string, namespacedInformers []schema.GroupKind) cache.NewCacheFunc {
+func NewMultiScopedCache(scheme *runtime.Scheme, namespace string, namespacedInformers []schema.GroupKind) cache.NewCacheFunc {
```
JoshVanL, line 59: Why do we need to pass a scheme? What's wrong with the cache Options scheme?

pkg/bundle/multi_scoped_cache/cache.go
```diff
@@ -126,26 +134,62 @@ func (b *multiScopedCache) WaitForCacheSync(ctx context.Context) bool {
...
+		return err
+	}
 	return b.cacheFromGVK(list.GetObjectKind().GroupVersionKind()).List(ctx, list, opts...)
 }
 
 // cacheFromGVK returns either the cluster or namespaced cache, based on the
 // resource type given.
 func (b *multiScopedCache) cacheFromGVK(gvk schema.GroupVersionKind) cache.Cache {
+	if gvk.Group == "" && gvk.Kind == "" {
+		panic("The Group and/or Kind must be set")
```
JoshVanL, line 163: Do we have to panic? Why can we not just return an error?

test/integration/bundle/cache_test.go
```diff
@@ -0,0 +1,152 @@
...
+*/
+
+package test
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/cert-manager/trust-manager/pkg/bundle/multi_scoped_cache"
```
JoshVanL, line 24: Organize imports.

test/integration/bundle/cache_test.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+
+	scheme := runtime.NewScheme()
+	require.NoError(t, corev1.AddToScheme(scheme))
+	newCache := multi_scoped_cache.NewMultiScopedCache(scheme, namespace, []schema.GroupKind{{Group: "", Kind: "Secret"}})
+
+	cache, err := newCache(cacheRestConfig, cache.Options{})
+	require.NoError(t, err)
+
+	go cache.Start(ctx)
```
JoshVanL, line 138: We should check the error here. We can create a stop channel which is defer closed after `Start` has returned. At the end of the test we cancel the context and wait for this stop channel to be closed.

test/integration/bundle/cache_test.go
```diff
@@ -0,0 +1,152 @@
...
+
+	if !cache.WaitForCacheSync(ctx) {
+		t.Fatal("failed to sync cache")
+	}
```
JoshVanL, line 142:
```suggestion
	assert.True(t. cache.WaitForCacheSync(ctx), "failed to sync cache")
```

test/integration/bundle/cache_test.go
```diff
@@ -0,0 +1,152 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/rest"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/envtest"
+)
+
+func TestCache(t *testing.T) {
```
JoshVanL, line 37: Why do we not run this in ginkgo?

test/integration/bundle/cache_test.go
```diff
@@ -0,0 +1,152 @@
...
+	env := &envtest.Environment{}
+
+	_, err := env.Start()
+	require.NoError(t, err)
+	defer func() {
+		err := env.Stop()
+		require.NoError(t, err)
+	}()
+
+	ctx, cancel := context.WithCancel(context.Background())
```
JoshVanL, line 47: We should have a timeout so that the test doesn't block forever

JoshVanL (changes requested):

pkg/bundle/multi_scoped_cache/cache.go
inteon: @JoshVanL that pointer can be nil, how would you get a default value in case that pointer is nil?

JoshVanL: If nil is not acceptable, we can default to runtime.NewScheme()

pkg/bundle/multi_scoped_cache/cache.go
inteon: I updated the code

test/integration/bundle/cache_test.go
inteon: I changed it.

inteon:
@JoshVanL can you approve the PR too?

PR #92, opened 2023-01-11
Fix non-cached client

### LINKED ISSUE #91: Cache correctness investigations
In #90 @inteon implemented some changes to help with cache correctness, since trust-manager has to work around some limitations currently present in controller-runtime (see https://github.com/kubernetes-sigs/controller-runtime/pull/1962 which should help).

We believe that PR was an improvement, but there might be other issues that need addressing which Tim is looking into. 

> It seems like my PR just makes Secrets the same as the other resources, but there is a fix necessary to make everything work as intended (minimal caching etc.).
> [and to avoid] duplicate caching (see this comment for more info: https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/builder/options.go#L103-L132)
> My current hypothesis is that we cache every resource 2 times (one time metadata only and one time fully), I still have to check by running an actual example.

This is a tracking issue for that work, which is likely a sensible prerequisite for a 0.4.0 release.

inteon:
Here, we incorrectly assume that the client is a non-cached client:
https://github.com/cert-manager/trust-manager/blob/f2da89271eb887093f434a4e79b80d1fee02a212/pkg/bundle/bundle.go#L57-L60
This results in caching all these resources fully. Additionally, we are already caching the metadata-only representation of these resources, effectively caching each resource twice (more info: https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/builder/options.go#L103-L132).

In this PR, I fix this issue and also rename the `client` field to `directClient` to make clear that this client is not cached.
Also, I simplified the cache.go code a bit by using an external library.

fixes #91

### DIFF
--- a/pkg/bundle/controller.go
+++ b/pkg/bundle/controller.go
@@ -47,12 +47,20 @@ import (
 // when any related resource event in the Bundle source and target.
 // The controller will only cache metadata for ConfigMaps and Secrets.
 func AddBundleController(ctx context.Context, mgr manager.Manager, opts Options) error {
+	directClient, err := client.New(mgr.GetConfig(), client.Options{
+		Scheme: mgr.GetScheme(),
+		Mapper: mgr.GetRESTMapper(),
+	})
+	if err != nil {
+		return fmt.Errorf("failed to create client: %w", err)
+	}
+
 	b := &bundle{
-		client:   mgr.GetClient(),
-		lister:   mgr.GetCache(),
-		recorder: mgr.GetEventRecorderFor("bundles"),
-		clock:    clock.RealClock{},
-		Options:  opts,
+		directClient: directClient,
+		lister:       mgr.GetCache(),
+		recorder:     mgr.GetEventRecorderFor("bundles"),
+		clock:        clock.RealClock{},
+		Options:      opts,
 	}
 
 	if b.Options.DefaultPackageLocation != "" {
--- a/pkg/bundle/bundle.go
+++ b/pkg/bundle/bundle.go
@@ -54,10 +54,10 @@ type Options struct {
 // bundle is a controller-runtime controller. Implements the actual controller
 // logic by reconciling over Bundles.
 type bundle struct {
-	// client is a Kubernetes client that makes calls to the API for every request.
+	// directClient is a Kubernetes client that makes calls to the API for every request.
 	// Should be used for updating, deleting, and when requesting data from
 	// resources whose informer only caches metadata.
-	client client.Client
+	directClient client.Client
 
 	// lister makes requests to the informer cache. Beware that resources whose
 	// informer only caches metadata, will not return underlying data of that
@@ -126,7 +126,7 @@ func (b *bundle) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, 
 				},
 			}
 
-			err := b.client.Get(ctx, client.ObjectKeyFromObject(configMap), configMap)
+			err := b.directClient.Get(ctx, client.ObjectKeyFromObject(configMap), configMap)
 
 			// Ignore ConfigMaps that have not been created yet, as they will be
 			// created later on in the sync.
@@ -142,7 +142,7 @@ func (b *bundle) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, 
 
 			delete(configMap.Data, bundle.Status.Target.ConfigMap.Key)
 
-			if err := b.client.Update(ctx, configMap); err != nil {
+			if err := b.directClient.Update(ctx, configMap); err != nil {
 				log.Error(err, "failed to delete old ConfigMap target key")
 				b.recorder.Eventf(&bundle, corev1.EventTypeWarning, "TargetUpdateError", "Failed to remove old key from ConfigMap target: %s", err)
 				return ctrl.Result{}, fmt.Errorf("failed to delete old ConfigMap target key: %w", err)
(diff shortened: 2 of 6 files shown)


pkg/bundle/controller.go
```diff
@@ -47,12 +47,20 @@ import (
 // when any related resource event in the Bundle source and target.
 // The controller will only cache metadata for ConfigMaps and Secrets.
 func AddBundleController(ctx context.Context, mgr manager.Manager, opts Options) error {
+	directClient, err := client.New(mgr.GetConfig(), client.Options{
+		Scheme: mgr.GetScheme(),
+		Mapper: mgr.GetRESTMapper(),
+	})
+	if err != nil {
+		return fmt.Errorf("failed to create client: %w", err)
+	}
```
inteon, line 56: Here I create the direct client instead of using `mgr.Client()` which is a cached client.

irbekrm (approved):
Thanks @inteon this makes sense to me.

I can reproduce the issue (with increased log level), when trust manager is run from current main branch, and some resources including a ConfigMap source is deployed, I can see that a watch is started:
```
...
I0111 18:49:09.429622       1 controller.go:220] trust/manager/controller/bundle "msg"="Starting workers" "reconciler group"="trust.cert-manager.io" "reconciler kind"="Bundle" "worker count"=1
I0111 18:49:09.429687       1 bundle.go:86] trust/bundle "msg"="syncing bundle" "bundle"="my-org.com" 
I0111 18:49:09.430027       1 reflector.go:219] Starting reflector *v1.ConfigMap (9h13m35.601236097s) from pkg/mod/k8s.io/client-go@v0.23.6/tools/cache/reflector.go:167
I0111 18:49:09.430055       1 reflector.go:255] Listing and watching *v1.ConfigMap from pkg/mod/k8s.io/client-go@v0.23.6/tools/cache/reflector.go:167
I0111 18:49:09.530536       1 shared_informer.go:270] caches populated
I0111 18:49:09.530779       1 bundle.go:86] trust/bundle "msg"="syncing bundle" "bundle"="my-org.com" 
...
```
whereas from this branch only metadata, bundle and namespace watches are started which matches what I would expect looking at which resources are specified to be cached in full and for which metadata only should be cached:
```
...
I0111 18:54:00.811795       1 reflector.go:219] Starting reflector *v1.PartialObjectMetadata (9h38m36.286332225s) from pkg/mod/k8s.io/client-go@v0.23.6/tools/cache/reflector.go:167
I0111 18:54:00.811804       1 reflector.go:255] Listing and watching *v1.PartialObjectMetadata from pkg/mod/k8s.io/client-go@v0.23.6/tools/cache/reflector.go:167
I0111 18:54:00.811807       1 reflector.go:219] Starting reflector *v1.Namespace (10h14m43.088302663s) from pkg/mod/k8s.io/client-go@v0.23.6/tools/cache/reflector.go:167
I0111 18:54:00.811814       1 reflector.go:255] Listing and watching *v1.Namespace from pkg/mod/k8s.io/client-go@v0.23.6/tools/cache/reflector.go:167
I0111 18:54:00.811868       1 reflector.go:219] Starting reflector *v1alpha1.Bundle (9h11m15.662603274s) from pkg/mod/k8s.io/client-go@v0.23.6/tools/cache/reflector.go:167
I0111 18:54:00.811874       1 reflector.go:255] Listing and watching *v1alpha1.Bundle from pkg/mod/k8s.io/client-go@v0.23.6/tools/cache/reflector.go:167
I0111 18:54:00.812150       1 reflector.go:219] Starting reflector *v1.PartialObjectMetadata (9h9m53.876857508s) from pkg/mod/k8s.io/client-go@v0.23.6/tools/cache/reflector.go:167
I0111 18:54:00.812161       1 reflector.go:255] Listing and watching *v1.PartialObjectMetadata from pkg/mod/k8s.io/client-go@v0.23.6/tools/cache/reflector.go:167
I0111 18:54:00.814308       1 event.go:265] Server rejected event '&v1.Event{TypeMeta:v1.TypeMeta{Kind:"", APIVersion:""}, ObjectMeta:
...
```
As @wallrj mentioned there is an alternative way how to specifiy that certain resources should not be cached in new client func https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/cluster/cluster.go#L257

However, I prefer your solution- I agree that otherwise it'd be easy to forget to update the list of objects that should not be cached otherwise if we start interacting with a new resource type.

/lgtm

vincepri:
👋 Controller Runtime maintainer here!

Apologies for the confusion here with our UX, need to think more about how to improve it long term. A few things that I can clarify:

- Controllers using Manager's Clients are by default using a cached clients for read operations.
- Any `Get()` call with a typed (e.g. appsv1.Pod{}), or unstructured, or partial metadata starts an internal cache that watches those objects. If there are 3 calls with the same GroupVersionKind, but with a typed/unstructured/metadata object, 3 total caches are started, causing duplication.
- To avoid caching specific resources, I'd suggest using either a live client (APIReader), like this PR proposed, or even better [you can configure the manager](https://github.com/kubernetes-sigs/controller-runtime/blob/613648eda7831bea8928e9fb1e3ea06df5f3851a/pkg/manager/manager.go#L274-L276) to always avoid caching certain types; the check is done per GroupVersionKind, and applies to every object.

The above needs to be improved, and frankly all of the options we have should eventually go in a builder of sorts for both controllers and the manager itself, instead of being in Options.

irbekrm:
Thanks for that great description @vincepri !

> or even better [you can configure the manager](https://github.com/kubernetes-sigs/controller-runtime/blob/613648eda7831bea8928e9fb1e3ea06df5f3851a/pkg/manager/manager.go#L274-L276) to always avoid caching certain types;

What would be some reasons why this would be better than standalone live client? Maybe because it uses the same rate limiting for all requests? Otherwise it feels like it would be unlikely that a person down the line adding a GET request for a new type would actually remember to add it to uncached types unless they've memorized that this needs to be done.

vincepri:
What's better probably depends on what you're looking the default behavior to be like, if one prefers the default to be always uncached, an `APIReader` is probably a better choice. On the flip side, if you only want to avoid caching certain objects for reasons like size, freshness, etc, the cached client is preferred. Usually caches come in handy when lots of Get/List requests are performed against the same objects, it avoid overloading the API Server with many roundtrips.

irbekrm:
Thanks that makes sense! I was worried that perhaps there's something that we're missing out by using `APIReader` to GET objects that we know we don't need to cache (as opposed to using client obtained from the manager). We do use cache for the objects that need to be retrieved in full frequently 👍🏼

PR #97, opened 2023-01-16
Helm tweaks + better makefile "ensure"

SgtCoDFish:
See individual commits for details


deploy/charts/trust-manager/templates/NOTES.txt
```diff
@@ -0,0 +1,14 @@
...
+
+{{ .Values.defaultPackageImage.repository }}:{{ .Values.defaultPackageImage.tag }}
+
+It's imperative that you keep the default CA package image up to date.
+
+To find out more about securely running trust-manager and to get started
+with creating your first bundle, check out the documentation on the
+cert-manager website:
+
+https://cert-manager.io/docs/projects/trust-manager/
```
irbekrm, line 14: This whole note is a great addition

Makefile
```diff
@@ -36,6 +36,8 @@ CONTAINER_REGISTRY ?= quay.io/jetstack
 
 GOPROXY ?= https://proxy.golang.org,direct
 
+KUBECONFIG ?= $(BINDIR)/kubeconfig.yaml
```
irbekrm, line 39: What is the purpose of this variable? As a user I would expect that if I pass a custom path here, the current cluster from that kubeconfig will be used. However, that is not the case, this will always result in a kind cluster named 'trust' will be created, my local kubeconfig will be modified as well as one will be created in $BINDIR.
Should we just keep the kubeconfig hardcoded to not make folks think they can use this var to parameterize it?

Makefile
```diff
@@ -142,18 +144,18 @@ ensure-kind: depend ensure-ci-docker-network  ## create a trust-manager kind clu
...
 		$(BINDIR)/helm repo add jetstack https://charts.jetstack.io --force-update; \
-		$(BINDIR)/helm upgrade -i --create-namespace -n cert-manager cert-manager jetstack/cert-manager --set installCRDs=true --wait; \
+		$(BINDIR)/helm upgrade --kubeconfig $(KUBECONFIG) -i --create-namespace -n cert-manager cert-manager jetstack/cert-manager --set installCRDs=true --wait; \
 	fi
 
 .PHONY: ensure-trust-manager
 ensure-trust-manager: depend ensure-kind kind-load ensure-cert-manager
 	$(BINDIR)/helm uninstall -n cert-manager trust-manager || :
-	$(BINDIR)/helm upgrade -i -n cert-manager trust-manager deploy/charts/trust-manager/. --set image.tag=latest --set app.logLevel=2 --wait
+	$(BINDIR)/helm upgrade --kubeconfig $(KUBECONFIG) -i -n cert-manager trust-manager deploy/charts/trust-manager/. --set image.tag=latest --set defaultTrustPackage.tag=latest$(DEBIAN_TRUST_PACKAGE_SUFFIX) --set app.logLevel=2 --wait
```
irbekrm, line 158: (Not related to this PR) but it'd be nice if this make target had a comment so it appears in `make help` output. I assume that this is a top level target that users would run often

irbekrm (commented):
Thanks @SgtCoDFish I've added a couple nits and comments.

Generally from a perspective of someone who is going to perhaps deploy this project together with some other projects I prefer an approach where make targets for creating a (kind) cluster and deploying the actual software don't depend on each other, so if I already have a cluster with some stuff and I want to deploy trust-manager to that particular cluster, I have a straightforward make target where I can pass a kubeconfig or kind cluster name and it will do it.
However, that approach is also not perfect as if you want to load images it'll only work for kind clusters anyway. So happy to approve minus a couple nits and comments 👍🏼

Makefile
SgtCoDFish: > What is the purpose of this variable? 

Originally my idea was to allow users to put in their own kubeconfig, but you make a valid point - that won't work without more work.

>  Should we just keep the kubeconfig hardcoded to not make folks think they can use this var to parameterize it?

I agree with this yeah, makes sense. Great review comment, thanks!

Makefile
SgtCoDFish: I can add that in this PR, makes sense :+1:

Makefile
```diff
@@ -131,7 +133,7 @@ smoke: demo  ## ensure cluster, deploy trust-manager and run smoke tests
 	${BINDIR}/ginkgo -procs 1 test/smoke/ -- --kubeconfig-path ${BINDIR}/kubeconfig.yaml
 
 $(BINDIR)/kubeconfig.yaml: depend ensure-kind _FORCE | $(BINDIR)
-	$(BINDIR)/kind get kubeconfig --name trust > $@
+	$(BINDIR)/kind get kubeconfig --name trust > $@ && chmod 600 $@
```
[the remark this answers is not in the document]
SgtCoDFish: As I understand it, you're spot on for the case when you initially create a cluster for the first time.

With this, I'm trying to help in the case where a `trust` cluster already exists, and the person is coming back to the trust repo after doing some other work (e.g. on cert-manager).

Put another way: with this I'm trying to ensure that if the default kubeconfig was modified _after_ `ensure-kind` was run, we won't rely on the default kubeconfig being unchanging.

We _could_ use `kubectl --context kind-trust` - but we can't control that file and so it could be messed with so it feels a little scarier to rely on it.

Plus having this file lets us use it elsewhere; I used it in the [draft PR](https://github.com/cert-manager/website/pull/1160/files#diff-d81dfaf17a64dfde33b0e443f6b76b70c0c8d6543d9089ea60ffbcb7fa5c8511R171) for the new trust-manager docs, to ensure that users are only talking to the cluster created by the Makefile.

SgtCoDFish:
> Generally from a perspective of someone who is going to perhaps deploy this project together with some other projects I prefer an approach where make targets for creating a (kind) cluster and deploying the actual software don't depend on each other, so if I already have a cluster with some stuff and I want to deploy trust-manager to that particular cluster, I have a straightforward make target where I can pass a kubeconfig or kind cluster name and it will do it.

I think that's a totally valid thing but I'm not sure from the eyes of an end-user it's how I'd see it. As an end-user I ideally wouldn't be cloning the repo to use the makefile - I'd be installing from Helm, because that's the supported way.

Most of this is very dev-focused, I guess?

Makefile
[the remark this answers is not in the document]
irbekrm: That makes sense, thanks for that detailed explanation!

irbekrm:
> I think that's a totally valid thing but I'm not sure from the eyes of an end-user it's how I'd see it. As an end-user I ideally wouldn't be cloning the repo to use the makefile - I'd be installing from Helm, because that's the supported way.
Most of this is very dev-focused, I guess?

Yeah I also meant it from dev perspective - I could have scripts that I use that create a certain cluster configuration with a certain name where I want to install trust or I might want to install two cert-manager related projects from latest master to the same cluster- in both cases it'd be easier if they don't require a specific cluster name.
But this is just a non-actionable note- I don't want to block this PR which is overall valuable with this.

SgtCoDFish:
I've pushed the requested changes and a couple more:

- Fixed smoke test argument passing so `kubeconfig-path` is actually used. It never previously worked!
- Maybe more controversial: Only writes kubeconfig into bin, so that `$HOME/.kube/config` isn't touched. I'm kinda unsure about that, since it'll make it a little harder to play with a local demo cluster, but I like that it exposes bugs like the smoke test thing above. WDYT?

irbekrm:
Thanks for making the changes @SgtCoDFish !

> Maybe more controversial: Only writes kubeconfig into bin, so that $HOME/.kube/config isn't touched. I'm kinda unsure about that, since it'll make it a little harder to play with a local demo cluster, but I like that it exposes bugs like the smoke test thing above. WDYT?

I think it makes sense to do whatever is needed to make the tests less flakey.
I do think that it makes it harder to play with a local demo cluster- to me this seems to be another argument to split the local deploy flow into separate make targets that aren't dependent on each other, so that a user can run a combination of
`make ensure-kind`, `make load-images KIND_CLUSTER_NAME=X, KUBECONFIG=./X`, `make deploy-trust-manager KUBECONFIG=./X`. Or they can run a single target that does all of that with less configurability that the e2e tests also use if they are okay with it being a bit harder to make the cluster usable for other scenarios. However, that'd probably not be in this PR.

In this PR, perhaps we could print something like `kind cluster X was created, if you want to access it run kind export kubeconfig --name trust` in the target that creates the cluster?
Otherwise happy to approve this PR.

SgtCoDFish:
> In this PR, perhaps we could print something like kind cluster X was created, if you want to access it run kind export kubeconfig --name trust in the target that creates the cluster?

I like this, makes a lot of sense. Will add :+1:

irbekrm (approved):
Thanks for working on this and for making all the changes @SgtCoDFish !

This looks good to go in to me 👍🏼 I think it's definitely an improvement

/lgtm

PR #98, opened 2023-01-18
Cert formats proposal

SgtCoDFish:
Initial point of discussion around supporting multiple formats of certificate on both sources / targets!


erikgb:
@SgtCoDFish Is this PR still relevant? We now have PEM, JKS and PKCS12 as supported target formats.

SgtCoDFish:
I think it's still relevant to an extent, because this also proposes adding various input formats... but realistically there's no rush and I won't be working on this any time soon - so I'll close this PR at least.

PR #102, opened 2023-01-20
Make CRD installation optional

maxsargentdev:
Signed-off-by: Max Sargent <maxsargentdev@gmail.com>

### DIFF
--- a/hack/update-codegen.sh
+++ b/hack/update-codegen.sh
@@ -46,7 +46,7 @@ ${BIN_DIR}/deepcopy-gen --input-dirs "$(codegen::join , "${FQ_APIS[@]}")" -O zz_
 echo "Generating CRDs in ./deploy/crds"
 ${BIN_DIR}/controller-gen crd schemapatch:manifests=./deploy/crds output:dir=./deploy/crds paths=./pkg/apis/...
 
-echo "Updating CRDs with helm templating, writing to /deploy/charts/trust-manager/templates"
+echo "Updating CRDs with helm templating, writing to ./deploy/charts/trust-manager/templates"
 for i in $(ls ./deploy/crds); do
 
   cat << EOF > ./deploy/charts/trust-manager/templates/$i
--- a/hack/update-codegen.sh
+++ b/hack/update-codegen.sh
@@ -43,5 +43,16 @@ done
 echo "Generating deepcopy funcs"
 ${BIN_DIR}/deepcopy-gen --input-dirs "$(codegen::join , "${FQ_APIS[@]}")" -O zz_generated.deepcopy --bounding-dirs "${APIS_PKG}" -h $BOILERPLATE
 
-echo "Generating CRDs in ./deploy/charts/trust-manager/templates"
-${BIN_DIR}/controller-gen crd schemapatch:manifests=./deploy/charts/trust-manager/templates output:dir=./deploy/charts/trust-manager/templates paths=./pkg/apis/...
+echo "Generating CRDs in ./deploy/crds"
+${BIN_DIR}/controller-gen crd schemapatch:manifests=./deploy/crds output:dir=./deploy/crds paths=./pkg/apis/...
+
+echo "Updating CRDs with helm templating, writing to /deploy/charts/trust-manager/templates"
+for i in $(ls ./deploy/crds); do
+
+  cat << EOF > ./deploy/charts/trust-manager/templates/$i
+{{ if .Values.crds.enabled }}
+$(cat ./deploy/crds/$i)
+{{ end }}
+EOF
+
+done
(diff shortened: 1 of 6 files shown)


inteon:
@maxsargentdev could you explain what your use case is?

maxsargentdev:
Hi @inteon,

Sure, we have a local development environment using a few tools Tilt, K3s etc and the structure of it means we separate out CRDs and install them all first.

Mainly because in the past we have been stung by Helm and how it manages CRDs.

For now I have forked this repo and created a chart manually which omits the bundle CRD but would prefer to use your guys chart so we can more easily receive updates in future.

Thanks,
Max Sargent

SgtCoDFish (changes requested):
Thanks for your contribution! I think this makes sense - you're not the first person to mention issues with Helm and CRDs.

Unfortunately I'm not sure if this is a complete solution. The CRD YAML file you've modified is automatically generated and would be generated without the Helm templating logic you've added.

By way of an example, if you apply the following patch to add a new dummy field to the CRD:

```patch
diff --git a/pkg/apis/trust/v1alpha1/types_bundle.go b/pkg/apis/trust/v1alpha1/types_bundle.go
index 9ed239a..87dbd32 100644
--- a/pkg/apis/trust/v1alpha1/types_bundle.go
+++ b/pkg/apis/trust/v1alpha1/types_bundle.go
@@ -87,6 +87,9 @@ type BundleSource struct {
 	// defaultCAPackageVersion field of the Bundle's status field.
 	// +optional
 	UseDefaultCAs *bool `json:"useDefaultCAs,omitempty"`
+
+	// Dummy does nothing
+	Dummy *string `json:"dummy,omitempty"`
 }
 
 // BundleTarget is the target resource that the Bundle will sync all source
```

and then run `make generate`, your changes will be overwritten.

One solution might be to update the codegen script (https://github.com/cert-manager/trust-manager/blob/698875cce5f694c6153c2afe3256a7e2d1f1ac96/hack/update-codegen.sh#L46-L47). Rather than writing directly to `deploy/charts/trust-manager/templates` we could write to a new folder - `deploy/crds` and then copy the result into `deploy/charts/trust-manager/templates` with the new Helm templating logic added.

What do you think? Thanks again for getting involved!

maxsargentdev:
This makes perfect sense and I will push a new commit tomorrow with the updates you have described.

Thanks for taking the time to look at my pull request.

maxsargentdev:
Hey @SgtCoDFish 

I have updated the code gen script as you have described, its not a perfect solution so if you want it changed let me know.

I have also not committed any of the changes that are caused when running the new script, let me know if you want them in this PR.

Thanks,
Max Sargent

hack/update-codegen.sh
```diff
@@ -43,5 +43,16 @@ done
 echo "Generating deepcopy funcs"
 ${BIN_DIR}/deepcopy-gen --input-dirs "$(codegen::join , "${FQ_APIS[@]}")" -O zz_generated.deepcopy --bounding-dirs "${APIS_PKG}" -h $BOILERPLATE
 
-echo "Generating CRDs in ./deploy/charts/trust-manager/templates"
-${BIN_DIR}/controller-gen crd schemapatch:manifests=./deploy/charts/trust-manager/templates output:dir=./deploy/charts/trust-manager/templates paths=./pkg/apis/...
+echo "Generating CRDs in ./deploy/crds"
+${BIN_DIR}/controller-gen crd schemapatch:manifests=./deploy/crds output:dir=./deploy/crds paths=./pkg/apis/...
+
```
SgtCoDFish, line 48: suggestion: I would expect that we'd have a new file in `./deploy/crds` in this PR if we're now generating CRDs there as well?

hack/update-codegen.sh
```diff
@@ -43,5 +43,16 @@ done
...
+echo "Updating CRDs with helm templating, writing to /deploy/charts/trust-manager/templates"
```
SgtCoDFish, line 49: nitpick: The path looks like it's wrong maybe?

```suggestion
echo "Updating CRDs with helm templating, writing to ./deploy/charts/trust-manager/templates"
```

SgtCoDFish (changes requested):
Couple of minor things but I like this and appreciate your work on it! :grin:

maxsargentdev:
Have updated with the changes requested, thanks again for spending time looking at this.

maxsargentdev:
I will run the make command for the README when I get in from the office & trigger a retest.

SgtCoDFish:
Thank you! Sorry I forgot the ok-to-test earlier :sweat_smile:

SgtCoDFish (approved):
/lgtm
/approve

Thanks so much @maxsargentdev  - really appreciate your contribution!

SgtCoDFish:
Ahhh, that's a shame - looks like the tests actually had the old path in there and expect the CRDs to be plain old YAML. I think https://github.com/cert-manager/trust-manager/blob/main/test/integration/bundle/integration.go#L40 will need to be updated to the new `./deploy/crds` directory.

I'm afraid it's 19:00 in my timezone so I've gotta sign off for the day - but I'll take another look at this tomorrow!

maxsargentdev:
@SgtCoDFish 

No worries, I have updated it was only a small change.

Thanks again for this, its my first small contribution on GitHub!

SgtCoDFish (approved):
/lgtm
/approve

Thanks again for this! I was ill yesterday so forgot about this, but glad to merge it now :tada: 

> Thanks again for this, its my first small contribution on GitHub!

Then congratulations are in order too! :tada: Glad to be part of it!

PR #104, opened 2023-01-25
Rename '_test.go' suffix to test file

inteon:
The `cnf` variable is defined in a `_test` file and referred to by a non `_test` file, this makes my IDE complain.
So, in this PR I rename the file and added a `_test` suffix.


SgtCoDFish (approved):
/lgtm
/approve

seems reasonable, thanks!

PR #107, opened 2023-01-26
Use a source and a target cache instead of a MultiScopedCache

inteon:
Instead of the `MultiScopedCache` mechanism, this PR proposed to manually create two caches: one for the source resources and one for the target resources.

- The source resources are fully cached (not metadata-only) and limited to the trust Namespace.
- The target resources are metadata-only cached and cached for all namespaces in the cluster.

This change removes the `MultiScopedCache` and also allows us to cache more resources, making trust-manager more efficient.


pkg/bundle/bundle.go
```diff
@@ -59,10 +59,9 @@ type bundle struct {
 	// resources whose informer only caches metadata.
 	directClient client.Client
 
-	// lister makes requests to the informer cache. Beware that resources whose
-	// informer only caches metadata, will not return underlying data of that
-	// resource. Use client instead.
-	lister client.Reader
+	// sourceLister makes requests to the informer cache. All cached source resources
+	// are expected to be full objects in a single namespace (the TrustNamespace).
+	sourceLister client.Reader
```
SgtCoDFish, line 64: suggestion (non-blocking): I like the lister -> sourceLister change a lot, it makes a lot of sense to treat sources and targets differently.

Does this mean that `directClient` would now only be used for interacting with targets? If so, might it be worth renaming it to `targetClient` or similar to make the name better match what the client is used for? Totally not needed though!

pkg/bundle/controller.go
```diff
@@ -76,29 +133,26 @@ func AddBundleController(ctx context.Context, mgr manager.Manager, opts Options)
...
+
+		////// Targets //////
 
 		// Reconcile over owned ConfigMaps in all Namespaces. Only cache metadata.
 		// These ConfigMaps will be Bundle Targets
-		Owns(new(corev1.ConfigMap), builder.OnlyMetadata).
+		Watches(source.NewKindWithCache(new(corev1.ConfigMap), targetCache), &handler.EnqueueRequestForOwner{
+			OwnerType:    new(trustapi.Bundle),
+			IsController: true,
+		}, builder.OnlyMetadata).
```
SgtCoDFish, line 145: question: Would it be possible for you to expand on why we want to do caching for targets here, even metadata-only? I'm struggling to think of situations in which it helps!

SgtCoDFish (commented):
Really enjoyed looking over this and I learned a bunch. Absolutely love getting rid of the custom cache impl.

Got a few comments and questions, just to make sure I understood. Mostly it's not important stuff though - this looks good to me generally!

pkg/bundle/controller.go
inteon: It is not necessary, but this is the easiest way to watch for resource changes.

SgtCoDFish: > It is not necessary, but this is the easiest way to watch for resource changes/ obtain an Informer.

From looking more at the code for controller-runtime, I can see that we'd have to go out of our way to _avoid_ using a cache here, so it makes sense that it'd be easier to just use one. That's good by me!

What I don't understand now is if we were to use the following:

```go
Watches(&source.Kind{Type: &corev1.ConfigMap{}}, &handler.EnqueueRequestForOwner{
			OwnerType:    new(trustapi.Bundle),
			IsController: true,
		}, builder.OnlyMetadata).
```

(Or `Owns(new(corev1.ConfigMap), builder.OnlyMetadata).` as before? That seems equivalent?)

Then from the docs, at some point `InjectCache` will be called which will add a cache [initialized by the ControllerManager](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/source#Kind.InjectCache).

Why do we need to define our own target cache here rather than the one initialized by the ControllerManager? What're we gaining from that? Would it be worth adding that reasoning when we define the targetCache?

SgtCoDFish (approved):
/lgtm
/approve
/hold

Holding just because I still had a question, but I'd merge this as is! Thanks for removing hundreds of lines of code :tada:

Issue #10, opened 2022-01-17
Feature: support secret target


cortopy:
Currently only `configMap` is supported as `target`. It feels to me like that's the obvious choice for distributing a ca file. However, a lot of off-the-shelf applications out there only support mounting ca files via secrets

Would it be possible to add it for compatibility reasons?

JoshVanL:
Hi @cortopy, there is no technical reason why adding `Secret` as a target could not be done, however is poses security implications (trust would be able read + write all Secrets in all namespaces). If added, we would likely want this feature disabled by default, and users would have to explicitly toggle that they 1. want this target available, and 2. grant and understand they are giving trust the permissions to do this.

cortopy:
thanks @JoshVanL for such a quick answer. I hadn't thought of that but the path you propose sounds excellent

james-callahan:
As a mitigation to that risk, perhaps a suggestion for RBAC on secret to explicitly list resource names?

erikgb:
I support adding secrets as an opt-in target for bundles. There are just so many tools that only support obtaining CA certificates from secrets. And one could argue that even if a CA certificate bundle is not strictly secret, defining who you trust is definitely important for overall security.

> As a mitigation to that risk, perhaps a suggestion for RBAC on secret to explicitly list resource names?

@james-callahan I don't think this is possible in general. As a user of trust-manager, I would expect to be allowed to express the secret name in the bundle resource. AFAIK the controller-runtime mechanics (and Kubernetes API) do not really support watching resources cluster-wide by name(s).

erikgb:
I believe is issue was fixed by https://github.com/cert-manager/trust-manager/pull/193

/close

PR #108, opened 2023-01-30
Supporting a secret target

rudexi:
This was requested in https://github.com/cert-manager/trust-manager/issues/10

Motivation:
Many helm charts/operators support a certificate authority mount only in the form of a Secret, and do not support ConfigMap. This is due to the fact that certificate authority is often shipped with TLS-type secrets (along with their certificate and key). cert-manager also require a certificate authority in the form of a Secret for its Vault issuer.

trust-manager should support the target being a Secret, along with a ConfigMap, to guarantee maximum usability.

For security concerns, we will allow trust-manager to only view/update a given list of managed secrets in the helm chart.

### DIFF
--- a/deploy/charts/trust-manager/values.yaml
+++ b/deploy/charts/trust-manager/values.yaml
@@ -67,6 +67,8 @@ app:
     # -- If false, disables the default seccomp profile, which might be required to run on certain platforms
     seccompProfileEnabled: true
 
+authorizedSecrets: ['ca-bundle']
+
 resources: {}
   # -- Kubernetes pod resource limits for trust.
   # limits:
--- a/deploy/charts/trust-manager/values.yaml
+++ b/deploy/charts/trust-manager/values.yaml
@@ -67,6 +67,8 @@ app:
     # -- If false, disables the default seccomp profile, which might be required to run on certain platforms
     seccompProfileEnabled: true
 
+# -- List of secrets trust-manager is allowed to read/write. Used when the secret target is set when creating
+# a bundle.
 authorizedSecrets: ['ca-bundle']
 
 resources: {}
--- a/pkg/apis/trust/v1alpha1/types_bundle.go
+++ b/pkg/apis/trust/v1alpha1/types_bundle.go
@@ -96,6 +96,10 @@ type BundleTarget struct {
 	// data will be synced to.
 	ConfigMap *KeySelector `json:"configMap,omitempty"`
 
+    // Secret is the target Secret in Namespaces that all Bundle source
+    // data will be synced to.
+    Secret *KeySelector `json:"secret,omitempty"`
+
 	// NamespaceSelector will, if set, only sync the target resource in
 	// Namespaces which match the selector.
 	// +optional
--- a/pkg/bundle/bundle.go
+++ b/pkg/bundle/bundle.go
@@ -202,7 +202,26 @@ func (b *bundle) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, 
 			return ctrl.Result{Requeue: true}, b.targetDirectClient.Status().Update(ctx, &bundle)
 		}
 
-		if synced {
+        secretSynced := false
+        if bundle.Spec.Target.Secret != nil {
+            var syncErr error
+            secretSynced, syncErr = b.syncSecretTarget(ctx, log, &bundle, namespaceSelector, &namespace, []byte(resolvedBundle.data))
+            if syncErr != nil {
+                log.Error(syncErr, "failed sync bundle to target namespace")
+                b.recorder.Eventf(&bundle, corev1.EventTypeWarning, "SyncTargetFailed", "Failed to sync secret target in Namespace %q: %s", namespace.Name, syncErr)
+
+                b.setBundleCondition(&bundle, trustapi.BundleCondition{
+                    Type:    trustapi.BundleConditionSynced,
+                    Status:  corev1.ConditionFalse,
+                    Reason:  "SyncTargetFailed",
+                    Message: fmt.Sprintf("Failed to sync bundle target secret to namespace %q: %s", namespace.Name, syncErr),
+                })
+
+                return ctrl.Result{Requeue: true}, b.targetDirectClient.Status().Update(ctx, &bundle)
+            }
+        }
+
+		if synced || secretSynced {
 			// We need to update if any target is synced.
 			needsUpdate = true
 		}
--- a/pkg/bundle/bundle.go
+++ b/pkg/bundle/bundle.go
@@ -187,16 +187,16 @@ func (b *bundle) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, 
 			continue
 		}
 
-		synced, err := b.syncTarget(ctx, log, &bundle, namespaceSelector, &namespace, resolvedBundle.data)
+		cmSynced, err := b.syncTarget(ctx, log, &bundle, namespaceSelector, &namespace, resolvedBundle.data)
 		if err != nil {
 			log.Error(err, "failed sync bundle to target namespace")
-			b.recorder.Eventf(&bundle, corev1.EventTypeWarning, "SyncTargetFailed", "Failed to sync target in Namespace %q: %s", namespace.Name, err)
+			b.recorder.Eventf(&bundle, corev1.EventTypeWarning, "SyncTargetFailed", "Failed to sync configMap target in Namespace %q: %s", namespace.Name, err)
 
 			b.setBundleCondition(&bundle, trustapi.BundleCondition{
 				Type:    trustapi.BundleConditionSynced,
 				Status:  corev1.ConditionFalse,
 				Reason:  "SyncTargetFailed",
-				Message: fmt.Sprintf("Failed to sync bundle to namespace %q: %s", namespace.Name, err),
+				Message: fmt.Sprintf("Failed to sync bundle target configMap to namespace %q: %s", namespace.Name, err),
 			})
 
 			return ctrl.Result{Requeue: true}, b.targetDirectClient.Status().Update(ctx, &bundle)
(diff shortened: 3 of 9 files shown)


rudexi:
/assign @sgtcodfish
I am already running this PR in my environment, and so far so good.

SgtCoDFish:
Hi @rudexi thanks for the contribution! I'm afraid I don't have the bandwidth to review this right now (and unfortunately it looks like a merge conflict crept in today as @inteon refactored how we do caching).

I'll try to review this this week but I'm super busy and can't promise anything I'm afraid!

pkg/bundle/bundle.go
```diff
@@ -202,7 +202,26 @@ func (b *bundle) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result,
...
+                    Status:  corev1.ConditionFalse,
+                    Reason:  "SyncTargetFailed",
+                    Message: fmt.Sprintf("Failed to sync bundle target secret to namespace %q: %s", namespace.Name, syncErr),
+                })
+
+                return ctrl.Result{Requeue: true}, b.targetDirectClient.Status().Update(ctx, &bundle)
+            }
+        }
+
+		if synced || secretSynced {
```
SgtCoDFish, line 224: suggestion: Would it be a good idea here to rename `synced` to `configMapSynced` so it's clearer what that variable is doing?

The log messages and error messages for the config map sync above could maybe also use updating to directly mention that they're for configmaps

pkg/bundle/bundle.go
```diff
@@ -202,7 +202,26 @@ func (b *bundle) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result,
 			return ctrl.Result{Requeue: true}, b.targetDirectClient.Status().Update(ctx, &bundle)
 		}
 
-		if synced {
+        secretSynced := false
+        if bundle.Spec.Target.Secret != nil {
```
SgtCoDFish, line 206: suggestion: Since Secrets and ConfigMaps as targets would be mutually exclusive, we should probably avoid doing the sync with the ConfigMap above if we have a Secret target set. i.e., above we'd add something like `if bundle.Spec.Target.ConfigMap != nil`

pkg/bundle/sync.go
```diff
@@ -223,11 +224,100 @@ func (b *bundle) syncTarget(ctx context.Context, log logr.Logger,
...
+
+	// Here, the secret exists, but the selector doesn't match the namespace.
+	if !matchNamespace {
+		// The ConfigMap is owned by this controller- delete it.
+		if metav1.IsControlledBy(&secret, bundle) {
+			log.V(2).Info("deleting bundle from Namespace since namespaceSelector does not match")
+			return true, b.targetDirectClient.Delete(ctx, &secret)
+		}
+		// The ConfigMap isn't owned by us, so we shouldn't delete it. Return that
+		// we did nothing.
```
SgtCoDFish, line 290: suggestion: copy paste error!
```suggestion
		// The Secret isn't owned by us, so we shouldn't delete it. Return that
		// we did nothing.
```

pkg/bundle/sync.go
```diff
@@ -223,11 +224,100 @@ func (b *bundle) syncTarget(ctx context.Context, log logr.Logger,
...
+		b.recorder.Eventf(&secret, corev1.EventTypeWarning, "NotOwned", "Secret is not owned by trust.cert-manager.io so ignoring")
+		return false, nil
+	}
+
+	var needsUpdate bool
+	// If ConfigMap is missing OwnerReference, add it back.
```
SgtCoDFish, line 296: suggestion: copy paste!

pkg/bundle/sync.go
```diff
@@ -223,11 +224,100 @@ func (b *bundle) syncTarget(ctx context.Context, log logr.Logger,
...
+	// Exit early if no update is needed
+	if !needsUpdate {
+		return false, nil
+	}
+
+	if err := b.targetDirectClient.Update(ctx, &secret); err != nil {
+		return true, fmt.Errorf("failed to update secret %s/%s with bundle: %w", namespace, bundle.Name, err)
+	}
+
+	log.V(2).Info("synced bundle to namespace")
```
SgtCoDFish, line 320: suggestion: should this log line mention that it was a secret that was synced? likewise for the configmap equivalent!

pkg/apis/trust/v1alpha1/types_bundle.go
```diff
@@ -96,6 +96,10 @@ type BundleTarget struct {
 	// data will be synced to.
 	ConfigMap *KeySelector `json:"configMap,omitempty"`
 
+    // Secret is the target Secret in Namespaces that all Bundle source
+    // data will be synced to.
+    Secret *KeySelector `json:"secret,omitempty"`
+
```
SgtCoDFish, line 102: suggestion: We'll probably need to add validation to the [webhook](https://github.com/cert-manager/trust-manager/blob/main/pkg/webhook/validation.go) to ensure that people don't set both ConfigMap and Secret. Currently the webhook validation requires that all Bundles have a configMap set, so this PR would force users to set a configMap target and _optionally_ also a secret.

deploy/charts/trust-manager/values.yaml
```diff
@@ -67,6 +67,8 @@ app:
     # -- If false, disables the default seccomp profile, which might be required to run on certain platforms
     seccompProfileEnabled: true
 
+authorizedSecrets: ['ca-bundle']
+
```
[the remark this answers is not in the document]
rudexi: Added documentation

pkg/bundle/bundle.go
rudexi: I agree. Did the necessary changes.

pkg/bundle/sync.go
```diff
@@ -210,7 +211,7 @@ func (b *bundle) syncTarget(ctx context.Context, log logr.Logger,
 	}
 
 	// Match, return do nothing
-	if cmdata, ok := configMap.Data[target.ConfigMap.Key]; !ok || cmdata != data {
+	if cmdata, ok2 := configMap.Data[target.ConfigMap.Key]; !ok2 || cmdata != data {
```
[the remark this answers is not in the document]
rudexi: Not intentional, thanks for catching that. A piece of code I had before was also defining ok and err in the same scope, so I tried to work around it.

pkg/bundle/sync.go
```diff
@@ -223,11 +224,100 @@ func (b *bundle) syncTarget(ctx context.Context, log logr.Logger,
 		return false, nil
 	}
 
-	if err := b.targetDirectClient.Update(ctx, &configMap); err != nil {
+	if err = b.targetDirectClient.Update(ctx, &configMap); err != nil {
```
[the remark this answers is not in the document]
rudexi: I agree. Fixed it.

pkg/bundle/sync.go
rudexi: Fixed.

pkg/bundle/sync.go
rudexi: Fixed.

pkg/bundle/sync.go
rudexi: Done

pkg/bundle/bundle.go
rudexi: There is a misunderstanding on the intent.
I did not intend to have `target.secret` and `target.configmap` mutually exclusive.
In fact, in my usecase, I want both targets populated by the same bundle.

If you feel this should be mutually exclusive to simplify error handling, then I can do the necessary changes.

pkg/apis/trust/v1alpha1/types_bundle.go
rudexi: See comment regarding secrets and configmap being mutually exclusive.
I did not intend to change the current behavior of the spec, just doing an additive change that could be later refactored into another version of the API.

pkg/bundle/bundle.go
SgtCoDFish: Even if we don't make them mutually exclusive, we'll still need to change the validation because it requires that a configmap is set. That would make it impossible to have a Bundle which only targets a secret.

See this line: https://github.com/cert-manager/trust-manager/blob/60150c86bd873c5fa02752b63723920e4bf70b72/pkg/webhook/validation.go#L168-L169

SgtCoDFish (commented):
Few more comments - I'm thinking hard about mutual exclusivity!

SgtCoDFish:
We discussed this PR this morning in our standup and I think we agreed that it would be fine to have a ConfigMap and a Secret together, or just a Secret or just a ConfigMap. Users would have to specify at least one target.

We also think that we're likely to need to list and watch all Secrets so we can reconcile over them (but we should only need permission to get, delete and write secrets with a certain name, as this PR implements. Likely we'd need changes to `pkg/bundle/controller.go` to add that!

SgtCoDFish:
> Hello, what is the status on this, as there are no new updates in the last month. Can I support somehow? Would really love to have that feature!

No updates really - I've not really got the bandwidth currently to pick this up and I've got some leave coming up. If you wanted to try and pick it up and address comments, I'd try to review!

If you do pick it up I'd encourage you to drop a link to your PR in #cert-manager-dev on [Kubernetes slack](https://cert-manager.io/docs/contributing/#slack) just because like I said I've got some leave coming up!

morozovcookie:
Hello, I am also interested in this feature. 
What kind of help I can provide to merge this feature as soon as possible?

erikgb:
@rudexi Are you able to rebase this PR and address the comments so far? I am also interested in this feature. Please let us know if someone else should pick it up! :-)

Jiawei0227:
let me take a stab and rebase it and will open up a new PR

Jiawei0227:
The conflict is too large that I basically rewrite the whole thing.. anyways. https://github.com/cert-manager/trust-manager/pull/193 here is the PR and I tested in my local cluster it works

inteon:
Thanks @Jiawei0227, I propose we use this PR as reference and continue development in https://github.com/cert-manager/trust-manager/pull/193.

PR #109, opened 2023-02-02
Add missing permission to update bundle finalizers

SgtCoDFish:
This should enable trust-manager to work in OpenShift!

Based on a [slack thread](https://kubernetes.slack.com/archives/C4NV3DWUC/p1675277465934999) which reported the issue

### DIFF
--- a/deploy/charts/trust-manager/templates/clusterrole.yaml
+++ b/deploy/charts/trust-manager/templates/clusterrole.yaml
@@ -10,6 +10,15 @@ rules:
   resources:
   - "bundles"
   verbs: ["get", "list", "watch"]
+
+# Permissions to update finalizers are required for trust-manager to work correctly
+# on OpenShift, even though we don't directly use finalizers at the time of writing
+- apiGroups:
+  - "trust.cert-manager.io"
+  resources:
+  - "bundles/finalizers"
+  verbs: ["update"]
+
 - apiGroups:
   - "trust.cert-manager.io"
   resources:
@@ -21,6 +30,7 @@ rules:
   resources:
   - "configmaps"
   verbs: ["get", "list", "create", "update", "watch", "delete"]
+
 - apiGroups:
   - ""
   resources:
--- a/test/kind-cluster.yaml
+++ b/test/kind-cluster.yaml
@@ -1,0 +1,16 @@
+# This configuration is used to set up a Kind cluster for smoke tests or demo environments.
+
+kind: Cluster
+apiVersion: kind.x-k8s.io/v1alpha4
+# WARNING: Makefile commands which interact with the cluster _require_ that it be called "trust"
+# Changing the name here will stop the cluster being able to be controlled via Make!
+name: trust
+nodes:
+- role: control-plane
+  # Enable OwnerReferencesPermissionEnforcement to better match OpenShift environments in tests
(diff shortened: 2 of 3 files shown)


wallrj (changes requested):
Thanks @SgtCoDFish 

Please add a comment in the YAML explaining why the finalizers permission is required,
otherwise a future maintainer may come along and remove on the grounds that finalizers are not used directly by trust-manager.

Consider enabling the https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#ownerreferencespermissionenforcement admission controller in the E2E tests, so that we can see this working.

SgtCoDFish:
> Consider enabling the https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#ownerreferencespermissionenforcement admission controller in the E2E tests, so that we can see this working.

This is a great review comment, thanks. Would've have thought to do this. I added it (and removed the new permission) and saw the error after running `kind delete clusters trust && make smoke`:

```text
I0202 12:08:27.196772       1 recorder.go:103] trust/manager/events "msg"="Warning" 

"message"="Failed to sync target in Namespace \"kube-public\": configmaps \"test-bundle-nwwtg\" is forbidden:
cannot set blockOwnerDeletion if an ownerReference refers to a resource you can't set finalizers on: , <nil>"

"object"={"kind":"Bundle","name":"test-bundle-nwwtg","uid":"e0057bbf-d164-4865-8624-35d059489dc2","apiVersion":"trust.cert-manager.io/v1alpha1","resourceVersion":"1369"}

"reason"="SyncTargetFailed"
```

I re-added the permission and it passed

SgtCoDFish:
/test pull-trust-manager-verify

flake!

deploy/charts/trust-manager/templates/clusterrole.yaml
```diff
@@ -10,6 +10,15 @@ rules:
   resources:
   - "bundles"
   verbs: ["get", "list", "watch"]
+
+# Permissions to update finalizers are required for trust-manager to work correctly
+# on OpenShift, even though we don't directly use finalizers at the time of writing
```
wallrj, line 15: 👍

test/kind-cluster.yaml
```diff
@@ -0,0 +1,16 @@
+# This configuration is used to set up a Kind cluster for smoke tests or demo environments.
+
+kind: Cluster
+apiVersion: kind.x-k8s.io/v1alpha4
+# WARNING: Makefile commands which interact with the cluster _require_ that it be called "trust"
+# Changing the name here will stop the cluster being able to be controlled via Make!
+name: trust
+nodes:
+- role: control-plane
+  # Enable OwnerReferencesPermissionEnforcement to better match OpenShift environments in tests
```
wallrj, line 10: 👍

wallrj (approved):
Thanks. I had a doubt about whether the value of `--enable-admission-plugins` appends to the default set of plugins or whether it replaces them, but [the docs say](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/):
> admission plugins that should be enabled in addition to default enabled ones

...and I suppose the cluster and the tests would have failed if the defaults weren't also enabled.

/lgtm

PR #114, opened 2023-02-03
Bump values in helm chart

SgtCoDFish:
This should've been done before the tag + release of `v0.4.0` but was missed. We agreed in standup on 2023-02-03 that it was fine to release the container images built from the v0.4.0 tag and use this commit for releasing the chart.


irbekrm:
I guess you need to run `make update-helm-docs`?

Also this doesn't look right https://github.com/cert-manager/trust-manager/blob/main/hack/update-helm-docs.sh#L28

irbekrm:
Thank you @SgtCoDFish 

/lgtm
/approve

PR #115, opened 2023-02-14
Add verify-versions script

SgtCoDFish:
During the release of trust-manager 0.4.0 a version update was missed (see #114 ). This adds a simple script which should ensure that if one value is updated, they all are.

Grepping through YAML isn't really ideal, but I didn't want to add another dependency if it could be avoided.

Also makes a minor tweak to the boilerplate checker, following changes present in other cert-manager subprojects.

### DIFF
--- a/hack/verify-versions.sh
+++ b/hack/verify-versions.sh
@@ -1,0 +1,60 @@
+#!/usr/bin/env bash
+
+# Copyright 2023 The cert-manager Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+MAKEFILE_RELEASE_VERSION=$(grep -E "^RELEASE_VERSION" Makefile | awk '{print $3}')
+
+# This is a tortured way to avoid installing yq
+HELM_CHART_VALUES_VERSION=$(grep -E -A1 "Target image version" deploy/charts/trust-manager/values.yaml | grep "tag:" | awk '{print $2}')
+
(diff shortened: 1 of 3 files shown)


hack/verify-versions.sh
```diff
@@ -0,0 +1,60 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+MAKEFILE_RELEASE_VERSION=$(grep -E "^RELEASE_VERSION" Makefile | awk '{print $3}')
+
+# This is a tortured way to avoid installing yq
```
irbekrm, line 23: :)

hack/verify-versions.sh
```diff
@@ -0,0 +1,60 @@
...
+HELM_CHART_VALUES_VERSION=$(grep -E -A1 "Target image version" deploy/charts/trust-manager/values.yaml | grep "tag:" | awk '{print $2}')
+
+HELM_CHART_VERSION=$(grep -E "^version:" deploy/charts/trust-manager/Chart.yaml | awk '{print $2}')
+
+# We might not always want appVersion to match version, but for now it probably
+# makes sense to enforce that they remain the same
+HELM_CHART_APP_VERSION=$(grep -E "^appVersion:" deploy/charts/trust-manager/Chart.yaml | awk '{print $2}')
```
irbekrm, line 30: Question: what is the value of doing this here? I quite like that in the other projects if a Helm PR has been merged, we can quickly cut a new chart version with that change, yet don't need to cut another app release with no actual code changes.

Because historically we've done the above on OS projects that aren't cert-manager and because occassionally we need to make the assumptions on whether chart and app version will always be the same or not for projects, I am a bit concerned that we're introducing a differing practice here

irbekrm (commented):
I've some questions about this one. I'm not convinced that adding this check in just this one project is a good idea for consistency. I also quite like that we can quickly cut new releases for charts only if needed whilst a project is still in alpha

hack/verify-versions.sh
SgtCoDFish: I think I'm kinda OK with this differing here because it would be clear when someone raised a PR which failed this test what had happened.

I figure also that _most_ of the time, releasing a new version of the Helm chart might as well coincide with a new version of the application to include updated dependencies / base images even if there aren't any code changes.

Put another way: This stops us (or, well, me :grin:) making the same mistake again - and if it causes a problem it's also very easy to remove it. I think it's kinda worth it,  but I'm not really inclined to massively fight for it either - it's not a big deal!

SgtCoDFish:
I guess this doesn't _need_ to be merged, will close!

PR #118, opened 2023-04-01
Make seccompProfile optional in initContainer

aelbarkani:
When the `initcontainer` was added, we forgot to add the option to disable seccompProfile for that container.
This PR fixes that issue.


inteon:
I rebased the PR on master.

PR #121, opened 2023-04-13
helm: make installation of default trust package optional

siiimooon:
The init container is a bit noisy when distributing in-house trust in environments with no internet connection.


deploy/charts/trust-manager/Chart.yaml
```diff
@@ -17,4 +17,4 @@ sources:
 - https://github.com/cert-manager/trust-manager
 
 appVersion: v0.4.0
-version: v0.4.0
+version: v0.4.1
```
SgtCoDFish, line 20: suggestion: I'd probably prefer that we don't change this here, and instead leave this to be done in a separate PR just before a release. Doesn't really matter though because it'll get updated anyway - as long as the tests pass I'm happy to merge

SgtCoDFish (commented):
Happy to merge this once tests pass! Thanks very much this - great idea, great execution :ok_hand:

SgtCoDFish (approved):
/lgtm
/approve

Thanks, great contribution! :rocket:  Hopefully we should be able to get this into v0.5.0 which I'm hoping to release in the next couple of weeks!

Issue #17, opened 2022-04-12
Support distribution of PKCS12/JKS truststores


erikgb:
We are using cert-manager to issue short-lived certificates to Kafka-clients running in our Kubernetes cluster. Currently we use ca.crt truststore.jks or truststore.p12 in the certificate Secret to establish trust to the Kafka cluster (mTLS). This is wrong, and only works because the client and server have certificates issued by the same root CA. 😄 

To fix this, we were hoping to use this cert-manager component to distribute the trust. But since a great percentage of our clients are Java-based, we need to be able to provide the trust bundle in PKCS12/JKS format. Would it be possible to add an opt-in to provide the trust bundle in additional formats?

yrro:
You might find [cert-utils](https://github.com/redhat-cop/cert-utils-operator#Creating-java-keystore-and-truststore) useful - you can annotate a secret containing a certificate & it will add keystore & truststore keys to the secret.

erikgb:
> You might find [cert-utils](https://github.com/redhat-cop/cert-utils-operator#Creating-java-keystore-and-truststore) useful - you can annotate a secret containing a certificate & it will add keystore & truststore keys to the secret.

Well, cert-utils could be useful for other use-cases, but not for this IMO. I don't see the point in converting N Secrets with the same content. And I do not want to install additional software into my clusters. Especially not software that require RBAC to read secrets cluster-wide. And cert-manager already has native support for JKS and PKCS12 keystores: https://cert-manager.io/docs/reference/api-docs/#cert-manager.io/v1.CertificateKeystores. And I think this support should/could be implemented here also.

sgran:
is it not implemented here?  If you create a Bundle targeting the truststore in your CA, it gets distributed.

SpectralHiss:
Hey it does work to distribute anything you feed it, but this is not first class support as you still need to whip out `keytool` to create the java trust store.
I assume first-class support means you can feed it a PEM as input and it would generate the JKS and PKCS12 trust store, similarly to how cert-manager can now generate JKS and pkCS12 keystores in the certificate secret.

bakito:
@SpectralHiss I implemented a webhook that is capable to transform pem files into a truststore and add it to the same configmap. https://github.com/bakito/java-truststore-injection-webhook

PR #122, opened 2023-04-17
Add support for JKS truststores

aidy:
* Adds an AdditionalFormats option to target
 * Writes jks encoded truststore as binaryData in configmap

connects #17


pkg/bundle/sync.go
```diff
@@ -31,6 +34,11 @@ import (
 
 	trustapi "github.com/cert-manager/trust-manager/pkg/apis/trust/v1alpha1"
 	"github.com/cert-manager/trust-manager/pkg/util"
+	jks "github.com/pavlo-v-chernykh/keystore-go/v4"
+)
+
+const (
+	defaultJksPassword = "changeit"
```
maelvls, line 41: I don't use the JKS format much, but I imagine that we don't really care about the password for this specific keystore (it will never contain private key material)

Is there a possibility to set the password to `""`? 

Otherwise, the hardcoded `"changeit"` may be fine as long as we document it

aidy: Pretty sure it needs some kind of password, didn't try to see if an empty string was acceptable.

As far as I'm aware, java apps kinda expect `changeit` as a default - I think this is the least painful thing to set it to.
You're right though, there should be a comment at least to explain.

maelvls: Aah great, if changeit is a defacto standard then people won't be surprised! 👌

aidy: > Aah great, if "changeit" is a defacto standard then people won't be surprised!

That's my understanding from poking around, but I'm no java dev :)

pkg/apis/trust/v1alpha1/types_bundle.go
```diff
@@ -95,13 +95,21 @@ type BundleTarget struct {
...
 
 	// NamespaceSelector will, if set, only sync the target resource in
 	// Namespaces which match the selector.
 	// +optional
 	NamespaceSelector *NamespaceSelector `json:"namespaceSelector,omitempty"`
 }
 
+// AdditionalFormats specifies any additional formats to write to the configmap
+type AdditionalFormats struct {
+	Jks *KeySelector `json:"jks,omitempty"`
```
SgtCoDFish, line 110: nitpick: As mentioned in our call, I'd probably choose to have the Golang field be `JKS` and to keep the JSON key as `jks`.

pkg/apis/trust/v1alpha1/types_bundle.go
```diff
@@ -95,13 +95,21 @@ type BundleTarget struct {
 	// ConfigMap is the target ConfigMap in Namespaces that all Bundle source
 	// data will be synced to.
 	ConfigMap *KeySelector `json:"configMap,omitempty"`
+	// AdditionalFormats specifies any additional formats to write to the configmap
+	// +optional
```
SgtCoDFish, line 99: nitpick: We might add support for other targets, so it'd be good to avoid talking about ConfigMaps here!

```suggestion
	// AdditionalFormats specifies any additional formats to write to the target
	// +optional
```

pkg/apis/trust/v1alpha1/types_bundle.go
```diff
@@ -95,13 +95,21 @@ type BundleTarget struct {
...
+	AdditionalFormats *AdditionalFormats `json:"additionalFormats,omitempty"`
 
 	// NamespaceSelector will, if set, only sync the target resource in
 	// Namespaces which match the selector.
 	// +optional
 	NamespaceSelector *NamespaceSelector `json:"namespaceSelector,omitempty"`
 }
 
+// AdditionalFormats specifies any additional formats to write to the configmap
```
SgtCoDFish, line 108: nitpick: as above!

```suggestion
// AdditionalFormats specifies any additional formats to write to the target
```

pkg/bundle/sync.go
SgtCoDFish: I've seen at least one major system using certs with Java that used `changeit` as a default. I'm happy to have that be the default here!

pkg/bundle/sync_test.go
```diff
@@ -362,6 +463,25 @@ func Test_syncTarget(t *testing.T) {
...
+					assert.Len(t, entryNames, 1)
+					assert.True(t, ks.IsTrustedCertificateEntry(entryNames[0]))
+					// Safe to ignore errors here, we've tested that it's present and a TrustedCertificateEntry
+					cert, _ := ks.GetTrustedCertificateEntry(entryNames[0])
+					// Only one certificate block for this test, so we can safely ignore the `remaining` byte array
+					p, _ := pem.Decode([]byte(data))
+					assert.Equal(t, p.Bytes, cert.Certificate.Content)
+				} else {
+					assert.False(t, jksExists, "jks should not be present in configmap")
+				}
```
SgtCoDFish, line 484: suggestion: Would it be cleaner for us to `assert.Equal` on `jksExists` rather than having the else block? I think that would more closely match the usual style I expect from Golang

```suggestion

				jksData, jksExists := configMap.BinaryData[jksKey]
				assert.Equal(t, test.expJKS, jksExists)

				if test.expJks {
					reader := bytes.NewReader(jksData)
					ks := jks.New()
					err := ks.Load(reader, []byte(defaultJksPassword))
					assert.Nil(t, err)
					entryNames := ks.Aliases()
					assert.Len(t, entryNames, 1)
					assert.True(t, ks.IsTrustedCertificateEntry(entryNames[0]))
					// Safe to ignore errors here, we've tested that it's present and a TrustedCertificateEntry
					cert, _ := ks.GetTrustedCertificateEntry(entryNames[0])
					// Only one certificate block for this test, so we can safely ignore the `remaining` byte array
					p, _ := pem.Decode([]byte(data))
					assert.Equal(t, p.Bytes, cert.Certificate.Content)
				}
```

pkg/bundle/sync_test.go
```diff
@@ -69,6 +78,16 @@ func Test_syncTarget(t *testing.T) {
...
+		"if object doesn't exist with jks, expect update": {
+			object:            nil,
+			namespace:         corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: "test-namespace"}},
+			selector:          labelEverything,
+			withJks:           true,
+			expExists:         true,
+			expJks:            true,
+			expOwnerReference: true,
+			expNeedsUpdate:    true,
+		},
```
SgtCoDFish, line 90: suggestion: I don't know from these tests what would happen if I were to specify a target where the JKS key is (accidentally) the same as the key for the PEM trust store, i.e.:

```yaml
  target:
    configMap:
      key: "root-certs.pem"
    additionalFormats:
      jks:
        key: "root-certs.pem"
```

I'd expect we should error in this case. Could we handle it and add a test for it?

SgtCoDFish (changes requested):
Got a few comments - this is really awesome though and it's so nearly there!

pkg/bundle/sync_test.go
aidy: I expect that we'd get an error from kubernetes for attempting to write an invalid configmap, but I agree - we should catch that ahead of time if we can.

pkg/bundle/sync_test.go
aidy: Yes, that's much cleaner - thanks!

pkg/apis/trust/v1alpha1/types_bundle.go
aidy: I think I've renamed this everywhere appropriate

pkg/bundle/sync.go
```diff
@@ -162,6 +200,17 @@ func (b *bundle) syncTarget(ctx context.Context, log logr.Logger,
 	var configMap corev1.ConfigMap
 	err := b.targetDirectClient.Get(ctx, client.ObjectKey{Namespace: namespace.Name, Name: bundle.Name}, &configMap)
 
+	if target.AdditionalFormats != nil && target.AdditionalFormats.JKS != nil {
+		if target.AdditionalFormats.JKS.Key == target.ConfigMap.Key {
+			return false, errors.New("invalid target (duplicate key)")
+		}
```
SgtCoDFish, line 206: suggestion: I should've suggested this before, my bad - maybe we could move this check to the webhook? (https://github.com/cert-manager/trust-manager/blob/main/pkg/webhook/validation.go)

That would mean that when users make the change it'll error, rather than when we attempt to Sync it.

aidy: Yep - that's a better place for it. :)

xxmaestroxx:
any idea when a release of this feature will be published, we are exactly on this point atm to iterate of the ".pem" file and create by hand an own truststore in all our apps, by having this feature available it would take a lot of duplicate init code from the app side. 

Appreciate the work on this!

SgtCoDFish:
I can't give you a date for certain, but I intend to test this within the next couple of days and then cut a release. Hopefully should be released within a week, but with holidays and other work it could be delayed.

I'll [tweet](https://twitter.com/CertManager/) and [toot](https://infosec.exchange/@certmanager) about it when we release!

xxmaestroxx:
thanks for the fast feedback! we will be patient! 

If you have a test version available (RC) and we can help testing it in a real use case scenario, let me know.

SgtCoDFish:
https://github.com/cert-manager/trust-manager/releases/tag/v0.5.0-beta.0

I've published images and a helm chart for v0.5.0-beta.0! Please feel free to test and let me know how it goes :grin: 

Checking out for the day now but will check back in tomorrow!

pkg/bundle/sync.go
```diff
@@ -141,6 +150,34 @@ func (b *bundle) secretBundle(ctx context.Context, ref *trustapi.SourceObjectKey
 	return string(data), nil
 }
 
+func encodeJKS(trustbundle string) ([]byte, error) {
+	remaining := []byte(trustbundle)
+	ks := jks.New()
```
vinzent, line 155: @aidy Sorry to chime in late.  I recently fixed an issue in redhat-cop/cert-utils-operator where the JKS store was updated even when no certs changed. One part of the issue was the missing  `jks.WithOrderedAliases()` option of the jks lib. Once there were many certificates (example: mozilla ca-bundle) in the truststore, the newly generated file was no more binary equivalent to the last one, which lead to updates of ConfigMaps which triggered app restarts due to them listening for changes of the configmaps.



https://github.com/vinzent/cert-utils-operator/blob/cdcb7a30b1bf44a269beeeb016e7ca93af270bd9/controllers/configmaptokeystore/configmap_to_keystore_controller.go#L129-L131

aidy: Ah. I'd spotted that the output varied, I'd assumed that it was due to including randomness during encoding (as it took a password and you could inject your own crypto).

In any case, we should not have a problem with constant configmap refreshes here, as we don't check for binary equivalence:
https://github.com/cert-manager/trust-manager/pull/122/files#diff-4def846e69c9bdd8e89ae4c9d881e4bfaa70c20b7b024cad43f71268c5f2af77R271-R274

SgtCoDFish: @vinzent thank you for that, I didn't know of that option! You're a star!

I think it seems worthwhile adding that for the actual release. More determinism is good. I'll raise a PR for it :grin:

vinzent: for the records, these are the comments where I learned about this:

https://github.com/pavlo-v-chernykh/keystore-go/pull/23#issuecomment-726298997

https://github.com/pavlo-v-chernykh/keystore-go/pull/23#issuecomment-726347215


it works for certificates but not for private-keyse where a random salt is added  every time.

SgtCoDFish:
> line 166 in sync.go: the key used for entries is c.Issuer.String(), but a self signed certificate and an intermediate one in the chain can have the same Issuer. Would it be possible to use the Subject of a certificate, which I believe should ensure uniqueness for all variations?

Oh wow, I really should've caught that one! Nice debugging and thank you so much for testing :grin:

I don't think we can (safely) use the subject because it's totally possible to issue two certs with the same subject :thinking:  I'm thinking we should use the subject (for human readability) with a small hash of the whole certificate appended.

I'll address that and look to release v0.5.0-beta.1 for testing!

hazmat345:
Came here to report seeing some intermediate CAs missing in the jks, but looks like I don't need to. You all are awesome 😄

SgtCoDFish:
@claudiuavat1 @hazmat345 thanks to you both for your debugging efforts. I've just released v0.5.0-beta.1 which should fix the JKS bug - https://github.com/cert-manager/trust-manager/releases/tag/v0.5.0-beta.1

I'm now going to be away until next week, but I'll check in again then. If you're able to test v0.5.0-beta.1 and it works for you, I'll be sure to release v0.5.0 as soon as I can after that :rocket:

xxmaestroxx:
We will try the beta1 and will give you feedback beginning of next week.

xxmaestroxx:
We tried it and currently it is not working: 

the following error appears:

pod: E0511 13:28:40.240929 1 controller.go:329] trust/manager "msg"="Reconciler error" "error"="Operation cannot be fulfilled on bundles.trust.cert-manager.io \"test-trust\": the object has been modified; please apply your changes to the latest version and try again" "controller"="bundles" "name"="test-trust" "namespace"="" "reconcileID"="26653c63-b346-4d7a-a85e-a70a9bbf9baf"

and no configmap with an jks store will be created. 

Thats our "Bundle" CR:
`
kind: Bundle 
metadata: 
  name: test-trust 
spec: 
  sources: 
    - useDefaultCAs: true 
  target: 
    additionalFormats: 
      jks: 
        key: test.jks 
    configMap: 
      key: 
        test.pem
`

Logging of the CR:

`
I0511 13:28:37.838305 1 recorder.go:103] trust/manager/events "msg"="Successfully synced Bundle to all namespaces" "object"={"kind":"Bundle","name":"test-trust","uid":"64cb2b9d-9919-498e-9d70-955982389b56","apiVersion":"trust.cert-manager.io/v1alpha1","resourceVersion":"21536529"} "reason"="Synced" "type"="Normal"


I0511 13:28:37.838354 1 recorder.go:103] trust/manager/events "msg"="Successfully synced Bundle to all namespaces" "object"={"kind":"Bundle","name":"test-trust","uid":"64cb2b9d-9919-498e-9d70-955982389b56","apiVersion":"trust.cert-manager.io/v1alpha1","resourceVersion":"21536529"} "reason"="Synced" "type"="Normal"


I0511 13:28:40.232675 1 recorder.go:103] trust/manager/events "msg"="Successfully synced Bundle to all namespaces" "object"={"kind":"Bundle","name":"test-trust","uid":"64cb2b9d-9919-498e-9d70-955982389b56","apiVersion":"trust.cert-manager.io/v1alpha1","resourceVersion":"21536529"} "reason"="Synced" "type"="Normal"


I0511 13:28:40.232736 1 recorder.go:103] trust/manager/events "msg"="Successfully synced Bundle to all namespaces" "object"={"kind":"Bundle","name":"test-trust","uid":"64cb2b9d-9919-498e-9d70-955982389b56","apiVersion":"trust.cert-manager.io/v1alpha1","resourceVersion":"21536529"} "reason"="Synced" "type"="Normal"


E0511 13:28:40.240929 1 controller.go:329] trust/manager "msg"="Reconciler error" "error"="Operation cannot be fulfilled on bundles.trust.cert-manager.io \"test-trust\": the object has been modified; please apply your changes to the latest version and try again" "controller"="bundles" "name"="test-trust" "namespace"="" "reconcileID"="26653c63-b346-4d7a-a85e-a70a9bbf9baf"


I0511 13:40:56.192917 1 recorder.go:103] trust/manager/events "msg"="Successfully synced Bundle to all namespaces" "object"={"kind":"Bundle","name":"test-trust","uid":"64cb2b9d-9919-498e-9d70-955982389b56","apiVersion":"trust.cert-manager.io/v1alpha1","resourceVersion":"21541615"} "reason"="Synced" "type"="Normal"


I0511 13:40:56.192958 1 recorder.go:103] trust/manager/events "msg"="Successfully synced Bundle to all namespaces" "object"={"kind":"Bundle","name":"test-trust","uid":"64cb2b9d-9919-498e-9d70-955982389b56","apiVersion":"trust.cert-manager.io/v1alpha1","resourceVersion":"21541615"} "reason"="Synced" "type"="Normal"
`

xxmaestroxx:
Sorry for the broken indentation, somehow the code block is wrong, not sure what I'm doing wrong.

SgtCoDFish:
Sorry it took me a while to get back to this - been a busy few days!

I can't reproduce the issue you're showing - in fact, I can't actually see a problem in any of the logs you posted.

I just created a test cluster and a dummy bundle using the JSON from https://github.com/cert-manager/trust-manager/releases/tag/v0.5.0-beta.1 and it worked exactly as I'd expect, with a valid JKS file.

We see the reconciler error you posted a fair bit but usually that's no problem - usually it'll end up working after you see that message. Are you able to test again?

SgtCoDFish:
I've now released trust-manager v0.5.0: https://github.com/cert-manager/trust-manager/releases/tag/v0.5.0 :tada: 

I couldn't see a problem with how this was working when I tested locally so I figured it was best to get it out there.

If anyone finds any issues with the new JKS functionality (or any issues generally!) we can release v0.5.1 easily enough :grin: 

Thanks to everyone who was involved with this! :rocket:

PR #123, opened 2023-04-23
Explicitly define resource namespaces in Helm chart

inteon:
This is very useful in combination with Helm template. It makes sure that all templated resources have the namespace set. This way, it is not possible to `k apply -f` in a different namespace; potentially breaking some of our references that contain namespace names.


SgtCoDFish (approved):
/lgtm
/approve

Wow, didn't realise there was so much discussion around this! (https://github.com/helm/helm/issues/5465). Happy to do this though, thanks!

PR #124, opened 2023-04-26
Minor restructure for ko compatibility

aidy:
Move cmd/main.go to cmd/trust-manager/main.go so that `ko` correctly builds as a `trust-manager` binary. This brings it into line with other go project structures.

This is a noop as far as the current build mechanism is concerned - it should generate the same thing in the same way, but this restructure facilitates an easier development/test cycle.


SgtCoDFish (approved):
/lgtm
/approve

Even without thinking about `ko` I prefer this structure. Thanks very much!

PR #125, opened 2023-04-27
Bump to v0.5.0-beta.0 for release

SgtCoDFish:


SgtCoDFish:
>  Ideally, we wouldn't have to change this version in the repo.

Agreed, this could be tag based as I've done in https://github.com/cert-manager/boilersuite (and cert-manager).

Harder to get around it for the helm charts, though. Still, now's probably not the time for me to start shaving that yak :grin:

PR #129, opened 2023-05-02
Ensure unique aliases in JKS files

SgtCoDFish:
Each commit should have a descriptive commit message.

This PR is intended to address the issue raised in [this comment](https://github.com/cert-manager/trust-manager/pull/122#issuecomment-1529783592) where two certs with the same Issuer field would be de-duped when added to the JKS file, despite them being different certificates.

### DIFF
--- a/pkg/bundle/sync.go
+++ b/pkg/bundle/sync.go
@@ -177,6 +177,8 @@ func encodeJKS(trustBundle string, password []byte) ([]byte, error) {
 			return nil, fmt.Errorf("got invalid cert when trying to encode JKS: %w", err)
 		}
 
+		alias := jksAlias(c.Raw, c.Subject.String())
+
 		// Note on CreationTime:
 		// Debian's JKS trust store sets the creation time to match the time that certs are added to the
 		// trust store (i.e., it's effectively time.Now() at the instant the file is generated).
@@ -185,8 +187,6 @@ func encodeJKS(trustBundle string, password []byte) ([]byte, error) {
 		// - Using something from the cert being added (e.g. NotBefore / NotAfter)
 		// - Using a fixed time (i.e. unix epoch)
 		// We use NotBefore here, arbitrarily.
-
-		alias := jksAlias(c)
 
 		err = ks.SetTrustedCertificateEntry(alias, jks.TrustedCertificateEntry{
 			CreationTime: c.NotBefore,
@@ -212,13 +212,14 @@ func encodeJKS(trustBundle string, password []byte) ([]byte, error) {
 	return buf.Bytes(), nil
 }
 
-// jksAlias creates a JKS-safe alias for the given certificate, such that
-// any two certificates will have a different aliases unless they're
-// identical in every way.
-// This unique alias fixes an issue where we used the Issuer field as an
-// alias, leading to different certs being treated as identical.
-func jksAlias(cert *x509.Certificate) string {
-	certHashBytes := sha256.Sum256(cert.Raw)
+// jksAlias creates a JKS-safe alias for the given DER-encoded certificate, such that
+// any two certificates will have a different aliases unless they're identical in every way.
+// This unique alias fixes an issue where we used the Issuer field as an alias, leading to
+// different certs being treated as identical.
+// The friendlyName is included in the alias as a UX feature when examining JKS files using a
+// tool like `keytool`.
+func jksAlias(derData []byte, friendlyName string) string {
+	certHashBytes := sha256.Sum256(derData)
 	certHash := hex.EncodeToString(certHashBytes[:])
 
 	// Since certHash is the part which actually distinguishes between two
@@ -226,7 +227,7 @@ func jksAlias(cert *x509.Certificate) string {
 	// with a really long subject is added. Not sure what the upper limit
 	// for length actually is, but it shouldn't matter here.
 
-	return certHash[:8] + "|" + cert.Subject.String()
+	return certHash[:8] + "|" + friendlyName
 }
 
 // syncTarget syncs the given data to the target ConfigMap in the given namespace.
--- a/pkg/bundle/sync.go
+++ b/pkg/bundle/sync.go
@@ -19,13 +19,16 @@ package bundle
 import (
 	"bytes"
 	"context"
+	"crypto/sha256"
 	"crypto/x509"
+	"encoding/hex"
 	"encoding/pem"
 	"errors"
 	"fmt"
 	"strings"
 
 	"github.com/go-logr/logr"
+	jks "github.com/pavlo-v-chernykh/keystore-go/v4"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -34,14 +37,13 @@ import (
 
 	trustapi "github.com/cert-manager/trust-manager/pkg/apis/trust/v1alpha1"
 	"github.com/cert-manager/trust-manager/pkg/util"
-	jks "github.com/pavlo-v-chernykh/keystore-go/v4"
 )
 
 const (
-	// defaultJKSPassword is the default password that Java uses; it's a Java convention to use this exact password.
-	// Since we're not storing anything secret in the JKS files we generate, this password is essentially irrelevant
-	// but is required anyway.
-	defaultJKSPassword = "changeit"
+	// DefaultJKSPassword is the default password that Java uses; it's a Java convention to use this exact password.
+	// Since we're not storing anything secret in the JKS files we generate, this password is not a meaningful security measure
+	// but seems often to be expected by applications consuming JKS files
+	DefaultJKSPassword = "changeit"
 )
 
 type notFoundError struct{ error }
@@ -152,9 +154,16 @@ func (b *bundle) secretBundle(ctx context.Context, ref *trustapi.SourceObjectKey
 	return string(data), nil
 }
 
-func encodeJKS(trustBundle string) ([]byte, error) {
+// encodeJKS creates a binary JKS file from the given PEM-encoded trust bundle and password.
+// Note that the password is not treated securely; JKS files generally seem to expect a password
+// to exist and so we have the option for one.
+func encodeJKS(trustBundle string, password []byte) ([]byte, error) {
 	remaining := []byte(trustBundle)
+
+	// WithOrderedAliases ensures that trusted certs are added to the JKS file in order,
+	// which makes the files appear to be reliably deterministic.
 	ks := jks.New(jks.WithOrderedAliases())
+
 	for len(remaining) > 0 {
 		var p *pem.Block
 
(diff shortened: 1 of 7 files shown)


pkg/bundle/jks.go
```diff
@@ -0,0 +1,110 @@
...
+	certHashBytes := sha256.Sum256(cert.Raw)
+	certHash := hex.EncodeToString(certHashBytes[:])
+
+	// Since certHash is the part which actually distinguishes between two
+	// certificates, put it first so that it won't be truncated if a cert
+	// with a really long subject is added. Not sure what the upper limit
+	// for length actually is, but it shouldn't matter here.
+
+	return certHash[:8] + "|" + cert.Subject.String()
+}
```
SgtCoDFish, line 110: note: this is the real core of the bug fix, and this is what fixes the reported issue.

For comparison, the Debian JKS file uses the filename that a trusted cert was taken from as the alias, which is obviously a unique measure. We don't have filenames, but this approach of using a hash seems to work well.

To check the JKS truststore on a debian based machine with Java installed:

```console
$ keytool -list -keystore /etc/ssl/certs/java/cacerts
<snip>
...
debian:verisign_universal_root_certification_authority.pem, 2 May 2023, trustedCertEntry,
Certificate fingerprint (SHA-256): 23:99:56:11:27:A5:71:25:DE:8C:EF:EA:61:0D:DF:2F:A0:78:B5:C8:06:7F:4E:82:82:90:BF:B8:60:E8:4B:3C
debian:xramp_global_ca_root.pem, 2 May 2023, trustedCertEntry,
Certificate fingerprint (SHA-256): CE:CD:DC:90:50:99:D8:DA:DF:C5:B1:D2:09:B7:37:CB:E2:C1:8C:FB:2C:10:C0:FF:0B:CF:0D:32:86:FC:1A:A2
```

pkg/bundle/jks.go
```diff
@@ -0,0 +1,110 @@
...
+	// DefaultJKSPassword is the default password that Java uses; it's a Java convention to use this exact password.
+	// Since we're not storing anything secret in the JKS files we generate, this password is not a meaningful security measure
+	// but seems often to be expected by applications consuming JKS files
+	DefaultJKSPassword = "changeit"
+)
+
+// encodeJKS creates a binary JKS file from the given PEM-encoded trust bundle and password.
+// Note that the password is not treated securely; JKS files generally seem to expect a password
+// to exist and so we have the option for one.
+func encodeJKS(trustBundle string, password []byte) ([]byte, error) {
```
SgtCoDFish, line 40: note: this function was moved from sync.go and has a few changes from the original, but the idea is essentially the same. Most of the changes here are adding comments.

test/env/data.go
```diff
@@ -288,3 +290,32 @@ func EventuallyBundleHasSyncedAllNamespacesStartsWith(ctx context.Context, cl cl
...
+		return err
+	}
+
+	expectedCertList, err := util.ValidateAndSplitPEMBundle([]byte(expectedCertPEMData))
+	if err != nil {
+		return fmt.Errorf("invalid PEM data passed to CheckJKSFileSynced: %s", err)
+	}
+
+	// TODO: check that the cert content matches expectedCertPEMData exactly, not just
+	// that the count is the same
```
SgtCoDFish, line 311: note: implementing this would take a bit more work. It's doable but I didn't want this PR to balloon so I skipped it for this fix.

pkg/bundle/sync.go
```diff
@@ -165,24 +174,61 @@ func encodeJKS(trustBundle string) ([]byte, error) {
...
+	certHash := hex.EncodeToString(certHashBytes[:])
+
+	// Since certHash is the part which actually distinguishes between two
+	// certificates, put it first so that it won't be truncated if a cert
+	// with a really long subject is added. Not sure what the upper limit
+	// for length actually is, but it shouldn't matter here.
+
+	return certHash[:8] + "|" + cert.Subject.String()
+}
+
```
SgtCoDFish, line 231: note: this is the real core of the bug fix, and this is what fixes the reported issue.

For comparison, the Debian JKS file uses the filename that a trusted cert was taken from as the alias, which is obviously a unique measure. We don't have filenames, but this approach of using a hash seems to work well.

To check the JKS truststore on a debian based machine with Java installed:

```console
$ keytool -list -keystore /etc/ssl/certs/java/cacerts
<snip>
...
debian:verisign_universal_root_certification_authority.pem, 2 May 2023, trustedCertEntry,
Certificate fingerprint (SHA-256): 23:99:56:11:27:A5:71:25:DE:8C:EF:EA:61:0D:DF:2F:A0:78:B5:C8:06:7F:4E:82:82:90:BF:B8:60:E8:4B:3C
debian:xramp_global_ca_root.pem, 2 May 2023, trustedCertEntry,
Certificate fingerprint (SHA-256): CE:CD:DC:90:50:99:D8:DA:DF:C5:B1:D2:09:B7:37:CB:E2:C1:8C:FB:2C:10:C0:FF:0B:CF:0D:32:86:FC:1A:A2
```

pkg/bundle/sync.go
```diff
@@ -165,24 +174,61 @@ func encodeJKS(trustBundle string) ([]byte, error) {
...
 	return buf.Bytes(), nil
 }
 
+// jksAlias creates a JKS-safe alias for the given certificate, such that
+// any two certificates will have a different aliases unless they're
+// identical in every way.
+// This unique alias fixes an issue where we used the Issuer field as an
+// alias, leading to different certs being treated as identical.
+func jksAlias(cert *x509.Certificate) string {
+	certHashBytes := sha256.Sum256(cert.Raw)
```
inteon, line 221: Can we do a panic in case cert.Raw is nil/ empty? I assume `jksAlias` only works for `*x509.Certificate` values that are returned by `x509.ParseCertificate`?

pkg/bundle/sync.go
```diff
@@ -152,9 +154,16 @@ func (b *bundle) secretBundle(ctx context.Context, ref *trustapi.SourceObjectKey
 	return string(data), nil
 }
 
-func encodeJKS(trustBundle string) ([]byte, error) {
+// encodeJKS creates a binary JKS file from the given PEM-encoded trust bundle and password.
+// Note that the password is not treated securely; JKS files generally seem to expect a password
+// to exist and so we have the option for one.
+func encodeJKS(trustBundle string, password []byte) ([]byte, error) {
```
JoshVanL, line 160: Don't know much about JKS but a cursory look seems as though the passwords should be string.

```suggestion
func encodeJKS(trustBundle string, password string) ([]byte, error) {
```

pkg/bundle/sync.go
JoshVanL, line 221: md5 is probably fine here.

pkg/bundle/sync.go
```diff
@@ -165,24 +174,61 @@ func encodeJKS(trustBundle string) ([]byte, error) {
...
+func jksAlias(cert *x509.Certificate) string {
+	certHashBytes := sha256.Sum256(cert.Raw)
+	certHash := hex.EncodeToString(certHashBytes[:])
+
+	// Since certHash is the part which actually distinguishes between two
+	// certificates, put it first so that it won't be truncated if a cert
+	// with a really long subject is added. Not sure what the upper limit
+	// for length actually is, but it shouldn't matter here.
+
+	return certHash[:8] + "|" + cert.Subject.String()
```
JoshVanL, line 229: Why do we need to append the subject string?

test/env/data.go
```diff
@@ -288,3 +290,32 @@ func EventuallyBundleHasSyncedAllNamespacesStartsWith(ctx context.Context, cl cl
...
 	).Should(BeNil(), fmt.Sprintf("checking bundle %s has synced to all namespaces with correct starting data", bundleName))
 }
+
+// CheckJKSFileSynced ensures that the given JKS data
+func CheckJKSFileSynced(jksData []byte, expectedPassword string, expectedCertPEMData string) error {
+	reader := bytes.NewReader(jksData)
+
+	ks := jks.New()
+
+	err := ks.Load(reader, []byte(expectedPassword))
```
JoshVanL, line 300: nit
```suggestion
	ks := jks.New()

	err := ks.Load(bytes.NewReader(jksData), []byte(expectedPassword))
```

pkg/bundle/sync.go
inteon: @SgtCoDFish

JoshVanL: Please don't panic in kube controllers. Bubble up the error that this is an irreconcilable state, and update the crd object with this state + condition.

SgtCoDFish: Makes sense to be defensive like that yeah :+1:

pkg/bundle/sync.go
SgtCoDFish: The API we're calling expects `[]byte`: https://pkg.go.dev/github.com/pavlo-v-chernykh/keystore-go/v4#KeyStore.Store

That was the main reason I went with `[]byte`, but it's also easier to zero out a byte slice than it is to reason about clearing the contents of a `string`. That's not massively relevant here though given we're not treating these as meaningful security boundaries.

pkg/bundle/sync.go
SgtCoDFish: It's fine from a security standpoint sure, but sha256 cert fingerprints aren't uncommon in the industry and it's easier to stick to hash algorithms which are still used and considered secure for cryptographic purposes. I don't think we're risking any performance issues by using sha256 over md5 here, either.

SgtCoDFish: Also note from my [self review](https://github.com/cert-manager/trust-manager/pull/129/files#r1184949988) that keytool shows the sha256 fingerprint.

pkg/bundle/sync.go
inteon: this is still a private function, so it probably is not important,

pkg/bundle/sync.go
inteon: so the `debian:verisign_universal_root_certification_authority.pem` values are actually the names.
and for trust manager it would look like `ffsdffds|Subject Of Certificate`

SgtCoDFish: The signature doesn't work here. A certificate comprises three fields as specified in [RFC 5280](https://www.rfc-editor.org/rfc/rfc5280#section-4.1.1) and the signature is just one of them.

If you just use the signatureValue, you're missing `tbsCertificate` and `signatureAlgorithm`.

Two certificates which have identical `signatureValue` but different `tbsCertificate` or `signatureAlgorithm` would be rare - and one of them would almost certainly be invalid. But we should still reflect that they're different, hence the use of the fingerprint here (i.e. a hash of the whole DER encoded certificate).

inteon (approved):
Thank you for resolving the remaining issues & feedback.
/approve
/lgtm

hazmat345:
Just tried out [v0.5.0-beta.1](https://github.com/cert-manager/trust-manager/releases/tag/v0.5.0-beta.1) and everything looks good to me. Thanks!

PR #130, opened 2023-05-03
Use boilersuite for boilerplate verification

SgtCoDFish:
Removes the need for Python in development environments for trust-manager!

This also requires some small boilerplate changes since boilersuite is a little more exacting than the old script was.


inteon:
/approve
/lgtm
Thanks for this improvement!

PR #134, opened 2023-05-09
Bump version to v0.5.0-beta.1

SgtCoDFish:
Note that the prerelease annotation wasn't committed for v0.5.0-beta.0 (in #125 ) and I added it manually before I generated the helm chart. I've added it here, so we just need to remember to update that annotation when we release the next non-prerelease version!


irbekrm (approved):
/lgtm

PR #137, opened 2023-05-22
Add description for JKS field for better docs

SgtCoDFish:
This will improve [this page](https://cert-manager.io/docs/projects/trust-manager/api-reference/) (after a followup PR for the website!)

I now realise (of course, just after the official release :grin: ) that JKS shouldn't be a `KeySelector`, it should be a `JKSFormat` or something. That'll have to be fixed in `v1alpha2` or `v1beta1`!


inteon (approved):
LGTM
I don't know whether KeySelector should be used or not.

PR #138, opened 2023-06-01
Bump dependencies. Changes to get latest c/r library working.

irbekrm:
This PR:
- updates go module dependencies
- bumps Go 1.19 -> 1.20
- makes changes required to use latest c/r library:
  - updates controller watch setup
  - changes to use the new webhook setup mechanism as the one we currently use has been removed
  - changes how the fake cache gets loaded with objects
  - ⚠️ BREAKING ⚠️ changes validating webhook path `/validate` -> `/validate-trust-cert-manager-io-v1alpha1-bundle` as this is [hardcoded in the new c/r webhook setup](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.15.0-alpha.0/pkg/builder/webhook.go#L247-L250). This has a potential of causing webhook downtime during upgrade to the version of trust-manager that adds this change (if the `ValidatingWebhookConfiguration` has been updated, but the webhook is still running the old version that does not yet serve to the new path)

Look at https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.15.0 to understand the changes

This PR is also very similar to https://github.com/cert-manager/approver-policy/pull/233


pkg/bundle/controller.go
```diff
@@ -95,28 +93,52 @@ func AddBundleController(ctx context.Context, mgr manager.Manager, opts Options)
...
-		}, builder.OnlyMetadata).
-
-		////// Sources //////
+		// Reconcile a Bundle on events against a ConfigMap that it
+		// owns. Only cache ConfigMap metadata.
+		WatchesMetadata(&corev1.ConfigMap{}, handler.EnqueueRequestForOwner(mgr.GetScheme(), mgr.GetRESTMapper(), &trustapi.Bundle{}, handler.OnlyControllerOwner())).
 
 		// Reconcile trust.cert-manager.io Bundles
-		Watches(source.NewKindWithCache(new(trustapi.Bundle), sourceCache), &handler.EnqueueRequestForObject{}).
+		WatchesRawSource(&source.Informer{Informer: bundleInformer}, &handler.EnqueueRequestForObject{}).
```
irbekrm, line 136: This is the new set up of how to use external cache

pkg/webhook/validation_test.go
```diff
@@ -20,11 +20,10 @@ import (
 	"context"
 	"testing"
 
-	admissionv1 "k8s.io/api/admission/v1"
```
irbekrm, line 23: The test change is purely in response to the new webhook setup mechanism. A couple test cases that were dealing with populating admission response/retrieving object from admission request have been removed as we no longer need to do that- c/r gives our webhook validator the decoded object from the request and only expects back an error and warnings.

deploy/charts/trust-manager/templates/webhook.yaml
```diff
@@ -49,4 +49,4 @@ webhooks:
       service:
         name: {{ include "trust-manager.name" . }}
         namespace: {{ .Release.Namespace | quote }}
-        path: /validate
+        path: /validate-trust-cert-manager-io-v1alpha1-bundle
```
irbekrm, line 52: See https://github.com/cert-manager/approver-policy/pull/233#discussion_r1194138573

pkg/bundle/controller.go
```diff
@@ -172,8 +186,8 @@ func AddBundleController(ctx context.Context, mgr manager.Manager, opts Options)
 
 		// Watch Secrets in trust Namespace. Only cache metadata.
 		// Reconcile Bundles who reference a modified source Secret.
-		Watches(source.NewKindWithCache(new(corev1.Secret), sourceCache), handler.EnqueueRequestsFromMapFunc(
-			func(obj client.Object) []reconcile.Request {
+		WatchesRawSource(&source.Informer{Informer: secretInformer}, handler.EnqueueRequestsFromMapFunc(
+			func(_ context.Context, obj client.Object) []reconcile.Request {
```
SgtCoDFish, line 190: question: here we ignore the context that's passed to this function in favour of the one that's in scope (see e.g. `bundleList := b.mustBundleList(ctx)`).

Why not use this context instead? It seems possible that the controller-runtime logic which calls this function could modify the context in some way.

```suggestion
			func(ctx context.Context, obj client.Object) []reconcile.Request {
```

SgtCoDFish (commented):
Couple of questions / suggestions on this - thanks for doing it!

pkg/bundle/controller.go
irbekrm: Thanks, didn't notice this one was actually using context!

irbekrm:
Thanks for taking a look and spotting the issues @SgtCoDFish ! Think I've now fixed those

pkg/bundle/controller.go
```diff
@@ -103,28 +93,52 @@ func AddBundleController(ctx context.Context, mgr manager.Manager, opts Options)
...
 		// Reconcile trust.cert-manager.io Bundles
-		Watches(source.NewKindWithCache(new(trustapi.Bundle), sourceCache), &handler.EnqueueRequestForObject{}).
+		WatchesRawSource(&source.Informer{Informer: bundleInformer}, &handler.EnqueueRequestForObject{}).
 
 		// Watch all Namespaces. Cache whole Namespaces to include Phase Status.
 		// Reconcile all Bundles on a Namespace change.
-		Watches(source.NewKindWithCache(new(corev1.Namespace), sourceCache), handler.EnqueueRequestsFromMapFunc(
-			func(obj client.Object) []reconcile.Request {
+		WatchesRawSource(&source.Informer{Informer: namespaceInformer}, handler.EnqueueRequestsFromMapFunc(
+			func(_ context.Context, obj client.Object) []reconcile.Request {
```
SgtCoDFish, line 141: suggestion: This function also uses context by the looks of it!

```suggestion
			func(ctx context.Context, obj client.Object) []reconcile.Request {
```

pkg/bundle/controller.go
```diff
@@ -142,8 +156,8 @@ func AddBundleController(ctx context.Context, mgr manager.Manager, opts Options)
 
 		// Watch ConfigMaps in trust Namespace. Only cache metadata.
 		// Reconcile Bundles who reference a modified source ConfigMap.
-		Watches(source.NewKindWithCache(new(corev1.ConfigMap), sourceCache), handler.EnqueueRequestsFromMapFunc(
-			func(obj client.Object) []reconcile.Request {
+		WatchesRawSource(&source.Informer{Informer: configMapInformer}, handler.EnqueueRequestsFromMapFunc(
+			func(_ context.Context, obj client.Object) []reconcile.Request {
```
SgtCoDFish, line 160: suggestion: same here!

```suggestion
			func(ctx context.Context, obj client.Object) []reconcile.Request {
```

SgtCoDFish (changes requested):

pkg/bundle/controller.go
irbekrm: Thanks for spotting those!

irbekrm:
weird docker flake

/retest

SgtCoDFish (approved):
/lgtm
/approve

This looks great and this is a piece of work that would've sucked to do so I really appreciate you doing it! :rocket:

PR #140, opened 2023-06-01
Add deprecation notes for the old cert-manager-trust name

SgtCoDFish:
The intent here is to encourage users to use 'trust-manager' instead.

This was discussed in the cert-manager biweekly meeting on 2023-06-01.

### DIFF
--- a/deploy/charts/trust/templates/NOTES.txt
+++ b/deploy/charts/trust/templates/NOTES.txt
@@ -1,0 +1,18 @@
+**************
+* DEPRECATED *
+**************
+
+'cert-manager-trust' has been renamed to 'trust-manager'.
+
+No further releases will be made under this old, deprecated name.
+
+Use 'trust-manager' instead and get all kinds of useful features including:
+
+- Support for publicly trusted certificate bundles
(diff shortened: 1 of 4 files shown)


SgtCoDFish:

There's no need to make the tests work here!

deploy/charts/trust/templates/NOTES.txt
```diff
@@ -0,0 +1,18 @@
...
+'cert-manager-trust' has been renamed to 'trust-manager'.
+
+No further releases will be made under this old, deprecated name.
+
+Use 'trust-manager' instead and get all kinds of useful features including:
+
+- Support for publicly trusted certificate bundles
+- JKS support
+- Much improved validation and error checking
+- And much more!
```
irbekrm, line 14: Nice :)

irbekrm (approved):
Makes sense, thanks! I guess we'll have to try out how it looks once the chart has been released, I've just read through now.

/lgtm

SgtCoDFish:
/override pull-trust-manager-smoke

(as above, no point in running tests here)

PR #141, opened 2023-06-01
Manually change chart packaging and add clarifications

SgtCoDFish:
The previous version would always generate a chart with appVersion == version, which we explicitly don't want here. Since we should ideally never change this again, it seems fine to hardcode and add a comment to warn!


SgtCoDFish:
Manually merging this because it's so trivial and is simply for visibility. We can change later, but this is just to warn people not to install the chart and doesn't impact how we actually install anything.

PR #143, opened 2023-06-09
Setting useDefaultCAs: false no longer causes failures

### LINKED ISSUE #133: BUG:  unsuppotted `useDefaultCAs: false`
If  trying to set `source.useDefaultCAs` to `false` the admission webhook prevents creation of the bundle.

How to reproduce:
```sh
cat  <<EOF | kubectl apply -f -
kind: Bundle
metadata:
  name: my-org.com-ca
spec:
  sources:
  - useDefaultCAs: false

  #- secret:
  #    name: my-org.com-ca
  #    key: "tls.crt"

  target:
    configMap:
      key: "certs.pem"
    namespaceSelector:
      matchLabels: {}
EOF
```

Error output:
`admission webhook "trust.cert-manager.io" denied the request: spec.sources.[0]: Forbidden: must define exactly one source type for each item but found 0 defined types`

Workaround is trivial comment the `useDefaultCAs` entry, but still one would expect the `false` is supported value.

hazmat345:
This fixes #133 - an issue where setting `useDefaultCAs: false` would cause validation (and also sync) errors.

In order to support this the validation code needed to change slightly. It now counts the number of sources in a bundle and will fail if that count is still zero after inspection. Setting `useDefaultCAs: true` is considered a source for this count, but `useDefaultCAs: false` is not.

So these specs are valid:
```yaml
spec:
  sources:
  - useDefaultCAs: true
```
```yaml
spec:
  sources:
  - secret:
      name: "our-org-ca"
      key: "ca.crt"
```
```yaml
spec:
  sources:
  - useDefaultCAs: true
  - secret:
      name: "our-org-ca"
      key: "ca.crt"
```
```yaml
spec:
  sources:
  - useDefaultCAs: false
  - secret:
      name: "our-org-ca"
      key: "ca.crt"
```

But this one is not:
```yaml
spec:
  sources:
  - useDefaultCAs: false

### DIFF
--- a/pkg/bundle/sync.go
+++ b/pkg/bundle/sync.go
@@ -105,10 +105,7 @@ func (b *bundle) buildSourceBundle(ctx context.Context, bundle *trustapi.Bundle)
 		bundles = append(bundles, string(sanitizedBundle))
 	}
 
-	// NB: bundles should never be empty here, since ValidateAndSanitizePEMBundle errors when a bundle source
-	// contains no valid certificates. Plus, the webhook validation should confirm that there's at least one source
-	// defined to avoid otherwise empty bundles.
-	// Still, just in case, we check and return an error in case somehow an empty bundle snuck through.
+	// NB: empty bundles are not valid so check and return an error if one somehow snuck through.
 
 	if len(bundles) == 0 {
 		return bundleData{}, fmt.Errorf("couldn't find any valid certificates in bundle")
--- a/pkg/bundle/sync.go
+++ b/pkg/bundle/sync.go
@@ -80,7 +80,11 @@ func (b *bundle) buildSourceBundle(ctx context.Context, bundle *trustapi.Bundle)
 		case source.InLine != nil:
 			sourceData = *source.InLine
 
-		case source.UseDefaultCAs != nil && *source.UseDefaultCAs:
+		case source.UseDefaultCAs != nil:
+			if *source.UseDefaultCAs == false {
+				continue
+			}
+
 			if b.defaultPackage == nil {
 				err = notFoundError{fmt.Errorf("no default package was specified when trust-manager was started; default CAs not available")}
 			} else {
--- a/pkg/webhook/validation.go
+++ b/pkg/webhook/validation.go
@@ -65,67 +65,71 @@ func (v *validator) validate(ctx context.Context, obj runtime.Object) (admission
 		warnings admission.Warnings
 		path     = field.NewPath("spec")
 	)
-	if len(bundle.Spec.Sources) == 0 {
-		el = append(el, field.Forbidden(path.Child("sources"), "must define at least one source"))
-	} else {
-		path := path.Child("sources")
 
-		defaultCAsCount := 0
+	sourceCount := 0
+	defaultCAsCount := 0
 
-		for i, source := range bundle.Spec.Sources {
-			path := path.Child("[" + strconv.Itoa(i) + "]")
+	for i, source := range bundle.Spec.Sources {
+		path := path.Child("sources").Child("[" + strconv.Itoa(i) + "]")
 
-			unionCount := 0
+		unionCount := 0
 
-			if configMap := source.ConfigMap; configMap != nil {
-				path := path.Child("configMap")
-				unionCount++
+		if configMap := source.ConfigMap; configMap != nil {
+			path := path.Child("configMap")
+			sourceCount++
+			unionCount++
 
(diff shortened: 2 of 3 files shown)


Jamstah:
This will leave us open to a user creating a bundle with no certs: https://github.com/cert-manager/trust-manager/blob/main/pkg/bundle/sync.go#L104

Personally, I feel like producing an empty bundle shouldn't be an error and we could just remove the empty bundle checks, but at a minimum we should update that comment because it would no longer be an unreachable path.

hazmat345:
@Jamstah I updated the comment but you bring up a good point about how empty bundles are handled. With this PR a bundle with `useDefaultCAs: false` and no other sources will pass validation but error when syncing, which isn't great. How should this be fixed?

Probably the least intrusive option is to just make that scenario fail validation. I tend to agree that an empty bundle isn't really an *error*, but I suspect allowing empty bundles has the potential to cause a lot of follow-on work - it would be reversing a pretty low-level assumption so I'm guessing lots of different areas might be affected. That kind of feels like it should be a dedicated PR.

Thoughts?

inteon:
https://github.com/cert-manager/trust-manager/issues/133#issuecomment-1628427746
The following item is just a no-op item, for this reason it is not accepted by the webhook:
`- useDefaultCAs: false`
Only `- useDefaultCAs: true` currently makes sense.

The alternative would have been to not have a boolean and let you specify `- useDefaultCAs: {}` instead of `- useDefaultCAs: true`, which seems harder to read.
What is your use case for having the `- useDefaultCAs: false` option? Can you remove the `- useDefaultCAs: false` item from your sources list instead?

Jamstah:
> @Jamstah I updated the comment but you bring up a good point about how empty bundles are handled. With this PR a bundle with `useDefaultCAs: false` and no other sources will pass validation but error when syncing, which isn't great. How should this be fixed?
> 
> Probably the least intrusive option is to just make that scenario fail validation. I tend to agree that an empty bundle isn't really an _error_, but I suspect allowing empty bundles has the potential to cause a lot of follow-on work - it would be reversing a pretty low-level assumption so I'm guessing lots of different areas might be affected. That kind of feels like it should be a dedicated PR.
> 
> Thoughts?

Personally, I think we should change the assumption, its a v0 release. My reasoning is that a popular kubernetes pattern is label selectors, and selecting certs to include by label shouldn't fail depending on the number of certs with that label that exist (full disclosure, I'm suggesting a similar patten here: #144)

On the negative side, if an empty trust store is created and used by an application, that application could fail. The existence of an empty trust store could allow a pod to be created that then won't respond to a change in the trust store, which is also bad. These are technically application bugs though.

The right thing to do is probably not change it here though, but to raise an issue for discussion.

Jamstah:
> [#133 (comment)](https://github.com/cert-manager/trust-manager/issues/133#issuecomment-1628427746) The following item is just a no-op item, for this reason it is not accepted by the webhook: `- useDefaultCAs: false` Only `- useDefaultCAs: true` currently makes sense.
> 
> The alternative would have been to not have a boolean and let you specify `- useDefaultCAs: {}` instead of `- useDefaultCAs: true`, which seems harder to read. What is your use case for having the `- useDefaultCAs: false` option? Can you remove the `- useDefaultCAs: false` item from your sources list instead?

For me, not having it work feels like a bug.

If I have a helm chart, for example, and want someone to be able to pass on a value to pass to the cr, the helm chart needs a conditional instead of being able to pass a Boolean, which feels like a workaround.

I agree `useDefaultCAs: {}` would be a better representation, and then require that conditional, but then it would be for a good reason.

hazmat345:
@inteon 

> The alternative would have been to not have a boolean and let you specify - useDefaultCAs: {} instead of - useDefaultCAs: true, which seems harder to read.

Definitely agree - boolean seems like the correct choice here.

> What is your use case for having the - useDefaultCAs: false option? Can you remove the - useDefaultCAs: false item from your sources list instead?

I'm running trust-manager in two environments - one where I want to use the default CAs and one where I don't. I want to be able to explicitly set this value in both bundle definitions.

@Jamstah 
> For me, not having it work feels like a bug.

Agree - `false` is a valid boolean value. A valid value causing errors feels like a bug to me.

hazmat345:
@inteon I updated the source validation based on your suggestion. If you see anything else that needs to be changed please just let me know 😄

inteon:
Thank you @hazmat345, this fixes the issue & this PR already has tests for this change too.
Great work!
/approve
/lgtm

Issue #69, opened 2022-11-10
Unable to run controller on RKE


xrl:
We are on RKE with Kubernetes 1.21. When I install the helm chart I see the replicaset is stuck with 0 copies and this attached event:

```
% % k -n cert-manager describe rs cert-manager-trust-6796dcc8f5
Events:
  Type     Reason        Age                      From                   Message
  ----     ------        ----                     ----                   -------
  Warning  FailedCreate  4m32s (x548 over 5d20h)  replicaset-controller  Error creating: pods "cert-manager-trust-6796dcc8f5-" is forbidden: PodSecurityPolicy: unable to admit pod: [pod.metadata.annotations[container.seccomp.security.alpha.kubernetes.io/cert-manager-trust]: Forbidden: seccomp may not be set pod.metadata.annotations[container.seccomp.security.alpha.kubernetes.io/cert-manager-trust]: Forbidden: seccomp may not be set]
```

and it does not appear that seccomp is a configurable field on a PSP. So I'm not sure how to allow the pod to define seccomp. Would the project be open to a flag for optionally hiding the securityContext option?

SgtCoDFish:
Hey, sorry this took a while to get back to you on!

Thanks for raising this and submitting a PR, you're a star! Seems reasonable to add this, I'll take a look at #70 :+1:

xrl:
How do we feel about cutting a release?

SgtCoDFish:
I think we'll cut a release after #46 gets merged because that's a pretty big new usability improvement. If it looks like that might not be merged any time soon, I'll cut 0.3.1 to release this.

Definitely not until at least next week though!

xrl:
Perfect, we'll keep using our local copy. No problem.

PR #151, opened 2023-07-26
Respect seccomp toggle in init container

xrl:
I worked on #69 in order to disable seccomp in the deployment but it looks like an initcontainer was added. I cannot deploy the bog-standard trust-manager until this too is disabled.


inteon:
@xrl could you signoff your commit?
https://cert-manager.io/docs/contributing/sign-off/

inteon:
duplicate PR (see https://github.com/cert-manager/trust-manager/pull/118)

Issue #132, opened 2023-05-08
Unable to run Trust Manager without cert manager


ishantagarwal95:
I want to install Trust Manager without cert manager. For that I removed the certificate.yaml file from helm chart. I also saved _ca.crt_, _tls.crt_, _tls.key_ files in trust-manager-tls secret. This corresponds to the certificate I got signed from AWS Private CA.
I then tried to install trust manager via helm chart. Trust Manager got installed successfully.
Now I applied a Bundle to sync the ca.crt file to example config maps. This gave me the following error:
```
Internal error occurred: failed calling webhook "trust.cert-manager.io": failed to call webhook: Post "https://trust-manager.cert-manager.svc:443/validate?timeout=5s": x509: certificate signed by unknown authority
```

Can someone please help me here. I just want to use a certificate issued by AWS Private CA and sync is CA certificate to other configmaps. For this, I don't want to use cert manager.

erikgb:
I think trust-manager (at least) depends on cert-manager to provide a certificate for the validating webhook in trust-manager. Obtaining webhook certificates is always a painful process, and there are at least a dozen different approaches to this in the Kubernetes operator landscape. As trust-manger is maintained by the cert-manager team, I think it seem legit to use cert-manager there. Why do you not want to install cert-manager?

SgtCoDFish:
Hey, thanks for raising this!  I'm gonna close since #157 implements this - keep an eye out for the next trust-manager release which should enable this 👍

PR #157, opened 2023-08-09
Add support for generating certificates with helm

SgtCoDFish:
This removes the hard dependency on cert-manager by allowing users to choose not to create a Certificate.

The extremely long lifetime of the cert is to remove the chance of it expiring in the lifetime of a typical cluster. Users who want rotation would be well advised to use cert-manager, which remains the default

### DIFF
--- a/deploy/charts/trust-manager/templates/webhook.yaml
+++ b/deploy/charts/trust-manager/templates/webhook.yaml
@@ -1,3 +1,52 @@
+{{- if and .Values.app.webhook.tls.helmCert.enabled .Values.app.webhook.tls.approverPolicy.enabled -}}
+{{- fail "cannot set .app.webhook.tls.helmCert.enabled and .Values.app.webhook.tls.approverPolicy.enabled" }}
+{{- end -}}
+
+{{- /*
+$ca is always generated here even if .Values.app.webhook.tls.helmCert.enabled is false because we need it to be
+visible in the scope for the ValidatingWebhookConfiguration below.
+
+genCA has two args - first is cert commonName (CN) and second is validity in days (9125 = ~25 years)
+
+We don't write this CA's private key to a secret because we don't want it to be used for any other purpose
+except for generating $cert below.
+
+DO NOT USE $ca ANYWHERE IF app.tls.helmCert.enabled IS FALSE
+*/}}
+
+{{- $ca := genCA (printf "*.%s.svc" .Release.Namespace ) 9125 -}}
+
+
+{{- if .Values.app.webhook.tls.helmCert.enabled -}}
+{{- $svcName := (printf "%s.%s.svc" (include "trust-manager.name" . ) .Release.Namespace ) -}}
+
+{{- /*
+genSignedCert has the following args, in order:
+1. The cert CN
+2. A list of IP addresses the cert is valid for
+3. A list of DNS altnames the cert is valid for
+4. The duration in days (3650 = ~10 years)
+5. The signing CA
+*/}}
+{{- $cert := genSignedCert $svcName nil (list $svcName) 3650 $ca -}}
+
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ include "trust-manager.name" . }}-tls
+  namespace: {{ .Release.Namespace }}
+  labels:
+{{ include "trust-manager.labels" . | indent 4 }}
+type: kubernetes.io/tls
+data:
+  ca.crt: {{ $ca.Cert | b64enc }}
+  tls.key: {{ $cert.Key | b64enc }}
+  tls.crt: {{ $cert.Cert | b64enc }}
+
+---
+
+{{ end }}
+
 apiVersion: v1
 kind: Service
 metadata:
@@ -24,7 +73,9 @@ spec:
       name: webhook
   selector:
     app: {{ include "trust-manager.name" . }}
+
 ---
+
 apiVersion: admissionregistration.k8s.io/v1
 kind: ValidatingWebhookConfiguration
 metadata:
@@ -32,8 +83,10 @@ metadata:
   labels:
     app: {{ include "trust-manager.name" . }}
     {{- include "trust-manager.labels" . | nindent 4 }}
+{{ if not .Values.app.webhook.tls.helmCert.enabled }}
   annotations:
     cert-manager.io/inject-ca-from: "{{ include "trust-manager.namespace" . }}/{{ include "trust-manager.name" . }}"
+{{ end }}
 
 webhooks:
   - name: trust.cert-manager.io
@@ -52,6 +105,9 @@ webhooks:
     failurePolicy: Fail
     sideEffects: None
     clientConfig:
+{{ if .Values.app.webhook.tls.helmCert.enabled }}
+      caBundle: "{{ $ca.Cert | b64enc }}"
+{{ end }}
       service:
         name: {{ include "trust-manager.name" . }}
         namespace: {{ include "trust-manager.namespace" . }}
--- a/deploy/charts/trust-manager/values.yaml
+++ b/deploy/charts/trust-manager/values.yaml
@@ -227,6 +227,12 @@ app:
       # nodePort: 8080
 
     tls:
+      helmCert:
+        # Whether to issue a webhook cert using Helm, which removes the need to install cert-manager.
+        # Helm-issued certificates can be challenging to rotate and maintain, and the issued cert will have a duration of 10 years and be modified when trust-manager is updated.
+        # It's safer and easier to rely on cert-manager for issuing the webhook cert - avoid using Helm-generated certs in production.
+        enabled: false
+
       approverPolicy:
         # Whether to create an approver-policy CertificateRequestPolicy allowing auto-approval of the trust-manager webhook certificate. If you have approver-policy installed, you almost certainly want to enable this.
         enabled: false
--- a/deploy/charts/trust-manager/README.md
+++ b/deploy/charts/trust-manager/README.md
@@ -401,6 +401,13 @@ Sets the families that should be supported and the order in which they should be
 
 The nodePort set on the Service used by the webhook.
 
+#### **app.webhook.tls.helmCert.enabled** ~ `bool`
+> Default value:
+> ```yaml
+> false
+> ```
+
+Whether to issue a webhook cert using Helm, which removes the need to install cert-manager. Helm-issued certificates can be challenging to rotate and maintain, and the issued cert will have a duration of 10 years and be modified when trust-manager is updated. It's safer and easier to rely on cert-manager for issuing the webhook cert - avoid using Helm-generated certs in production.
 #### **app.webhook.tls.approverPolicy.enabled** ~ `bool`
 > Default value:
 > ```yaml
--- a/deploy/charts/trust-manager/templates/certificate.yaml
+++ b/deploy/charts/trust-manager/templates/certificate.yaml
@@ -1,3 +1,5 @@
+{{- if not .Values.app.webhook.tls.helmCert.enabled -}}
+
 apiVersion: cert-manager.io/v1
 kind: Issuer
 metadata:
@@ -82,5 +84,7 @@ subjects:
 - kind: ServiceAccount
   name: {{ .Values.app.webhook.tls.approverPolicy.certManagerServiceAccount }}
   namespace: {{ .Values.app.webhook.tls.approverPolicy.certManagerNamespace }}
+
+{{ end }}
 
 {{ end }}
--- a/deploy/charts/trust-manager/templates/deployment.yaml
+++ b/deploy/charts/trust-manager/templates/deployment.yaml
@@ -12,9 +12,15 @@ spec:
       app: {{ include "trust-manager.name" . }}
   template:
     metadata:
-      {{- with .Values.app.podAnnotations }}
+      {{- if or .Values.app.podAnnotations .Values.app.webhook.tls.helmCert.enabled }}
       annotations:
-        {{- toYaml . | nindent 8 }}
+        {{- if .Values.app.podAnnotations }}
+        {{- toYaml .Values.app.podAnnotations | nindent 8 }}
+        {{- end }}
+        {{- if .Values.app.webhook.tls.helmCert.enabled }}
+        {{- /* When using a helm cert, the cert will be regenerated every time the chart is updated. When that happens, we need to restart the pods in the deployment to ensure the new cert is picked up. */}}
+        rollme-due-to-helm-cert: {{ randAlphaNum 5 | quote }}
+        {{- end }}
       {{- end }}
       labels:
         app: {{ include "trust-manager.name" . }}
--- a/deploy/charts/trust-manager/values.schema.json
+++ b/deploy/charts/trust-manager/values.schema.json
@@ -366,6 +366,9 @@
       "properties": {
         "approverPolicy": {
           "$ref": "#/$defs/helm-values.app.webhook.tls.approverPolicy"
+        },
+        "helmCert": {
+          "$ref": "#/$defs/helm-values.app.webhook.tls.helmCert"
         }
       },
       "type": "object"
@@ -398,6 +401,20 @@
     "helm-values.app.webhook.tls.approverPolicy.enabled": {
       "default": false,
       "description": "Whether to create an approver-policy CertificateRequestPolicy allowing auto-approval of the trust-manager webhook certificate. If you have approver-policy installed, you almost certainly want to enable this.",
+      "type": "boolean"
+    },
+    "helm-values.app.webhook.tls.helmCert": {
+      "additionalProperties": false,
+      "properties": {
+        "enabled": {
+          "$ref": "#/$defs/helm-values.app.webhook.tls.helmCert.enabled"
+        }
+      },
+      "type": "object"
+    },
+    "helm-values.app.webhook.tls.helmCert.enabled": {
+      "default": false,
+      "description": "Whether to issue a webhook cert using Helm, which removes the need to install cert-manager. Helm-issued certificates can be challenging to rotate and maintain, and the issued cert will have a duration of 10 years and be modified when trust-manager is updated. It's safer and easier to rely on cert-manager for issuing the webhook cert - avoid using Helm-generated certs in production.",
       "type": "boolean"
     },
     "helm-values.commonLabels": {


hawksight:
@SgtCoDFish I had a quick go from your branch with:

```shell
kind create cluster --name trust
gh repo clone SgtCoDFish/cert-manager-trust-manager
cd cert-manager-trust-manager
gco gco isolatedcert
k create ns cert-manager
helm install -n cert-manager trust-manager ./deploy/charts/trust-manager --set issueIsolatedCert=true
# Get before secret
k get secret -n cert-manager trust-manager-tls -o yaml > install-cert.yaml
# check upgrade
helm diff upgrade -n cert-manager trust-manager ./deploy/charts/trust-manager/ --set issueIsolatedCert=true
# This also seems to give the same diff result
helm diff upgrade -n cert-manager trust-manager ./deploy/charts/trust-manager/ --reuse-values
# get after secret
k get secret -n cert-manager trust-manager-tls -o yaml > after-cert.yaml
```

When I go to upgrade it looks like it will generate a new CA / cert, from a `helm diff`:

```console
-       caBundle: "LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCk1JSURKakNDQWc2Z0F3SUJBZ0lSQVBDSllpWXpoWVFSOXNhYThiL29uSEF3RFFZSktvWklodmNOQVFFTEJRQXcKSFRFYk1Ca0dBMVVFQXd3U0tpNWpaWEowTFcxaGJtRm5aWEl1YzNaak1CNFhEVEl6TURneE1ERXdNemMxTkZvWApEVFE0TURnd016RXdNemMxTkZvd0hURWJNQmtHQTFVRUF3d1NLaTVqWlhKMExXMWhibUZuWlhJdWMzWmpNSUlCCklqQU5CZ2txaGtpRzl3MEJBUUVGQUFPQ0FROEFNSUlCQ2dLQ0FRRUFxMXI1ZFpGL1RRcEh5UW5Gb0U5cE1mcloKZU1jYUZJK1pLb2t2UEdaU3NtSHhSSEh3YlZKbDNOZlRreUllZkxyeU5vZHloeTI0S3AzN1lDRmpyWFFqRDkwTwpTQWJoRjNJU0dzUndoNFZWa3RPTS91ZjlhUU1JL0RWb3hjK0VmMEY4Q3B1Rkk1NFNzT3d2S2F1bHhDVkZNS3V2CnpQTnNCVkRmbWh4c2kvOGxybkZ4NHJEM3RsQVFaWHllcnRpdTFSNlAvdENpRGU2SzVFdWE4ZzlQdnFpdnJIVnYKMVB6REx0RDdkN2Z6MGpYYkVpd0xTenQwMGpIcWJjZTZXSEdDS0dETlUwQ2dZUEswMzExTHNhdGRMeWdqRFFkZgowdThnd1d1cWR3K2NtT21BVllVMmh6am82VW5sajduK1RoV2V6djRreHpKOUdpQnRsQTBNNGdLK3BTNEpRUUlECkFRQUJvMkV3WHpBT0JnTlZIUThCQWY4RUJBTUNBcVF3SFFZRFZSMGxCQll3RkFZSUt3WUJCUVVIQXdFR0NDc0cKQVFVRkJ3TUNNQThHQTFVZEV3RUIvd1FGTUFNQkFmOHdIUVlEVlIwT0JCWUVGRTc3SFdvQlIxWUpSL2VZTllkbApRZ3BrdHJPdE1BMEdDU3FHU0liM0RRRUJDd1VBQTRJQkFRQlhtekJuc2hDdW5oSkVWb1NUdkF2b1pJYU5peGI2Cjc0QTJQVzgzVzJoY1ZKbUlPTWFJdHk0VHA3Ymk5RVRQaGc4M1NtbFc4SlhkeTQxZWp3ZEU4MXA5azN3QlBEcEgKU1dRNG9iOXpnazJVUUhVdGVKQ09ISnR6REFNUkM3dDBlb1MvWlJLVTJRUi9Zd1EvL2wrL3RIck5BUjdPcGI5WAovOUZkdWNVRlZjY2h1UGxpb1pOc01mSkFKTU44SUJnbmIzUmFoeVJranlkdjBuaStpQTlGdEhJK1JjNWh2eWNGCi91UXByN2xyYi9rR2VKVmxmdFFZYys1VldCc1ZwVEEzckdFVGRoS1I1UWluVENXek1kb3NJYWp0Y1JRYW1mVk4KNGdyZDh5Z1dqdy9CTWdFMDM4Qm80U011TTFiaHp1TmF3eHlza2I1Snpoa3VrcUFiMTl1azhIZEkKLS0tLS1FTkQgQ0VSVElGSUNBVEUtLS0tLQo="
+       caBundle: "LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCk1JSURKakNDQWc2Z0F3SUJBZ0lSQVA1dWxUSGlPTEEwUjlJS2xNQWhlVmd3RFFZSktvWklodmNOQVFFTEJRQXcKSFRFYk1Ca0dBMVVFQXd3U0tpNWpaWEowTFcxaGJtRm5aWEl1YzNaak1CNFhEVEl6TURneE1ERXdNemt6TjFvWApEVFE0TURnd016RXdNemt6TjFvd0hURWJNQmtHQTFVRUF3d1NLaTVqWlhKMExXMWhibUZuWlhJdWMzWmpNSUlCCklqQU5CZ2txaGtpRzl3MEJBUUVGQUFPQ0FROEFNSUlCQ2dLQ0FRRUE1alJVbGJOVVNJd3NHdm1uNW5rYzZZa0wKMTcyaW9VSW0zLzg0UzhpNStFNEpQeUFlK2tOaEh2ZGZqRll4RnA4dmcwQ0RJQ1QwMi92MlNwNENzT0plSlhXTApjYzM0Z2NxQTlSbW5yTW5UK2pSbEI1dTZXLzRTcTlZRHpxNVZJMUlrQ3RiYmFSY1MyTHprRXpHR1pkRWxNT2MvCnBvZ0M1b1NuQzFsOVdQVEttckt2Ty9xSW9RaEpoZ1hIemV6cmVrMmVEQUh6OFBSK2JQbm1OVXVld2YzUEZtNzIKSTE3QUtUYlljK2tSOHRZdmFubWR4QVN1UXBuM2JvOEpWMHZiMmxKSm9LTWxuRmtWdlFRMWpwWDRVVE8wYStOKwoxVUtrMStpOTVnN1VrTlRWNWJuSEpwUEgxU01Xd0xNT0xjdkJucnFSZWJqTjkwMU12bldlQTlLemZlWEdTUUlECkFRQUJvMkV3WHpBT0JnTlZIUThCQWY4RUJBTUNBcVF3SFFZRFZSMGxCQll3RkFZSUt3WUJCUVVIQXdFR0NDc0cKQVFVRkJ3TUNNQThHQTFVZEV3RUIvd1FGTUFNQkFmOHdIUVlEVlIwT0JCWUVGTDJqaE16S2tRNFRKckEzbm0vYQovVUcxSXN6M01BMEdDU3FHU0liM0RRRUJDd1VBQTRJQkFRREUzZTA2cHZ4OS9JTnpjSVhIZFdTTHFSa2xoM044CjZmM3J2SG83UFI0QTZDb1Q3cGozQXNBRzdLM3F5Tko1NVdUMUhMR0FkcUYrWnV6S0tJay9jU2V4WkZBRXN2OG0KRHVvQWduTlpudTltb0RHY1d6SmlRbHFKZlVlUXJYNmM3MVRuMDZoV1o4dTcydngvT2NVdGtKM3JUMStJVkV0bwpxd25MNlBKZHFhaTJHbHFpTjJ1R09Fc2NXb0VjLzZCVWJLRU11Z0NTbWFWOTk5TThzTSttcjJVU0dTaVRtd0w0CjMzcEdCbytkTEE4NnF1TTdhUEdFbGgrckFRUW54cG5VRUNwTVZTcVp1SlpYUzllUE5xRWYxRnh5SzN5SGtEVnYKdUU5dnNwRWx2bmZsS1NCZnpBcmYvanVvcVgzOUozQUhiUlkwTGFrNXdCdTRlc3cwMzJLN0JmTisKLS0tLS1FTkQgQ0VSVElGSUNBVEUtLS0tLQo="

        service:
          name: trust-manager
          namespace: "cert-manager"
          path: /validate-trust-cert-manager-io-v1alpha1-bundle
cert-manager, trust-manager-tls, Secret (v1) has changed:
  # Source: trust-manager/templates/webhook.yaml
  apiVersion: v1
  kind: Secret
  metadata:
    labels:
      app.kubernetes.io/instance: trust-manager
      app.kubernetes.io/managed-by: Helm
      app.kubernetes.io/name: trust-manager
      app.kubernetes.io/version: v0.5.0
      helm.sh/chart: trust-manager-v0.5.0
    name: trust-manager-tls
    namespace: cert-manager
  data:
-   ca.crt: '-------- # (1151 bytes)'
-   tls.crt: '-------- # (1229 bytes)'
-   tls.key: '-------- # (1679 bytes)'
+   ca.crt: '++++++++ # (1151 bytes)'
+   tls.crt: '++++++++ # (1229 bytes)'
+   tls.key: '++++++++ # (1675 bytes)'
  type: kubernetes.io/tls
```

Had a quick look at the `ca.crt` beore and after:

```shell
# before
Certificate:
    Data:
        Version: 3 (0x2)
        Serial Number:
            f0:89:62:26:33:85:84:11:f6:c6:9a:f1:bf:e8:9c:70
        Signature Algorithm: sha256WithRSAEncryption
        Issuer: CN = *.cert-manager.svc
        Validity
            Not Before: Aug 10 10:37:54 2023 GMT
            Not After : Aug  3 10:37:54 2048 GMT

# after
Certificate:
    Data:
        Version: 3 (0x2)
        Serial Number:
            51:be:bc:4e:a1:a2:33:d1:f9:f8:62:79:b9:79:9e:85
        Signature Algorithm: sha256WithRSAEncryption
        Issuer: CN = *.cert-manager.svc
        Validity
            Not Before: Aug 10 10:44:06 2023 GMT
            Not After : Aug  3 10:44:06 2048 GMT
```

I guess I have two questions:
1. Does it matter if we have a new version every time we upgrade, functionally?
2. Is there a way to not have it do this maybe, so on upgrade there is less visual change? (if the answer to 1 is yes)

inteon:
I think this is a easy way to omit the cert-manager & ca-injector dependencies.
I think it is great for development & testing.
For production this is probably also ok for a good amount of use cases.
I can see the following limitations currently (in what cases a user should not use this feature):
- no rotation -> no short lived tls certificates possible
- no non-selfsigned/ private pki certs
- private key is part of Helm state & static
- secret is dynamic (each Helm template call gives you another cert)
- is not supported in static yaml

Currently, other controllers often have a "create self-signed cert" option that allows them to create a secret that can be injected using ca-injector (so no need for cert-manager to be installed).
https://github.com/cert-manager/cert-manager/blob/master/pkg/webhook/authority/authority.go
https://github.com/cert-manager/approver-policy/blob/main/pkg/internal/webhook/tls/tls.go
https://github.com/cert-manager/webhook-lib/blob/main/server/tls/dynamic_source.go#L37
Maybe we should also try to create a more consistent story here?

SgtCoDFish:
> I guess I have two questions:

They're great questions 😁 

> Does it matter if we have a new version every time we upgrade, functionally?

It doesn't as long as we also update the caBundle on the ValidatingWebhookConfiguration, which it looks like we do. There's also a need to restart pods because of that behaviour, which I'll talk about below.

> Is there a way to not have it do this maybe, so on upgrade there is less visual change? (if the answer to 1 is yes)

I'm not sure! I think it'd be preferable to not reissue (it's not like the cert's gonna expire) but I can't really see how to do it. That might just be because my Helm-foo isn't good enough though!

## Restarting the Pod

Your super-helpful test script led me to realise there'd be an error here because the webhook would be updated but the pod would still be using the old cert. I'll post my modified test script below, then add an extra commit which fixes that problem.

```bash
#!/usr/bin/env bash

set -eu -o pipefail

kind delete clusters trust

kind create cluster --name trust

kubectl create ns cert-manager

helm install -n cert-manager trust-manager ./deploy/charts/trust-manager --set issueIsolatedCert=true --wait

# Get before secret
kubectl get secret -n cert-manager trust-manager-tls -o yaml > install-cert.yaml

kubectl apply -f - <<EOF
kind: Bundle
metadata:
  name: mybundle
spec:
  sources:
  - useDefaultCAs: true
  target:
    configMap:
      key: "target-key"
EOF

sleep 2

# check upgrade
helm upgrade -n cert-manager trust-manager ./deploy/charts/trust-manager/ --set issueIsolatedCert=true --wait

# get after secret
kubectl get secret -n cert-manager trust-manager-tls -o yaml > after-cert.yaml

kubectl apply -f - <<EOF
kind: Bundle
metadata:
  name: mybundle-2
spec:
  sources:
  - useDefaultCAs: true
  target:
    configMap:
      key: "target-key"
EOF
```

SgtCoDFish:
Oh, also - testing with a locally built chart will fail unless you've also built new containers. The v0.5.0 images aren't supported by the chart on master because of a change to the validation path, so I used this locally (temporarily, just for testing):

```diff
diff --git a/deploy/charts/trust-manager/templates/webhook.yaml b/deploy/charts/trust-manager/templates/webhook.yaml
index 382cb3e..0ca95ae 100644
--- a/deploy/charts/trust-manager/templates/webhook.yaml
+++ b/deploy/charts/trust-manager/templates/webhook.yaml
@@ -100,4 +100,5 @@ webhooks:
       service:
         name: {{ include "trust-manager.name" . }}
         namespace: {{ .Release.Namespace | quote }}
-        path: /validate-trust-cert-manager-io-v1alpha1-bundle
+        #path: /validate-trust-cert-manager-io-v1alpha1-bundle
+        path: /validate
```

erikgb:
I believe this PR will close https://github.com/cert-manager/trust-manager/issues/132 if merged.

deploy/charts/trust-manager/values.yaml
```diff
@@ -68,6 +68,9 @@ app:
       type: ClusterIP
 
     tls:
+      # -- Whether to issue an "isolated" cert, which removes the requirement for cert-manager to be installed by using Helm to issue a webhook certificate.
```
maelvls, line 71: It would be great to explain why this approach is fine as long as you upgrade trust-manager at least every once every 10 years. Am I understanding correctly?

maelvls: Great, the new comment answers all the of questions I had.

deploy/charts/trust-manager/templates/webhook.yaml
```diff
@@ -1,3 +1,47 @@
+{{- /*
+$ca is always generated here even if issueIsolatedCert is false because we need it to be
+visible in the scope for the ValidatingWebhookConfiguration below.
+
+genCA has two args - first is cert commonName and second is validity in days (9125 = ~25 years)
+
+We don't write this CA to a secret because we don't want it to be used for any other purpose!
```
erikgb, line 7: This statement seems untrue. The CA is written to a secret below, or am I misunderstanding something?

deploy/charts/trust-manager/values.yaml
```diff
@@ -227,6 +227,11 @@ app:
       # nodePort: 8080
 
     tls:
+      # Whether to issue an "isolated" cert, which removes the requirement for cert-manager to be installed by using Helm to issue a webhook certificate.
+      # Helm-issued certificates can be challenging to rotate and maintain. The issued cert will have a duration of 10 years.
+      # It's safer and easier to rely on cert-manager for issuing the webhook cert.
+      issueIsolatedCert: false
```
erikgb, line 233: The term "isolated cert" confuses me. How is it more isolated than cert-manager certificates? Naming is hard, but I would personally be happier with something in the direction of "staticCert".

erikgb (changes requested):
I didn't know Helm could do things like this! Thanks for the demo, @SgtCoDFish!

I am personally not a big fan of this method of making the cert-manager dependency "softer", and would rather see an opt-in to reuse the bootstrap mechanism used in cert-manager and approver-policy. First and foremost, I prefer to make things idempotent, and it seems like this Helm solution will create a new certificate on each render.

deploy/charts/trust-manager/templates/webhook.yaml
```diff
@@ -1,3 +1,47 @@
+{{- /*
+$ca is always generated here even if issueIsolatedCert is false because we need it to be
+visible in the scope for the ValidatingWebhookConfiguration below.
```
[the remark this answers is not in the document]
SgtCoDFish: I think the thing here is that $ca is a root cert which is always generated. $cert (below) is the actual cert which is only generated if requested

deploy/charts/trust-manager/templates/webhook.yaml
SgtCoDFish: I'll improve the comment - it should say that the ca's private key isn't written to a secret!

deploy/charts/trust-manager/values.yaml
SgtCoDFish: As we discussed in standup this morning, I'll rename this to `helmCert.enabled`!

SgtCoDFish:
I've updated the commit in a few ways:

1. Improved wording and descriptions, in line with Erik's review
2. Fixed an issue with how annotations were added to pods in the deployment. Tested to confirm it works
3. Changed the value to `helmCert.enabled` as discussed
4. Added a check for whether approver-policy and helmCert are both enabled. If they are, the chart fails to install (since they don't make sense together).

I tested installing and templating locally using the following values.yaml file:

```yaml
app:
  podAnnotations:
    abc: "123"
  webhook:
    tls:
      helmCert:
        enabled: true
image:
  tag: v0.11.1
```

deploy/charts/trust-manager/templates/webhook.yaml
```diff
@@ -1,3 +1,52 @@
+{{- if and .Values.app.webhook.tls.helmCert.enabled .Values.app.webhook.tls.approverPolicy.enabled -}}
+{{- fail "cannot set .app.webhook.tls.helmCert.enabled and .Values.app.webhook.tls.approverPolicy.enabled" }}
+{{- end -}}
+
+{{- /*
+$ca is always generated here even if .Values.app.webhook.tls.helmCert.enabled is false because we need it to be
+visible in the scope for the ValidatingWebhookConfiguration below.
+
+genCA has two args - first is cert commonName and second is validity in days (9125 = ~25 years)
```
erikgb, line 9:
```suggestion
genCA has two args - first is cert CN and second is validity in days (9125 = ~25 years)
```
Nit: To be consistent in the terminology.

erikgb (approved):
/lgtm
/approve
/hold

Thanks Ash! Adding a hold in case someone else is reviewing. Feel free to unhold. I will search for issues I think could be closed when merging this (vintage) PR.

erikgb:
I think merging this PR should close https://github.com/cert-manager/trust-manager/issues/382 https://github.com/cert-manager/trust-manager/issues/132.

deploy/charts/trust-manager/templates/webhook.yaml
SgtCoDFish: I've changed to "common name (CN)" so the abbreviation below is defined - good suggestion, thanks 👍

erikgb:
/lgtm

Maybe update the description to auto-close the relevant issues before merging?

SgtCoDFish:
/unhold

I'll close the relevant issues manually with a message 👍

PR #161, opened 2023-08-24
Upgrade to kubernetes 1.28 & c/r 0.16

inteon:


Makefile
```diff
@@ -182,31 +181,30 @@ $(BINDIR)/validate-trust-package: cmd/validate-trust-package/main.go pkg/fspkg/p
...
+depend: $(BINDIR)/kubectl
+depend: $(BINDIR)/kubebuilder/bin/kube-apiserver
 
 $(BINDIR)/deepcopy-gen: | $(BINDIR)
-	go build -o $@ k8s.io/code-generator/cmd/deepcopy-gen
+	cd hack/tools && go build -o $@ k8s.io/code-generator/cmd/deepcopy-gen
 
 $(BINDIR)/controller-gen: | $(BINDIR)
-	go build -o $@ sigs.k8s.io/controller-tools/cmd/controller-gen
+	cd hack/tools && go build -o $@ sigs.k8s.io/controller-tools/cmd/controller-gen
```
erikgb, line 198: I think it makes sense to control the version of `controller-gen`, as the version is included in the CRDs.

pkg/bundle/controller.go
```diff
@@ -130,14 +67,21 @@ func AddBundleController(ctx context.Context, mgr manager.Manager, opts Options)
...
+			handler.OnlyControllerOwner(),
+		), builder.OnlyMetadata).
+
+		////// Sources //////
 
 		// Reconcile trust.cert-manager.io Bundles
-		WatchesRawSource(&source.Informer{Informer: bundleInformer}, &handler.EnqueueRequestForObject{}).
+		Watches(&trustapi.Bundle{}, &handler.EnqueueRequestForObject{}).
 
 		// Watch all Namespaces. Cache whole Namespaces to include Phase Status.
```
erikgb, line 82: Does the last really sentence make sense here now? Delete?

erikgb (commented):
This is great @inteon, thanks! A couple of nit comments.

Makefile
inteon: It is still versioned, in the hack/tools/go.mod file.

pkg/bundle/controller.go
inteon: I think it is still correct, we watch all Namespaces and we watch the full objects (not just the metadata).

Makefile
erikgb: Thanks for the explanation! Still learning the setup.

pkg/bundle/controller.go
erikgb: It is still correct, but not really relevant here in the code. And you removed a corresponding comment for configmaps "Only cache metadata". But just a nit, so feel free to ignore.

inteon: Indeed, most of this namespacing logic has been moved to `cmd/trust-manager/app/app.go`.
I think the comment is probably still fine.
Also, I did change the ConfigMap and Secret watches so they watch full objects instead of metadata only in the "trusted" namespace.

erikgb:
/hold
/lgtm

Adding a hold to avoid immediate merge - in case I got new permissions now. Feel free to unhold.

erikgb:
/lgtm

(testing permissons)

PR #162, opened 2023-08-26
Refactor CM binary data reconcile preparing for PKCS#12 support

erikgb:
Before filing a PR to add support for PKCS#12 truststores (ref. remaining in https://github.com/cert-manager/trust-manager/issues/17), I think it makes sense to slightly refactor the JKS truststore support introduced in https://github.com/cert-manager/trust-manager/pull/122. I think it will make the follow-up PR easier to review.


jakexks (commented):
/ok-to-test

jakexks (approved):
Private interface refactor with passing tests so lgtm

/lgtm

PR #163, opened 2023-08-26
Add support for PKCS12 truststores

erikgb:
This adds support for PKCS12 truststores, which together with https://github.com/cert-manager/trust-manager/pull/122 closes https://github.com/cert-manager/trust-manager/issues/17. A million thanks to @aidy for https://github.com/cert-manager/trust-manager/pull/122 - this PR contains copy-modified tests from https://github.com/cert-manager/trust-manager/pull/122. 🎉

I am suggesting password-less PKCS truststores as the default (currently hard-coded), as this is the new default truststore format from Java 18, ref. https://bugs.openjdk.org/browse/JDK-8275252.

IMO we should review the API around target ConfigMap keys. The code, and especially the tests are becoming complex and somehow redundant. E.g. we could use API mechanisms more to validate unique keys in the target configmap.


pkg/bundle/sync.go
```diff
@@ -239,6 +234,27 @@ func (e jksEncoder) alias(derData []byte, friendlyName string) string {
...
+
+	var entries []pkcs12.TrustStoreEntry
+	for _, c := range cas {
+		entries = append(entries, pkcs12.TrustStoreEntry{
+			Cert:         c,
+			FriendlyName: certAlias(c.Raw, c.Subject.String()),
+		})
+	}
+
+	return pkcs12.EncodeTrustStoreEntries(rand.Reader, entries, DefaultPKCS12Password)
```
aidy, line 255:
```suggestion
	return pkcs12.EncodeTrustStoreEntries(rand.Reader, entries, string(p.password))
```

pkg/bundle/sync.go
```diff
@@ -261,9 +277,13 @@ func (b *bundle) syncTarget(ctx context.Context, log logr.Logger,
...
-	if target.AdditionalFormats != nil && target.AdditionalFormats.JKS != nil {
-		binData = map[string]binaryDataEncoder{
-			target.AdditionalFormats.JKS.Key: &jksEncoder{password: []byte(DefaultJKSPassword)},
+	if target.AdditionalFormats != nil {
+		binData = make(map[string]binaryDataEncoder, 2)
+		if target.AdditionalFormats.JKS != nil {
+			binData[target.AdditionalFormats.JKS.Key] = &jksEncoder{password: []byte(DefaultJKSPassword)}
+		}
+		if target.AdditionalFormats.PKCS12 != nil {
+			binData[target.AdditionalFormats.PKCS12.Key] = &pkcs12Encoder{password: []byte(DefaultJKSPassword)}
```
aidy, line 286:
```suggestion
			binData[target.AdditionalFormats.PKCS12.Key] = &pkcs12Encoder{password: []byte(DefaultPKCS12Password)}
```

pkg/bundle/sync.go
erikgb: Thanks! 😊

pkg/bundle/sync.go
erikgb: Thanks! 😊

PR #164, opened 2023-08-28
Add support for PKCS12 truststores (update CRD)

erikgb:
This a follow-up PR to https://github.com/cert-manager/trust-manager/pull/163. Somehow I (and also reviewers) did not see that the Bundle CRD wasn't updated. 🤣 Sorry! In my defense, I can only say that the controllers I usually work with (on-prem) have CI to verify that generated files are up-to-date. I will create an issue suggesting the same in this project.

I should probably also update the API docs somewhere, or is that automated?

### DIFF
--- a/deploy/charts/trust-manager/templates/trust.cert-manager.io_bundles.yaml
+++ b/deploy/charts/trust-manager/templates/trust.cert-manager.io_bundles.yaml
@@ -110,6 +110,15 @@ spec:
                             key:
                               description: Key is the key of the entry in the object's `data` field to be used.
                               type: string
+                        pkcs12:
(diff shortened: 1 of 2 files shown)


inteon:
@erikgb I'm also surprised that CI did not notice this. Might be a bug?

erikgb:
> @erikgb I'm also surprised that CI did not notice this. Might be a bug?

I am about to check if we have CI verifying that generated files are up-to-date. I am guessing it's just missing, but I can fix that.

inteon:
/approve
/lgtm
Thank you for noticing and fixing!

Issue #165, opened 2023-08-28
Should have CI to verify generated files are up-to-date


erikgb:
When filing https://github.com/cert-manager/trust-manager/pull/163, I somehow didn't realize that the generated CRDs were not up-to-date 🦆, and had to fix this in https://github.com/cert-manager/trust-manager/pull/164. This is unfortunate, and I think we should ensure that the project is consistent in CI.

With some guidance from a maintainer, I was linked to the CI configuration: https://github.com/jetstack/testing/tree/master/config/jobs/cert-manager/trust-manager

Definition of done: CI that fails if the following generated artifacts are not updated:

- CRDs
- Deepcopy functions
- Helm docs
- Boilerplate header in source files
- Other?

PR #166, opened 2023-08-28
ci: verify all generated files are up-to-date

erikgb:
Closes https://github.com/cert-manager/trust-manager/issues/165

Also fixing leftovers (deepcopy gen) from https://github.com/cert-manager/trust-manager/pull/163 bymigrating from `deepcopy-gen` to `controller-gen` - which can also generate the needed code.


hack/verify-generated.sh
```diff
@@ -18,21 +18,9 @@ set -o errexit
...
-trap '{ rm -f -- "$TEMP_FILE"; }' EXIT
-
-$HELM_DOCS_BIN ./deploy/charts/trust-manager -d -l error > ${TEMP_FILE}
-
-if ! cmp -s "./deploy/charts/trust-manager/README.md" "${TEMP_FILE}"; then
-  echo "Helm chart README.md is out of date."
-  echo "Please run 'make update-helm-docs'."
+git --no-pager diff
+git status --short --untracked-files=no
+if [ ! -z "$(git status --short --untracked-files=no)" ]; then
```
inteon, line 23: I remember that @wallrj had some issues with this git status setup.
I think at that point we settled for a script like the script here: https://github.com/cert-manager/approver-policy/blob/main/hack/util/verify.sh

And a verify target like this:
```make
verify:
	./hack/util/verify.sh $(MAKE) -s generate-all
```

inteon: I think that the issue was as follows:
- not considering untracked-files might mean that you miss files that have to be freshly created
- considering untracked-files means that local changes/ files will cause the check to fail, which was undesirable for development.

Not sure if I remember this 100% correctly.

hack/verify-generated.sh
```diff
@@ -18,21 +18,9 @@ set -o errexit
...
+  echo -e "\e[31mGenerated files do not match those committed. Please run 'make generate-all' locally and commit the changes. \e[0m"
```
SgtCoDFish, line 24: comment: As I'm sure you know, this is a stronger check than just checking if the generated files are out of date - but I think it's a check we should make anyway. Ideally files shouldn't be getting modified in CI so in CI it's good to verify they're not.

I might be wrong, but I think this would also fail if I the code generation was up-to-date, I'd edited something unrelated to code generation and I ran `make verify` locally, as part of development? That seems unfortunate - I use targets like `make verify` quite often when I'm developing a new feature.

This new way of validating would then mean I'd have to commit my changes before CI could pass. I suppose that's not the end of the world!

SgtCoDFish (approved):
/lgtm
/approve

Thanks! I think checking git has the downside of potentially interfering with local development, but it's better to have a solution for this than not!

SgtCoDFish:
/hold

for tim's comments!

hack/verify-generated.sh
erikgb: I'll take a look and fix. Thanks for the pointers.

erikgb: I have now made a second attempt - based on the referenced work in approver-policy. PTAL!

hack/verify-generated.sh
erikgb: True, but outdated comment now. 😉

erikgb:
@inteon There seem to have been something funky about the deepcopy gen, so I migrated to using controller-gen. Please let me know if I should put this is a separate PR.

erikgb:
I have now fixed the excluded paths as noted in today's stand-up and updated the PR description. PTAL!

inteon (commented):
/approve
/lgtm

PR #169, opened 2023-08-30
Filter resources triggered by namespace

inteon:
When a namespace changes, we only want to re-reconcile all bundles that have that namespace as a target.


pkg/bundle/bundle.go
```diff
@@ -244,3 +244,14 @@ func (b *bundle) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result,
 
 	return ctrl.Result{}, b.client.Status().Update(ctx, &bundle)
 }
+
+func (b *bundle) bundleTargetNamespaceSelector(bundleObj *trustapi.Bundle) (labels.Selector, error) {
+	namespaceSelector := labels.Everything()
```
erikgb, line 249: Nit: Is this variable needed? Why not just inline it?

inteon: I updated the code, also made some more changes. PTAL

erikgb (approved):
/lgtm

PR #170, opened 2023-08-30
Refactor util functions in preparation for SSA

inteon:
The updated util functions apply more directly on the fields of the bundle instead on the full bundle object.

### DIFF
--- a/pkg/apis/trust/v1alpha1/types_bundle.go
+++ b/pkg/apis/trust/v1alpha1/types_bundle.go
@@ -17,7 +17,6 @@ limitations under the License.
 package v1alpha1
 
 import (
-	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
@@ -163,7 +162,7 @@ type BundleCondition struct {
 	Type BundleConditionType `json:"type"`
 
 	// Status of the condition, one of ('True', 'False', 'Unknown').
-	Status corev1.ConditionStatus `json:"status"`
+	Status metav1.ConditionStatus `json:"status"`
 
 	// LastTransitionTime is the timestamp corresponding to the last status
 	// change of this condition.
--- a/pkg/bundle/bundle.go
+++ b/pkg/bundle/bundle.go
@@ -236,13 +236,14 @@ func (b *bundle) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, 
 	}
 
 	syncedCondition := trustapi.BundleCondition{
-		Type:    trustapi.BundleConditionSynced,
-		Status:  metav1.ConditionTrue,
-		Reason:  "Synced",
-		Message: message,
+		Type:               trustapi.BundleConditionSynced,
+		Status:             metav1.ConditionTrue,
+		Reason:             "Synced",
+		Message:            message,
+		ObservedGeneration: bundle.Generation,
 	}
 
-	if !needsUpdate && bundleHasCondition(&bundle, syncedCondition) {
+	if !needsUpdate && bundleHasCondition(bundle.Status.Conditions, syncedCondition) {
 		return ctrl.Result{}, nil
 	}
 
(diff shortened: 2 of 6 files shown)


pkg/bundle/util.go
```diff
@@ -48,48 +48,52 @@ func bundleHasCondition(bundle *trustapi.Bundle, condition trustapi.BundleCondit
 // bundle object.
 // LastTransitionTime will not be updated if an existing condition of the same
 // Type and Status already exists.
-func (b *bundle) setBundleCondition(bundle *trustapi.Bundle, condition trustapi.BundleCondition) {
+func (b *bundle) setBundleCondition(conditions *[]trustapi.BundleCondition, generation int64, condition trustapi.BundleCondition) {
```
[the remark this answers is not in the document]
inteon: I updated the PR based on what I found in some other projects.
What I did there is quite opposite from what you are asking unfortunately 😓.
I made all fields of the condition positional parameters because that way it was clear what fields are expected and what fields are not expected.

pkg/apis/trust/v1alpha1/types_bundle.go
```diff
@@ -163,7 +162,7 @@ type BundleCondition struct {
 	Type BundleConditionType `json:"type"`
 
 	// Status of the condition, one of ('True', 'False', 'Unknown').
-	Status corev1.ConditionStatus `json:"status"`
+	Status metav1.ConditionStatus `json:"status"`
```
inteon, line 165: I think this was a mistake in the original code base.

erikgb: > I think this is a mistake in the original code base.

I can agree with this.

pkg/bundle/util.go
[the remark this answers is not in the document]
erikgb: What about two functions? One for creating the condition you want to be present as part of the reconcile logic ("constructor") and another for merging it into the bundle status?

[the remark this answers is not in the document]
inteon: Thanks for the example. I updated the PR to match how https://github.com/kubernetes/apimachinery/blob/1629459297c77f725c47f7121851db2ee94572bd/pkg/api/meta/conditions.go#L30 works.

pkg/bundle/util.go
```diff
@@ -44,52 +44,72 @@ func bundleHasCondition(bundle *trustapi.Bundle, condition trustapi.BundleCondit
...
-	condition.ObservedGeneration = bundle.Generation
-
-	var updatedConditions []trustapi.BundleCondition
-	for _, existingCondition := range bundle.Status.Conditions {
-		// Ignore any existing conditions which don't match the incoming type and
-		// add back to set.
-		if existingCondition.Type != condition.Type {
-			updatedConditions = append(updatedConditions, existingCondition)
+func (b *bundle) setBundleCondition(
+	existingConditions []trustapi.BundleCondition,
```
erikgb, line 50: Do we gain something by supplying the same slice two times (pointer and non-pointer), or can this parameter be removed? Please let me know if you think we've had enough discussion on this. 😉

inteon: This is because in the SSA PR, these arrays will be different.
The first array will be the conditions that we get from the API server, the second array are the conditions in the patch object.

erikgb: Aha, thanks for the explanation!

pkg/bundle/bundle.go
```diff
@@ -238,7 +248,17 @@ func (b *bundle) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result,
...
-	b.setBundleCondition(&bundle, syncedCondition)
+	b.setBundleCondition(
+		bundle.Status.Conditions,
+		&bundle.Status.Conditions,
+		trustapi.BundleCondition{
+			Type:               syncedCondition.Type,
+			Status:             syncedCondition.Status,
+			Reason:             syncedCondition.Reason,
+			Message:            syncedCondition.Message,
+			ObservedGeneration: bundle.Generation,
```
erikgb, line 259: Could it make sense to set `ObservedGeneration` when creating `syncedCondition` - to simplify the code? Or is there a reason for doing it like this?

erikgb (approved):
/lgtm
/hold

Just one remaining question from me. Feel free to unhold.

pkg/bundle/bundle.go
inteon: I also refactored bundleHasCondition and moved the ObservedGeneration to syncedCondition

PR #171, opened 2023-08-31
Move from k8s.io/utils/pointer to k8s.io/utils/ptr

inteon:
`k8s.io/utils/pointer` is deprecated


erikgb (approved):
/lgtm

Nice! Do we have a standard for organizing the import declarations - beyond `go fmt`?

inteon:
> /lgtm
> 
> Nice! Do we have a standard for organizing the import declarations - beyond `go fmt`?

I don't think so. golangci-lint has a gci linter that we could use.

PR #172, opened 2023-08-31
Fix misinterpretation, ByObject cache settings are GVK specific

inteon:
In https://github.com/cert-manager/trust-manager/pull/161, I switched to a single cache that uses the ByObject cache option.
I did no realise that partial metadata settings collide with the typed object settings.
This PR fixes the confusing config and adds a TODO note.
We aim to fix the TODO in https://github.com/cert-manager/trust-manager/pull/89

### DIFF
--- a/cmd/trust-manager/app/app.go
+++ b/cmd/trust-manager/app/app.go
@@ -23,7 +23,6 @@ import (
 
 	"github.com/spf13/cobra"
 	corev1 "k8s.io/api/core/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 	clientv1 "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/tools/record"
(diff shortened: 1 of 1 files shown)


cmd/trust-manager/app/app.go
```diff
@@ -103,18 +102,10 @@ func NewCommand() *cobra.Command {
...
+							// TODO: create a seperate cache for targets and sources and only
+							// cache full config maps for the sources.
 						},
-						// Only cache partial metadata for targets.
-						&metav1.PartialObjectMetadata{
-							TypeMeta: metav1.TypeMeta{
-								APIVersion: "v1",
-								Kind:       "ConfigMap",
-							},
-						}: {},
```
erikgb, line 117: Does it work to set the `Namespaces` field on the cache options here? And instead remove the typed cache above? Wouldn't that fix the TODO, or am I not understanding how this works?

inteon: We want to fully cache the configmaps in the "trusted namespace", but also want to cache the metadata of all configmaps across all namespaces.
It does not seem like that is achievable using the current API.

inteon: Setting the Namespaces field will break the all-namespace metadata caching.

erikgb: Thanks for reminding me that the goal here is to manage configmaps throughout the cluster. 😊

erikgb (approved):
/lgtm

PR #174, opened 2023-09-05
Allow user to specify the name of cert-manager's ServiceAccount

SgtCoDFish:
The cert-manager ServiceAccount is not fixed and so can vary between different installations. [1] [2]

cert-manager users might well want to customise it, and we should support those users in trust-manager!

ℹ️ This also removes the namespace from the CertificateRequestPolicy, since that resource is cluster scoped and so passing a namespace here is a no-op!

[1] https://github.com/cert-manager/cert-manager/blob/cab2b3b68ca834b60931fd76d1fb74f757a03550/deploy/charts/cert-manager/templates/serviceaccount.yaml#L10
[2] https://github.com/cert-manager/cert-manager/blob/cab2b3b68ca834b60931fd76d1fb74f757a03550/deploy/charts/cert-manager/values.yaml#L108

### DIFF
--- a/deploy/charts/trust-manager/values.yaml
+++ b/deploy/charts/trust-manager/values.yaml
@@ -78,8 +78,11 @@ app:
         # -- Whether to create an approver-policy CertificateRequestPolicy allowing auto-approval of the trust-manager webhook certificate. If you have approver-policy installed, you almost certainly want to enable this.
         enabled: false
 
-        # -- Namespace in which cert-manager was installed. Only used if approverPolicy has been enabled.
+        # -- Namespace in which cert-manager was installed. Only used if app.webhook.tls.approverPolicy.enabled is true
         certManagerNamespace: "cert-manager"
+
+        # -- Name of cert-manager's ServiceAccount. Only used if app.webhook.tls.approverPolicy.enabled is true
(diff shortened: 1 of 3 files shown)


deploy/charts/trust-manager/values.yaml
```diff
@@ -78,7 +78,10 @@ app:
         # -- Whether to create an approver-policy CertificateRequestPolicy allowing auto-approval of the trust-manager webhook certificate. If you have approver-policy installed, you almost certainly want to enable this.
         enabled: false
 
-        # -- Namespace in which cert-manager was installed. Only used if approverPolicy has been enabled.
+        # -- Name of cert-manager's ServiceAccount. Only used if app.webhook.tls.approverPolicy.enabled is true
+        certManagerServiceAccount: "cert-manager"
```
erikgb, line 82: Nit: Maybe put this value below the `certManagerNamespace` value? Seems more conceptually correct to me. 😉

erikgb (commented):
Just a minor nit comment, but absolutely no blocker. Thanks!

erikgb (approved):
/lgtm

wallrj (approved):
Thanks Ash.

/approve
/lgtm

PR #177, opened 2023-09-08
Release v0.7.0-alpha.0

inteon:
preparing v0.7.0-alpha.0 release by replacing all versions


erikgb (approved):
/lgtm

Thanks, Tim! This will allow me (and others) to early-test the SSA migration. I would expect the first alpha version of version 0.7.0 to be 0.7.0-alpha.**1** (ref. https://semver.org/), but if you already have a common pattern for these pre-releases, I am find with that.

PR #178, opened 2023-09-10
BUGFIX: fix migration from csa to ssa

inteon:
@erikgb noticed that the client-side-apply to server-side-apply migration functions are not working as expected (https://kubernetes.slack.com/archives/CDEQJ0Q8M/p1694172356236499).
In https://github.com/cert-manager/trust-manager/pull/89, I made the wrong assumption that the fieldmanager was set to "trust-manager" in previous versions.
In this PR, I fixed this mistake. I also added a test to make sure that the confimap resources are also correctly migrated.

### DIFF
--- a/pkg/bundle/sync.go
+++ b/pkg/bundle/sync.go
@@ -55,7 +55,11 @@ const (
 	DefaultPKCS12Password = ""
 )
 
-const fieldManager = "trust-manager"
+const (
+	// oldFieldManager is the field manager that was used by trust-manager before the migration to the Apply field manager.
+	oldFieldManager = "Go-http-client"
+	fieldManager    = "trust-manager"
+)
 
 type notFoundError struct{ error }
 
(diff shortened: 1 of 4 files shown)


pkg/bundle/sync.go
```diff
@@ -509,18 +508,63 @@ func (b *bundle) patchResource(ctx context.Context, obj interface{}) error {
...
+			continue
+		}
+		needsUpdate = true
+	}
+	if !needsUpdate {
+		return false, nil
+	}
+
+	var cm corev1.ConfigMap
+	if err := b.directClient.Get(ctx, client.ObjectKeyFromObject(obj), &cm); err != nil {
```
erikgb, line 553: Is this the reason for introducing an additional client (`directClient`)? My guess is that you are trying to avoid polluting the cache with all full configmaps in the cluster. Did you consider disabling cache for configmaps - to prevent the additional client, or will this not work?

inteon: Maybe it can be done without caching all the full objects, but I just don't want to add the complexity of such an exception. Also, this migration can be a temporary solution that is removed after a few versions. I feel like this new uncached client is the most elegant solution (please let me know if you don't agree).

erikgb: Yeah, good point that this migration code is going to be deleted after a couple of releases. But with an additional client, the SSA code could be greatly simplified - if the additional client was a standard K8s client-go client - with support for apply confgurations. 😉

erikgb (approved):
/lgtm

Issue #155, opened 2023-08-01
de-duplicate CA from the trust bundle


Jiawei0227:
It is possible that the trust bundle contains multiple sources which point to the same CA. It would be nice if we can dedup in the final trust bundle

SgtCoDFish:
De-duplication is definitely a worthwhile thing to have in the case that we have two identical certificates in the resulting bundle. I've marked this as a good first issue in case anyone wanted to take a look - I think it'd be worth chatting about how we'd expose the API for this but it shouldn't be too hard to do this and save a few bytes!

arsenalzp:
Hello,
Any deadlines for this issue? First of all I would like to examine entire project before start to work with even minor issue.

SgtCoDFish:
> Any deadlines for this issue?

Not really, no! This feature would be a minor improvement which could save a little storage space, so it's not really a thing which we need in a hurry!

arsenalzp:
I take it, please assign it to me!

arsenalzp:
PR related to this issue was created.
Let's discuss about de-duplication method?
In my code I used hash `sha256` as unique signature of certificates; additional it's possible to check certificates serial numbers, however serial number might be set manually during signing CSR.
It is also possible to put  each certificate into `x509 Certificates Pool` then check uniqueness of each certificates against certificates in the pool; I decided to don't use this solution due to its memory consuming: `x509 Certificates Pool` consume much more memory than just a set of certificates hash sums, but it is possible to have hash collision, so let's discuss about solution in my PR.

SgtCoDFish:
> PR related to this issue was created.

Amazing, great!

> Let's discuss about de-duplication method? In my code I used hash sha256 as unique signature of certificates

That's what I'd have done too. It's sometimes called the "sha256 fingerprint". The fingerprint is the way to do this - checking based on serial number or anything else is basically a subtle way of introducing little bugs IMO 😁

erikgb:
I believe this was fixed by https://github.com/cert-manager/trust-manager/pull/184

/close

PR #184, opened 2023-09-23
Add certificates deduplication feature

arsenalzp:
This PR is related to the issue `de-duplicate CA from the trust bundle` #155 
It add feature of certificates de-duplication, meaning to exclude certificates which were already added to a certificates bundle, thus save memory.

### DIFF
--- a/pkg/util/cert_pool.go
+++ b/pkg/util/cert_pool.go
@@ -1,0 +1,107 @@
+/*
+Copyright 2022 The cert-manager Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package util
+
+import (
+	"crypto/sha256"
+	"crypto/x509"
+	"encoding/pem"
+	"fmt"
+)
+
+// CertPool is a set of certificates.
+type certPool struct {
+	certificatesHashes map[[32]byte]struct{}
+	certificates       []*x509.Certificate
+}
+
+// NewCertPool returns a new, empty CertPool.
+func newCertPool() *certPool {
+	return &certPool{
+		certificatesHashes: make(map[[32]byte]struct{}),
+		certificates:       make([]*x509.Certificate, 0),
+	}
+}
+
+// Check if the given certificate was added to certificates bundle already
+func (cp *certPool) isCertificateDuplicate(certData []byte) bool {
+	// calculate hash sum of the given certificate
+	hash := sha256.Sum256(certData)
+
+	// check a hash existence
+	if _, ok := cp.certificatesHashes[hash]; ok {
+		return ok
+	}
+
+	// put certificate hash into a set of hashes
+	cp.certificatesHashes[hash] = struct{}{}
+
+	return false
+}
+
+// Append certificate to a pool
+func (cp *certPool) appendCertFromPEM(PEMdata []byte) error {
+	if PEMdata == nil {
+		return fmt.Errorf("certificate data can't be nil")
+	}
+
+	for {
+		var block *pem.Block
+		block, PEMdata = pem.Decode(PEMdata)
+
+		if block == nil {
+			break
+		}
+
+		if block.Type != "CERTIFICATE" {
+			// only certificates are allowed in a bundle
+			return fmt.Errorf("invalid PEM block in bundle: only CERTIFICATE blocks are permitted but found '%s'", block.Type)
+		}
+
+		if len(block.Headers) != 0 {
+			return fmt.Errorf("invalid PEM block in bundle; blocks are not permitted to have PEM headers")
+		}
+
+		certificate, err := x509.ParseCertificate(block.Bytes)
+		if err != nil {
+			// the presence of an invalid cert (including things which aren't certs)
+			// should cause the bundle to be rejected
+			return fmt.Errorf("invalid PEM block in bundle; invalid PEM certificate: %w", err)
+		}
+
+		if certificate == nil {
+			return fmt.Errorf("failed appending a certificate: certificate is nil")
+		}
+
+		if !cp.isCertificateDuplicate(block.Bytes) {
+			cp.certificates = append(cp.certificates, certificate)
+		}
+	}
+
+	return nil
+}
+
+// Get PEM certificates from pool
+func (cp *certPool) getCertsPEM() [][]byte {
+	var certsData [][]byte = make([][]byte, len(cp.certificates))
+
+	for i, cert := range cp.certificates {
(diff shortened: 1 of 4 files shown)


pkg/util/pem.go
```diff
@@ -120,3 +125,19 @@ func DecodeX509CertificateChainBytes(certBytes []byte) ([]*x509.Certificate, err
 
 	return certs, nil
 }
+
+// Check if the given certificate was added to certificates bundle already
+func IsCertificateDuplicate(certsHash map[[32]byte]struct{}, certData []byte) bool {
```
erikgb, line 130: What do you think about creating a type for this function? Maybe it could be named `certPool`?  😉 With an `appendCertFromPEM` func, the type could also hold the certificates returned from the outer function.

pkg/util/pem.go
```diff
@@ -120,3 +125,19 @@ func DecodeX509CertificateChainBytes(certBytes []byte) ([]*x509.Certificate, err
...
+	// calculate hash sum of the given certificate
+	hash := sha256.Sum256(certData)
```
erikgb, line 132: I note that `crypto.CertPool` uses `sha256.Sum224` for deduplicating certs. Any particular reason you went for something else?

erikgb (commented):
@arsenalzp, thanks for this! This is almost LGTM from me. I have just a few comments/questions.

pkg/util/pem.go
SgtCoDFish: Interesting that CertPool uses sha224! I think sha256 is fine here though, it's what I'd have reached for as a "default"

arsenalzp: Hello,
Thank you so much for your feedback, it is very appreciated!
There weren't any reasons to use sha224 against sha256, they have almost equal computation complexity, bus some differences have present:

https://www.rfc-editor.org/rfc/rfc3874.txt

> 1.1.  Usage Considerations
> 
>    Since SHA-224 is based on SHA-256, roughly the same amount of effort
>    is consumed to compute a SHA-224 or a SHA-256 digest message digest
>    value.  Even though SHA-224 and SHA-256 have roughly equivalent
>    computational complexity, SHA-224 is an appropriate choice for a
>    one-way hash function that provides 112 bits of security.  The use of
>    a different initial value ensures that a truncated SHA-256 message
>    digest value cannot be mistaken for a SHA-224 message digest value
>    computed on the same data.
> 
>    Some usage environments are sensitive to every octet that is
>    transmitted.  In these cases, the smaller (by 4 octets) message
>    digest value provided by SHA-224 is important.
> 
>    These observations lead to the following guidance:
> 
>    *  When selecting a suite of cryptographic algorithms that all offer
>       112 bits of security strength, SHA-224 is an appropriate choice
>       for one-way hash function.
> 
>    *  When terseness is not a selection criteria, the use of SHA-256 is
>       a preferred alternative to SHA-224.

arsenalzp: For our purposes we have to choice algorithm which proposes:
- acceptable computation complexity and
- minimize hash collisions

erikgb: I'm fine with sha256. 😃

pkg/util/cert_pool.go
```diff
@@ -0,0 +1,92 @@
...
+
+import (
+	"crypto/sha256"
+	"crypto/x509"
+	"encoding/pem"
+	"fmt"
+)
+
+// CertPool is a set of certificates.
+type CertPool struct {
```
erikgb, line 27: Ref. https://github.com/cert-manager/trust-manager/pull/184#discussion_r1335625480 I think this type should be defined in an internal package. Or as an unexported type in the package it's used. We mustn't tempt anyone to use it. :-)

pkg/util/pem.go
arsenalzp: I added `CertPool` type type with related methods.
It's not possible to put that structure into `internal` directory, due to limitation of importing packages which are located under `internal` directory.

pkg/util/cert_pool.go
arsenalzp: Sounds good!
Let's do it!

arsenalzp: `CertPool` and its methods to unreported ones, so `certPool` is used by `util` package only

pkg/util/pem.go
```diff
@@ -87,7 +88,10 @@ func ValidateAndSplitPEMBundle(data []byte) ([][]byte, error) {
 			return nil, fmt.Errorf("invalid PEM block in bundle; invalid PEM certificate: %w", err)
 		}
 
-		certificates = append(certificates, pem.EncodeToMemory(b))
+		// check the duplicate certificates
+		if !certPool.IsCertificateDuplicate(b.Bytes) {
```
erikgb, line 92: Some changes missing here? I think the idea was to use the new certPool type.

arsenalzp: Ah, sure! I didn't recognize your requirement properly!
I'm going to fix the code.

erikgb: Why can't you just ` return certPool.getCertsPEM()` from `ValidateAndSplitPEMBundle`? Am I missing something? 🤔

arsenalzp: Oh, in that case we can! I thought you wanted to return `certPool`

arsenalzp: > Why can't you just `return certPool.getCertsPEM()` from `ValidateAndSplitPEMBundle`? Am I missing something? 🤔

I fixed a code as you required, all tests were passed.
Could you please review?

pkg/util/cert_pool.go
```diff
@@ -0,0 +1,112 @@
...
+	for {
+		var block *pem.Block
+		block, PEMdata = pem.Decode(PEMdata)
+
+		if block == nil {
+			// //If no PEM data is found,
+			// // p is nil and the whole of the input is returned in rest.
+			// if len(PEMdata) != 0 {
+			// 	return fmt.Errorf("error of decoding PEM block")
+			// }
```
erikgb, line 71: Is the out-commented code intentional? I am not very familiar with this PEM decoding, but if this code is not worth having, I think we should delete it.

erikgb (commented):
This looks good, thanks for addressing the review comments. I have just a question about some commented code.

/ok-to-test

pkg/util/cert_pool.go
arsenalzp: > Is the out-commented code intentional? I am not very familiar with this PEM decoding, but if this code is not worth having, I think we should delete it.

Yes of course, I will fix it.
Thank you so much for your help, hints and patience, it is very appreciated!

arsenalzp: > Is the out-commented code intentional? I am not very familiar with this PEM decoding, but if this code is not worth having, I think we should delete it.

Hello,
Comments were trimmed as you requested.

erikgb (approved):
/lgtm
/cc @SgtCoDFish 

Thanks @arsenalzp! With this new certPool type I think this can be improved even further.

arsenalzp:
> /lgtm /cc @SgtCoDFish
> 
> Thanks @arsenalzp! With this new certPool type I think this can be improved even further in a follow-up PR.
> 
> Maybe you can rebase and squash commits?

It was done.

SgtCoDFish (approved):
/lgtm
/approve

Thanks so much @arsenalzp (and @erikgb for reviewing)! 😁

arsenalzp:
> /lgtm /approve
> 
> Thanks so much @arsenalzp (and @erikgb for reviewing)! 😁

Thank you for your assistance!

PR #187, opened 2023-09-26
Eliminate dedicated ConfigMap target cache

erikgb:
This PR eliminates the dedicated target cache and fixes the TODO related to source/target ConfigMap caches. Now the cache should only cache full configmaps in source namespace and cache configmap metadata cluster-wide.

I wished to simplify this a bit before anyone tries to add secrets as targets.

I am pretty sure that ConfigMap and PartialMetadata (for configmaps) should be treated independently in caches now. But feel free to check this @inteon - if you think the tests are not covering this area well enough. I asked about this in controller-runtime channel on Slack: https://kubernetes.slack.com/archives/C02MRBMN00Z/p1695564370173869. And also had a look at https://github.com/kubernetes-sigs/controller-runtime/pull/1174. So if this isn't working as expected, I think we should create an issue in controller-runtime.

/cc @inteon

### DIFF
--- a/cmd/trust-manager/app/app.go
+++ b/cmd/trust-manager/app/app.go
@@ -23,8 +23,7 @@ import (
 
 	"github.com/spf13/cobra"
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/labels"
-	"k8s.io/apimachinery/pkg/selection"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 	clientv1 "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/tools/record"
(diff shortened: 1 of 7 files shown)


cmd/trust-manager/app/app.go
```diff
@@ -95,48 +94,34 @@ func NewCommand() *cobra.Command {
...
+						},
 						&corev1.Secret{}: {
 							// Only cache full Secrets in the "watched" namespace.
 							Namespaces: map[string]cache.Config{
 								opts.Bundle.Namespace: {},
 							},
 						},
-						&corev1.ConfigMap{}: {
-							// Watch full config maps across all namespaces.
-							// TODO: create a seperate cache for targets and sources and only
```
inteon, line 108: I'm surprised that this comment is still present in the codebase. Since it has been fixed &a separate cache was added.

cmd/trust-manager/app/app.go
```diff
@@ -95,48 +94,34 @@ func NewCommand() *cobra.Command {
 				Cache: cache.Options{
 					ReaderFailOnMissingInformer: true,
 					ByObject: map[client.Object]cache.ByObject{
+						// Cache metadata for ConfigMaps cluster-wide
+						&metav1.PartialObjectMetadata{
+							TypeMeta: metav1.TypeMeta{
```
inteon, line 99: I fear that controller runtime does not differentiate between the partial metadata object and the typed object. There will be a config collision.

inteon: See #172

inteon: This is the c/r code: https://github.com/kubernetes-sigs/controller-runtime/blob/v0.16.2/pkg/cache/cache.go#L304-L316

inteon (changes requested):
I propose we keep the dedicated cache until c/r supports differentiating between partial metadata objects and full objects (see https://github.com/cert-manager/trust-manager/pull/187#discussion_r1338054159).

However, I would like to remove the `// TODO:` comment already, I think this is something that we missed in https://github.com/cert-manager/trust-manager/pull/89 (I feel like it might have been lost while rebasing).

erikgb:
> I propose we keep the dedicated cache until c/r supports differentiating between partial metadata objects and full objects

@inteon agreed! Sorry for not testing this better. It got late yesterday. 😉 I'll update the PR.

erikgb:
Replaced by https://github.com/cert-manager/trust-manager/pull/188 - ref. comment above.

PR #189, opened 2023-09-28
Fix SSA migration field managers

erikgb:
This PR adapts the SSA migration code to the controller-runtime regression introduced in c/r version 0.15.0 and fixed by https://github.com/kubernetes-sigs/controller-runtime/pull/2435.

/cc @inteon

### DIFF
--- a/pkg/bundle/bundle.go
+++ b/pkg/bundle/bundle.go
@@ -339,7 +339,9 @@ func (b *bundle) migrateBundleStatusToApply(ctx context.Context, obj client.Obje
 	// subresource. We need to check for this as we need to migrate the entry to the new
 	// fieldManager.
 	isOldBundleStatusManagedFieldsEntry := func(mf *metav1.ManagedFieldsEntry) bool {
-		return mf.Manager == oldFieldManager && mf.Operation == metav1.ManagedFieldsOperationUpdate && mf.Subresource == "status"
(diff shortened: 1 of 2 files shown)


inteon:
Have you tested running an old trust-manager version to confirm that the field manager is correct? I didn't do this previously but I realised that it is more important than first thought.

erikgb:
> Have you tested running an old trust-manager version to confirm that the field manager is correct? I didn't do this previously but I realised that it is more important than first thought.

Good point. With version 0.5.0 of trust-manager, I get this:

````shell
$ k get configmaps -n default example-bundle -o json --show-managed-fields=true | jq '.metadata.managedFields'
[
  {
    "apiVersion": "v1",
    "fieldsType": "FieldsV1",
    "fieldsV1": {
      "f:data": {
        ".": {},
        "f:trust-bundle.pem": {}
      },
      "f:metadata": {
        "f:ownerReferences": {
          ".": {},
          "k:{\"uid\":\"a5814f40-1e51-422c-98b4-890874ddd0b2\"}": {}
        }
      }
    },
    "manager": "trust-manager",
    "operation": "Update",
    "time": "2023-09-28T20:01:21Z"
  }
]
````

inteon:
Thanks for finding the reason we were seeing this `Go-http-client` name in certain older versions & for making the migration path more robust.
/approve
/lgtm

PR #191, opened 2023-10-06
fix: add missing RBAC for CSA->SSA migration of bundles/status

erikgb:
When verifying v0.7.0-alpha.1 in our clusters, I notice the following error:

````
E1006 16:12:20.394650       1 bundle.go:134] trust/bundle "msg"="failed to migrate bundle status" "error"="bundles.trust.cert-manager.io \"my-bundle\" is forbidden: User \"system:serviceaccount:cert-manager:trust-manager\" cannot update resource \"bundles\" in API group \"trust.cert-manager.io\" at the cluster scope" "bundle"="my-bundle"
````

I believe this patch should fix this.

/cc @inteon

### DIFF
--- a/deploy/charts/trust-manager/templates/clusterrole.yaml
+++ b/deploy/charts/trust-manager/templates/clusterrole.yaml
@@ -23,7 +23,8 @@ rules:
   - "trust.cert-manager.io"
   resources:
   - "bundles/status"
-  verbs: ["patch"]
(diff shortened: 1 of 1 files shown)


inteon:
Thanks for fixing, we will need another alpha now 😅
/lgtm
/approve

inteon:
Thanks for fixing, we will need another alpha now 😅
/lgtm
/approve

PR #193, opened 2023-10-07
feat: support secret as a target

Jiawei0227:
Support https://github.com/cert-manager/trust-manager/issues/10

Add a new API Secret under taget to allow creating secret format CA certificates.
This is compatible with existing ConfigMap and can co-exist or exist alone.

### DIFF
--- a/deploy/charts/trust-manager/templates/clusterrole.yaml
+++ b/deploy/charts/trust-manager/templates/clusterrole.yaml
@@ -43,3 +43,25 @@ rules:
   resources:
   - "events"
   verbs: ["create", "patch"]
+
+- apiGroups:
+  - ""
+  resources:
+  - "secrets"
+  verbs: ["create", "list", "watch"]
+{{- if .Values.secretTargets.enabled }}
+{{- if .Values.secretTargets.authorizedSecretsAll }}
+- apiGroups:
+  - ""
+  resources:
+  - "secrets"
+  verbs: ["get", "update", "delete", "patch"]
+{{- else if .Values.secretTargets.authorizedSecrets }}
+- apiGroups:
+  - ""
+  resources:
+  - "secrets"
+  verbs: ["get", "update", "delete", "patch"]
+  resourceNames: {{ .Values.secretTargets.authorizedSecrets | toYaml | nindent 2 }}
+{{- end -}}
+{{- end -}}
--- a/deploy/charts/trust-manager/values.yaml
+++ b/deploy/charts/trust-manager/values.yaml
@@ -101,6 +101,20 @@ app:
     # -- If false, disables the default seccomp profile, which might be required to run on certain platforms
     seccompProfileEnabled: true
 
+secretTargets:
+  # -- If set to true, enable writing trust bundles to Kubernetes Secrets as a target.
+  #    trust-manager can only write to secrets which are explicitly allowed.
+  #    - see either authorizedSecrets or authorizedSecretsAll.
+  enabled: false
+  # -- If set to true, grant read/write permission to all secrets across the cluster. Use with caution!
+  #    When set, ignores authorizedSecrets list setting.
+  authorizedSecretsAll: false
+  # -- A list of secret names which trust-manager will be permitted to read and write across all namespaces. 
+  #    These will be the only allowable Secrets that can be used as targets. 
+  #    If the list is empty, trust-manager will not be able to write to secrets and will only be able to
+  #    read secrets in the trust namespace for use as sources.
+  authorizedSecrets: []
+
 resources: {}
   # -- Kubernetes pod resource limits for trust.
   # limits:
--- a/pkg/apis/trust/v1alpha1/types_bundle.go
+++ b/pkg/apis/trust/v1alpha1/types_bundle.go
@@ -98,6 +98,11 @@ type BundleTarget struct {
 	// data will be synced to.
 	ConfigMap *KeySelector `json:"configMap,omitempty"`
 
+	// Secret is the target Secret that all Bundle source data will be synced to.
+	// Using Secrets as targets is only supported if enabled at trust-manager startup.
+	// By default, trust-manager has no permissions for writing to secrets and can only read secrets in the trust namespace.
+	Secret *KeySelector `json:"secret,omitempty"`
+
 	// AdditionalFormats specifies any additional formats to write to the target
 	// +optional
 	AdditionalFormats *AdditionalFormats `json:"additionalFormats,omitempty"`
--- a/pkg/bundle/sync.go
+++ b/pkg/bundle/sync.go
@@ -265,11 +265,11 @@ func (e pkcs12Encoder) encode(trustBundle string) ([]byte, error) {
 	return pkcs12.EncodeTrustStoreEntries(rand.Reader, entries, e.password)
 }
 
-// syncTarget syncs the given data to the target ConfigMap in the given namespace.
+// syncConfigMapTarget syncs the given data to the target ConfigMap in the given namespace.
 // The name of the ConfigMap is the same as the Bundle.
 // Ensures the ConfigMap is owned by the given Bundle, and the data is up to date.
 // Returns true if the ConfigMap has been created or was updated.
-func (b *bundle) syncTarget(
+func (b *bundle) syncConfigMapTarget(
 	ctx context.Context,
 	log logr.Logger,
 	bundle *trustapi.Bundle,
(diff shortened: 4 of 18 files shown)


inteon:
Thanks @Jiawei0227 for rebasing this PR, could you dco signoff your commits & also add [rudexi](https://github.com/rudexi) as a co-author?

Jiawei0227:
> Thanks @Jiawei0227 for rebasing this PR, could you dco signoff your commits & also add @rudexi as a co-author?

Done. This CL is mostly a rewrite since the previous PR is completely un-rebaseable. But I am okay with co-authoring @rudexi since this PR is inspired by him

deploy/charts/trust-manager/values.yaml
```diff
@@ -101,6 +101,10 @@ app:
     # -- If false, disables the default seccomp profile, which might be required to run on certain platforms
     seccompProfileEnabled: true
 
+# -- List of secrets trust-manager is allowed to read/write. Used when the secret target is set when creating
+# a bundle. If the list is empty it means read/write access to all secrets.
+authorizedSecrets: ["ca-bundle"]
```
SgtCoDFish, line 106: suggestion: I'm not sure if this is the safest way to do this.

I have a few concerns:

- Having a default here feels wrong - since most clusters won't need to be able to read/write to secrets called `ca-bundle`. I'd rather the default be that there's nothing extra which we have access to, so that users have to _explicitly_ expand trust-manager's permissions.
- Having an empty list confer permissions to _everything_ is really confusing to me (the intuitive reading would be that an empty list means there are no allowed secrets at all)
- There doesn't seem to be an obvious way to allow no secrets, in the event that I don't want to use secrets at all.

I'd personally rather the default be "no secrets are authorized". How about something like the following?

```yaml
secretTargets:
    # -- If set to true, enable writing trust bundles to Kubernetes Secrets as a target
    enabled: false

    # -- A list of secret names which trust-manager will be permitted to read and write across all namespaces. These will be the only allowable Secrets that can be used as targets. Must be a non-empty list  if secretTargets.enabled is true
    authorizedSecrets: [""]
```

pkg/bundle/sync.go
```diff
@@ -441,7 +441,158 @@ func (b *bundle) syncTarget(
...
+
+		if target.AdditionalFormats.PKCS12 != nil {
+			encoded, err := pkcs12Encoder{password: DefaultPKCS12Password}.encode(data)
+			if err != nil {
+				return false, fmt.Errorf("failed to encode PKCS12: %w", err)
+			}
+
+			targetData[target.AdditionalFormats.PKCS12.Key] = encoded
+		}
+	}
```
SgtCoDFish, line 529: suggestion: I think stuff like this will always be identical for any target, whether ConfigMap or Secret or anything in the future. Ideally this should should be abstracted, and then `syncSecretTarget` and `syncConfigMapTarget` should just apply the generated bundle(s) to targets.

SgtCoDFish (changes requested):
Thanks for your work on this! I have a couple of suggestions - what do you think?

deploy/charts/trust-manager/values.yaml
inteon: What would be the difference between `authorizedSecrets: [""]` and `authorizedSecrets: []` ?

SgtCoDFish: Whoops, that was a typo, my bad! I think the default should be `authorizedSecrets: []`.

If enabled is set to true, we should check that authorizedSecrets is a list with at least one valid secret name. `""` wouldn't be a valid secret name.

pkg/bundle/sync.go
Jiawei0227: Done

deploy/charts/trust-manager/values.yaml
Jiawei0227: okay, I add another option authorizedSecretsAll to indicate permission to all. And now empty list of authorizedSecrets means access to nothing

deploy/charts/trust-manager/values.yaml
```diff
@@ -101,6 +101,18 @@ app:
     # -- If false, disables the default seccomp profile, which might be required to run on certain platforms
     seccompProfileEnabled: true
 
+secretTargets:
+  # -- If set to true, enable writing trust bundles to Kubernetes Secrets as a target
+  enabled: false
+  # -- If set to true, grant read/write permission to all secrets. Use it with CAUTIOUS!
+  #    When set, ignores authorizedSecrets list setting.
```
SgtCoDFish, line 108: nitpick: I think this has a typo, and I'd like to suggest a change.

```suggestion
  # -- If set to true, grant read/write permission to all secrets across the cluster. Use with caution!
  #    When set, ignores authorizedSecrets list setting.
```

deploy/charts/trust-manager/values.yaml
```diff
@@ -101,6 +101,18 @@ app:
...
+  authorizedSecretsAll: false
+  # -- A list of secret names which trust-manager will be permitted to read and write across all namespaces. 
+  #    These will be the only allowable Secrets that can be used as targets. 
+  #    If the list is empty it means read/write access to no secrets.
```
SgtCoDFish, line 112: suggestion: trust-manager will still be able to read some secrets from the "trust namespace" so that they can be used as `sources`.

```suggestion
  #    If the list is empty, trust-manager will not be able to write to secrets and will only be able to read secrets in the trust namespace for use as sources.
```

deploy/charts/trust-manager/values.yaml
```diff
@@ -101,6 +101,18 @@ app:
     # -- If false, disables the default seccomp profile, which might be required to run on certain platforms
     seccompProfileEnabled: true
 
+secretTargets:
+  # -- If set to true, enable writing trust bundles to Kubernetes Secrets as a target
+  enabled: false
```
SgtCoDFish, line 106: suggestion: I think it'd help to add a comment here indicating that users need to do something else in addition to setting this to true.

```suggestion
  # -- If set to true, enable writing trust bundles to Kubernetes Secrets as a target. trust-manager can only write to secrets which are explicitly allowed - see either authorizedSecrets or authorizedSecretsAll.
  enabled: false
```

pkg/apis/trust/v1alpha1/types_bundle.go
```diff
@@ -98,6 +98,10 @@ type BundleTarget struct {
 	// data will be synced to.
 	ConfigMap *KeySelector `json:"configMap,omitempty"`
 
+	// Secret is the target Secret in Namespaces that all Bundle source
+	// data will be synced to.
```
SgtCoDFish, line 102: suggestion: I think it'd be helpful to make clear that this isn't going to work without being enabled.

```suggestion
	// Secret is the target Secret that all Bundle source data will be synced to. Using Secrets as targets is only supported if enabled at startup - by default, trust-manager has no permissions for writing to secrets and can only read secrets in the trust namespace.
```

test/env/data.go
```diff
@@ -132,22 +132,53 @@ func NewTestBundle(ctx context.Context, cl client.Client, opts bundle.Options, t
...
 	Expect(cl.Create(ctx, &bundle)).NotTo(HaveOccurred())
 
 	return &bundle
 }
 
+// NewTestBundleSecretTarget creates a new Bundle in the API using the input test data.
+// Returns the create Bundle object.
+func NewTestBundleSecretTarget(ctx context.Context, cl client.Client, opts bundle.Options, td TestData) *trustapi.Bundle {
+	return newTestBundle(ctx, cl, opts, td, "Secret")
+}
```
SgtCoDFish, line 153: This is great 😁

test/env/data.go
```diff
@@ -39,7 +39,7 @@ import (
 )
 
 const (
-	EventuallyTimeout      = "30s"
+	EventuallyTimeout      = "90s"
```
SgtCoDFish, line 42: question: Why are we bumping this? 30s seems like it should be plenty for this use case to me, but I'd happily bump if you've been running into problems

SgtCoDFish (commented):
Few more comments - really appreciate the work you've put into this. We're nearly there 😁

pkg/apis/trust/v1alpha1/types_bundle.go
Jiawei0227: done

test/env/data.go
Jiawei0227: I run this locally, if its the first time I bring up KIND cluster it will timeout. It took around 40s etc to finish all the testing in my local.

deploy/charts/trust-manager/values.yaml
Jiawei0227: done

deploy/charts/trust-manager/values.yaml
Jiawei0227: Done

SgtCoDFish (approved):
/lgtm
/approve

I ran this locally and it worked great! Thank you so much for your efforts on this, really appreciate it!

deploy/charts/trust-manager/templates/clusterrole.yaml
```diff
@@ -43,3 +43,25 @@ rules:
   resources:
   - "events"
   verbs: ["create", "patch"]
+
+- apiGroups:
+  - ""
+  resources:
+  - "secrets"
+  verbs: ["create", "list", "watch"]
```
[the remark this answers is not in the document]
inteon: We still by default grant trust-manager access to read ALL secret resources.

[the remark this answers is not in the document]
erikgb: I think we must conditionally watch secrets from the controller if we remove it. And that might be the reason behind this bug.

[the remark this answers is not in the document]
erikgb: Here is a WIP PR to fix this: https://github.com/cert-manager/trust-manager/pull/207. My Tuesday (tomorrow) is filled with meetings, so feel free to pick it up. My gut feeling says that this new feature needs more work....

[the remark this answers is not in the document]
SgtCoDFish: Yeah that's my bad... thought I checked the permissions and clearly wasn't thorough / careful enough in my efforts to move this forwards. Sorry.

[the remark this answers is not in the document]
inteon: Thanks for confirming that this was not the intended behavior @SgtCoDFish.

PR #192, opened 2023-10-06
Update a couple of instances of the old project name

SgtCoDFish:
Spotted these while working on something unrelated and couldn't help but shave the yak!

PR #194, opened 2023-10-09
Vendor dependencies correctly

SgtCoDFish:
Make won't update files when a variable changes unless that variable is somehow included as a dependency for a target or in the path of the name of the target.

This commit adds versions to the paths of all tools we depend on, so that when one user bumps version numbers, everyone else won't need to run `make clean`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,7 +26,13 @@ OS     ?= $(shell go env GOOS)
 
 HELM_VERSION ?= 3.12.3
 KUBEBUILDER_TOOLS_VERISON ?= 1.28.0
+KUBECTL_VERSION ?= 1.28.2
+KIND_VERSION ?= $(shell grep "sigs.k8s.io/kind" go.mod | awk '{print $$NF}')
 GINKGO_VERSION ?= $(shell grep "github.com/onsi/ginkgo/v2" go.mod | awk '{print $$NF}')
+HELM_DOCS_VERSION ?= $(shell grep "github.com/norwoodj/helm-docs" hack/tools/go.mod | awk '{print $$NF}')
+BOILERSUITE_VERSION ?= $(shell grep "github.com/cert-manager/boilersuite" hack/tools/go.mod | awk '{print $$NF}')
+CONTROLLER_TOOLS_VERSION ?= $(shell grep "sigs.k8s.io/controller-tools" hack/tools/go.mod | awk '{print $$NF}')
+
 IMAGE_PLATFORMS ?= linux/amd64,linux/arm64,linux/arm/v7,linux/ppc64le
 
 RELEASE_VERSION ?= v0.7.0-alpha.1
@@ -61,8 +67,8 @@ unit-test:  ## runs unit tests, defined as any test which doesn't require extern
 	go test -v ./pkg/... ./cmd/...
 
 .PHONY: integration-test
-integration-test: depend  ## runs integration tests, defined as tests which require external setup (but not full end-to-end tests)
-	KUBEBUILDER_ASSETS=$(BINDIR)/kubebuilder/bin go test -v ./test/integration/...
+integration-test: | $(BINDIR)/kubebuilder-$(KUBEBUILDER_TOOLS_VERISON)/kube-apiserver $(BINDIR)/kubebuilder-$(KUBEBUILDER_TOOLS_VERISON)/etcd ## runs integration tests, defined as tests which require external setup (but not full end-to-end tests)
+	KUBEBUILDER_ASSETS=$(BINDIR)/kubebuilder-$(KUBEBUILDER_TOOLS_VERISON) go test -v ./test/integration/...
 
 .PHONY: lint
 lint: vet verify-boilerplate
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -26,7 +26,13 @@ OS     ?= $(shell go env GOOS)
 
 HELM_VERSION ?= 3.12.3
 KUBEBUILDER_TOOLS_VERISON ?= 1.28.0
+KUBECTL_VERSION ?= 1.28.2
```
SgtCoDFish, line 29: note: The kubectl version was previously updated dynamically by fetching the latest version from a remote URL. We should fix all dependencies so that for a given commit it's clear what versions of dependencies are tested together.

SgtCoDFish:
Follow up to #192

Makefile
```diff
@@ -26,7 +26,13 @@ OS     ?= $(shell go env GOOS)
...
+KIND_VERSION ?= $(shell grep "sigs.k8s.io/kind" go.mod | awk '{print $$NF}')
 GINKGO_VERSION ?= $(shell grep "github.com/onsi/ginkgo/v2" go.mod | awk '{print $$NF}')
+HELM_DOCS_VERSION ?= $(shell grep "github.com/norwoodj/helm-docs" hack/tools/go.mod | awk '{print $$NF}')
+BOILERSUITE_VERSION ?= $(shell grep "github.com/cert-manager/boilersuite" hack/tools/go.mod | awk '{print $$NF}')
+CONTROLLER_TOOLS_VERSION ?= $(shell grep "sigs.k8s.io/controller-tools" hack/tools/go.mod | awk '{print $$NF}')
```
erikgb, line 34: Is the go.mod in hack/tools really useful anymore? Could it make sense just to have the versions mastered from the Makefile, or is it used for something else?

I suspect this problem might be present in other cert-manager projects...

SgtCoDFish: I really strongly dislike hack/tools/go.mod for so many reasons:

- Introduces the risk of dependency conflicts between tools we depend on
- Builds tools with dependency versions from `hack/tools/go.mod` which might not be versions they were actually tested with
- Creates a confusing phantom module in the repo
- Noiser to update - updates can change dependencies / `go.sum`, as opposed to a single line version bump
- We still have to have another way of managing tools, because not all tools we use are go based - so it ends up being a second mechanism for tracking tools, not a unifying force

The upside is that `go get -u` makes it slightly easier to bump go-based tools to the latest version, but that upside isn't close to being worth it to me. `go install` is much better for this sort of thing.

I'd love to get rid of hack/tools everywhere, but because it's [in the golang/go wiki](https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module) I've not been able to successfully argue for its removal. (Obviously I disagree with the wiki on this)

If we could make the consensus be that we don't want to use it anywhere I'd gladly sign up for that and I think it would be an improvement across the board. But I also don't want to have to fight that fight to get this PR merged!

Makefile
```diff
@@ -194,49 +200,50 @@ $(BINDIR)/validate-trust-package: cmd/validate-trust-package/main.go pkg/fspkg/p
...
+	tar xfO $< kubebuilder/bin/etcd > $@ && chmod +x $@
+
+$(BINDIR)/kubebuilder-$(KUBEBUILDER_TOOLS_VERISON)/kube-apiserver: $(BINDIR)/kubebuilder-$(KUBEBUILDER_TOOLS_VERISON)/envtest-bins.tar.gz | $(BINDIR)/kubebuilder-$(KUBEBUILDER_TOOLS_VERISON)
+	tar xfO $< kubebuilder/bin/kube-apiserver > $@ && chmod +x $@
 
-$(BINDIR)/kubebuilder/bin/kube-apiserver: | $(BINDIR)/kubebuilder
-	curl -sSLo $(BINDIR)/envtest-bins.tar.gz "https://storage.googleapis.com/kubebuilder-tools/kubebuilder-tools-$(KUBEBUILDER_TOOLS_VERISON)-$(OS)-$(ARCH).tar.gz"
-	tar -C $(BINDIR)/kubebuilder --strip-components=1 -zvxf $(BINDIR)/envtest-bins.tar.gz
+$(BINDIR)/kubebuilder-$(KUBEBUILDER_TOOLS_VERISON)/envtest-bins.tar.gz: | $(BINDIR)/kubebuilder-$(KUBEBUILDER_TOOLS_VERISON)
+	curl -sSL -o $@ "https://storage.googleapis.com/kubebuilder-tools/kubebuilder-tools-$(KUBEBUILDER_TOOLS_VERISON)-$(OS)-$(ARCH).tar.gz"
```
erikgb, line 244: I think this might be possible to simplify by introducing https://pkg.go.dev/sigs.k8s.io/controller-runtime/tools/setup-envtest, but we could create a follow-up PR for that.

SgtCoDFish: I wasn't aware of that tool! Could be a good addition for sure - could save a bit of downloading in dev environments especially across our various projects!

PR #195, opened 2023-10-10
Make `Makefile`s reusable and automate release process

inteon:
ref: https://github.com/cert-manager/makefile-modules/issues/3

> [!Note]
> The trust-bundle adds some extra complexity 
> 1. The trust-bundle image uses a file in the "/debian-package" folder as copy-source, adding this file is not possible through ko and has to be done in an extra step. Luckily, we can reuse the layer-append logic that we have for the licenses module.
> 2. trust-bundles are versioned separately from trust-manager. However, when we do a trust-manager release, the Helm chart must refer to an existing trust-bundle. For that reason, the trust-manager release process starts with a `oci-maybe-push-package_debian` action to make sure the referenced trust package exists.
> 3. To simplify the release process, both trust-manager and the debian trust-bundle are published using GH actions now. This reduces the number of different systems (removes gcb) we have & pipeline secrets we have. It does however add extra changes to this PR.

ℹ️ I tested these changes on my personal branch and fixed all remaining issues in the release process.


CONTRIBUTING.md
```diff
@@ -1,64 +1,3 @@
...
-    person who certified (a), (b) or (c) and I have not modified
-    it.
-
-(d) I understand and agree that this project and the contribution
-    are public and that a record of the contribution (including all
-    personal information I submit with it, including my sign-off) is
-    maintained indefinitely and may be redistributed consistent with
-    this project or the open source license(s) involved.
-```
+Please refer to the [cert-manager Contributing guide](https://cert-manager.io/docs/contributing/).
```
SgtCoDFish, line 3: suggestion: I think the changes to this file would make a great standalone PR which we could merge really quickly!

make/debian-trust-package.mk
```diff
@@ -0,0 +1,46 @@
...
+# limitations under the License.
+
+debian_package_layer := $(bin_dir)/scratch/debian-trust-package
+debian_package_json := $(debian_package_layer)/debian-package/cert-manager-package-debian.json
+
+$(debian_package_layer)/debian-package $(bin_dir)/bin:
+	mkdir -p $@
+
+$(bin_dir)/bin/validate-trust-package: | $(bin_dir)/bin
+	go build -o $@ ./cmd/validate-trust-package
```
SgtCoDFish, line 22: suggestion: This target has dependencies which aren't listed here which will prevent it being rebuilt correctly.

Needs at least `cmd/validate-trust-package/main.go` and `pkg/fspkg/package.go`

SgtCoDFish (changes requested):
One thing I spotted!

I think it'd be nice and not too difficult to roll this change out in a way that doesn't require breaking the tests at any point, but I guess we can always roll the tests back so I don't think that's a requirement.

EDIT: That said, since the tests are failing right now maybe that would be a good idea so we can have some confidence before we merge

make/_shared/oci-build/image_tool/append_layers.go
```diff
@@ -0,0 +1,199 @@
...
+	"github.com/google/go-containerregistry/pkg/v1/mutate"
+	"github.com/google/go-containerregistry/pkg/v1/tarball"
+	"github.com/google/go-containerregistry/pkg/v1/types"
+	"github.com/spf13/cobra"
+)
+
+var CommandAppendLayers = cobra.Command{
+	Use:   "append-layers oci-path [path-to-tarball...]",
+	Short: "Appends a tarball or directory to every image in an OCI index.",
+	Args:  cobra.MinimumNArgs(1),
```
SgtCoDFish, line 40: suggestion (non-blocking): Not something that should change for this PR although it's an important thing: I think changing files in-place like this is a bit scary with make (and generally) and it might be safer to write a temporary file first and then do the modification to that temporary file and save it with a new name.

The current flow means that a tool outside of make can't ever safely read/use the tarball on disk as-is because it can't know if it's in the process of being changed or not. It either has to check that make isn't running (and ensure it doesn't start running before reading the file) or else check the contents of the file before using it to ensure it was correctly updated.

This would imply ko building a temporary tarball (e.g. `$(oci_layout_path_x).beforeappend`) and then the append-layers script writing to `$(oci_layout_path_x)`

make/debian-trust-package.mk
inteon: Fixed

inteon:
@SgtCoDFish I re-added the old make targets for backwards compatibility, now we don't have to break the tests.

make/_shared/oci-build/image_tool/append_layers.go
inteon: The tool actually changes the `oci-path` which is a folder structure representing a multi-arch OCI image.
The `convert_to_docker_tar.go` file actually takes the `oci-path` folder as input and generates a tarball for the local architecture.
I agree that there is some work-to-be-done to simplify this area. There is also some hope in me still that some of these workarounds will be deprecated in the future in favor of upstream improvements.

SgtCoDFish: Ah fair enough! Thanks for clarifying

SgtCoDFish (approved):
/lgtm
/approve
/hold

I think this is good to go and kudos to @inteon for the effort it took to get this done.

I'm adding a hold because Tim isn't working today as far as I can tell (despite the updates he made this morning!). This can wait until Monday to merge, and then Tim should be available to help with the followup PRs to cleanup the testing repo / remove the deprecated make targets afterwards.

make/debian-trust-package-fetch.sh
```diff
@@ -0,0 +1,143 @@
...
+installed_version=$(cat "$TMP_DIR/version.txt")
+version_suffix=".0"
+
+if [[ "$ACTION" == "latest" && "$installed_version" == "$target_ca_certificates_version" ]]; then
+	version_suffix=".${TARGET_DEBIAN_BUNDLE_VERSION##*.}"
+	echo "+++ installed version matches target version; reusing the current suffix '$version_suffix'"
+fi
+
+echo "{}" | jq \
+	--rawfile bundle /etc/ssl/certs/ca-certificates.crt \
```
SgtCoDFish, line 137: note: IMPORTANT

I missed this during the review. I shouldn't have missed it and it'll need to fixed.

This uses the _system's_ trust-bundle, not the one from the container.

It breaks development on all platforms which aren't Debian-based linux (e.g. macOS, RHEL, etc)

It shouldn't lead to an incorrect bundle being pushed upstream, but it needs fixing as a matter of urgency in case that changes in the future.

See https://github.com/cert-manager/trust-manager/pull/355

PR #197, opened 2023-10-11
Enable dependabot updates

inteon:
Enable dependabot & make it bundle all upgrades in a single PR. (example: https://github.com/cert-manager/approver-policy/blob/main/.github/dependabot.yaml)


SgtCoDFish (approved):
/lgtm
/approve

Makes sense, cheers!

PR #200, opened 2023-10-18
Fix build on macOS / values.yaml wording tweaks 

SgtCoDFish:
I fully accept that the `provision-buildx` stuff is weird and we should change it or document it better - but this seems to work for now, and I can run the build commands succesfully locally while they still (usually 😂 ) pass in CI


SgtCoDFish:
The helm values.yaml tweaks are a followup to #193

inteon (commented):
/lgtm
/approve
Thanks Ashley!

PR #201, opened 2023-10-19
add a CN for the trust-manager certificate

SgtCoDFish:
since this cert is self-signed, its subject will match its issuer

it's valid to have an empty subject, but not an empty issuer


erikgb (approved):
Thanks, Ash!

/lgtm

PR #204, opened 2023-10-20
Upgrade go to 1.21

inteon:
Upgrade all go versions & go.mod files to 1.21

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/cert-manager/trust-manager
 
-go 1.20
+go 1.21.3
 
 require (
 	github.com/go-logr/logr v1.2.4
--- a/hack/tools/go.mod
+++ b/hack/tools/go.mod
@@ -1,6 +1,6 @@
 module github.com/cert-manager/tools/hack/tools
 
-go 1.20
+go 1.21.3
 
 require (
 	github.com/cert-manager/boilersuite v0.1.0
--- a/trust-packages/debian/go.mod
+++ b/trust-packages/debian/go.mod
@@ -1,3 +1,3 @@
 module github.com/cert-manager/trust-manager/debian-bundle-static
 
-go 1.19
+go 1.21.3
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 # Build the trust binary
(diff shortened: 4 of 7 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/cert-manager/trust-manager
 
-go 1.20
+go 1.21.1
```
SgtCoDFish, line 3: suggestion: this feels wrong? If we're giving a specific patch release it should surely be 1.21.3 (latest) - but I'd think we should just use `1.21` instead, right?

hack/tools/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/cert-manager/tools/hack/tools
 
-go 1.20
+go 1.21.1
```
SgtCoDFish, line 3: suggestion: as above!

trust-packages/debian/go.mod
```diff
@@ -1,3 +1,3 @@
 module github.com/cert-manager/trust-manager/debian-bundle-static
 
-go 1.19
+go 1.21.1
```
SgtCoDFish, line 3: suggestion: as above!

SgtCoDFish (changes requested):

go.mod
erikgb, line 3: Any particular reason for selecting 1.21.1? Why not 1.21.0 (first 1.21.x) or 1.21.3 (current latest).

erikgb: I think you cannot set just 1.21 anymore.

inteon: Ah, sorry; this was quite an old change and 1.21.1 was the latest version at that point in time. I updated the PR to 1.21.3.

inteon: I updated it to 1.21.3

erikgb (approved):
/lgtm

But we might consider requiring just 1.21.0 in the go.mod files. Do we plan to bump the requirement as new Go releases are released?

inteon:
I don't think we have to bump it every time, I just did it now because in 1.21 the go.sum files contain more entries (I think they were missing in 1.20)

SgtCoDFish:
Closing the loop: https://go.dev/ref/mod#go-mod-file-go

The go directive is specifying a minimum. The fact that it's an older version here isn't a problem really!

PR #210, opened 2023-10-25
Cleanup patch functions Secret and ConfigMap targets

inteon:
Renamed some functions and removed some unnecessary type casting.

### DIFF
--- a/pkg/bundle/sync.go
+++ b/pkg/bundle/sync.go
@@ -313,7 +313,7 @@ func (b *bundle) syncConfigMapTarget(
 			)
 
 		if err = b.patchConfigMapResource(ctx, configMapPatch); err != nil {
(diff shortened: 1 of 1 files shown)


pkg/bundle/sync.go
```diff
@@ -312,7 +312,7 @@ func (b *bundle) syncConfigMapTarget(
 					WithController(true),
 			)
 
-		if err = b.patchResource(ctx, configMapPatch); err != nil {
+		if err = b.patchConfigMapResource(ctx, configMapPatch); err != nil {
 			return false, fmt.Errorf("failed to patch configMap %s/%s: %w", namespace, bundle.Name, err)
```
erikgb, line 316:
```suggestion
			return false, fmt.Errorf("failed to patch ConfigMap %s/%s: %w", namespace, bundle.Name, err)
```

For consistency.

PR #212, opened 2023-10-25
BUGFIX: fix bugs in validation logic for secret target & add tests

inteon:


erikgb (approved):
/lgtm

Thanks! We should probably try to DRY this up in a follow-up PR. There is currently quite a lot of semi-duplication which makes things hard to read/understand.

Issue #175, opened 2023-09-07
support extra annotations on resoures in helm chart


thesuperzapper:
I would like to be able to set "global" annotations for all resources in the trust-manager helm chart.

The main purpose of this is to allow a single "combined" chart of `cert-manager` and `trust-manager` to be applied successfully using Argo CD. Right now, if a combined chart is created, the first sync will always fail (because `trust-manager` depends on `cert-manager` being ready). 

However, if I could set annotations on all resources, I would be able to make use of [Argo CD's Sync Waves](https://argo-cd.readthedocs.io/en/stable/user-guide/sync-waves/#how-do-i-configure-waves) to put all the `trust-manager` resources in a wave that happens after `cert-manager`. 

For example, we might add a value called `app.annotations` to the chart (which adds no annotations by default), and use it like follows:

```yaml
app:
  annotations:
    ## assuming cert-manager has a lower wave (or no wave) set
    argocd.argoproj.io/sync-wave: "5"
```

thesuperzapper:
I think this issue is still important, and is causing me to run a forked version of the chart.

PR #214, opened 2023-10-26
Add helm values for annotations

thesuperzapper:
closes https://github.com/cert-manager/trust-manager/issues/175

This PR adds two new values to the helm chart:

- `app.deploymentAnnotations`: annotations to be applied to the trust-manager `Deployment` resource
- `app.certificateAnnotations`: annotations to be applied to the `Certificate` and `Issuer` resources

This is primarily so that users can create a combined chart with `cert-manager`, and use [ArgoCD ](https://argo-cd.readthedocs.io/en/stable/user-guide/sync-waves/#how-do-i-configure-waves) to prevent race conditions.

For example, here are the values I am using in [deployKF](https://www.deploykf.org/) for this purpose:

```yaml
app:
  certificateAnnotations:
    ## we must sync cert-manager resources after cert-manager itself
    argocd.argoproj.io/sync-wave: "10"

  deploymentAnnotations:
    ## we must sync the trust-manager deployment after its certificate is ready
    argocd.argoproj.io/sync-wave: "20"
```

### DIFF
--- a/deploy/charts/trust-manager/README.md
+++ b/deploy/charts/trust-manager/README.md
@@ -34,6 +34,8 @@ Kubernetes: `>= 1.25.0-0`
 | app.readinessProbe.path | string | `"/readyz"` | Path on which to expose trust-manager HTTP readiness probe using default network interface. |
 | app.readinessProbe.port | int | `6060` | Container port on which to expose trust-manager HTTP readiness probe using default network interface. |
 | app.securityContext.seccompProfileEnabled | bool | `true` | If false, disables the default seccomp profile, which might be required to run on certain platforms |
+| app.deploymentAnnotations | object | `{}` | Annotations for the trust-manager Deployment |
+| app.certificateAnnotations | object | `{}` | Annotations for the Certificate/Issuer resources created by trust-manager |
 | app.trust.namespace | string | `"cert-manager"` | Namespace used as trust source. Note that the namespace _must_ exist before installing trust-manager. |
 | app.webhook.host | string | `"0.0.0.0"` | Host that the webhook listens on. |
 | app.webhook.hostNetwork | bool | `false` | Specifies if the app should be started in hostNetwork mode. Required for use in some managed kubernetes clusters (such as AWS EKS) with custom CNI. |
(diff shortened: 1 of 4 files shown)


thesuperzapper:
@inteon I would appreciate a quick look at this small change to the helm chart, as it will let me stop forking the `trust-manager` helm chart for use in deployKF.

erikgb:
@thesuperzapper Have you seen https://github.com/cert-manager/trust-manager/pull/157? Being able to cut the dependency to cert-manager from trust-manager makes a lot of sense.

We should probably include the feature of adding annotations anyway, but do you really need to annotate the trust-manager deployment to solve your use case? Since the deployment mount the certificate secret as a volume mount, I don't think the pods will be scheduled before the secret is created by cert-manager.

/ok-to-test
/lgtm
/cc @inteon @SgtCoDFish

erikgb:
> Please run 'make -s update-helm-docs'

To fix CI

SgtCoDFish:
Could you provide a little more detail on why you want this? We discussed it a bit and we guessed that with a combined chart you're running into issues with the CRDs not being installed when the trust-manager `Certificate` is created?

thesuperzapper:
> Could you provide a little more detail on why you want this? We discussed it a bit and we guessed that with a combined chart you're running into issues with the CRDs not being installed when the trust-manager `Certificate` is created?

@SgtCoDFish the issue is not about the CRDs themselves being installed, it's about race conditions between the four "phases" of a combined `cert-manager` + `trust-manager` chart:

1. `cert-manager` 
2. Cert Manager Certificates (depend on `cert-manager` webhook)
3. `trust-manager` (mounts certificate secrets, for its webhook)
4. Trust Manager Bundles (depend on `trust-manager` webhook)

ArgoCD supports the idea of ["sync waves"](https://argo-cd.readthedocs.io/en/stable/user-guide/sync-waves/#how-do-i-configure-waves) which allows us to resolve this race condition by appropriately annotating the resources that depend on each other into separate waves.

SgtCoDFish (changes requested):
Thanks for explaining! I've gotta say I'm kinda uneasy about merging this for what feels like such a specific Argo thing - but I think that's probably because I'm just generally kinda uneasy about Helm 😂 

I've left a comment - do you think you could make that change? 

(Note I'm finishing for the day now, so I might not be able to re-review until tomorrow)

thesuperzapper:
I think this issue is still important, and is causing me to run a forked version of the chart.

PR #198, opened 2023-10-15
allow setting namespace for helm chart

vinny-sabatini:
This is based on cert-manager's helm chart override and resolves #145

PR #215, opened 2023-10-27
Use proper namespace for webhook

joemccall86:
# Context

Our releases were failing to deploy because it was trying to call a webhook that didn't exist. Our installation had the helm package in one namespace, but the resources for that package in another namespace. The result was something like this:

```
Error: UPGRADE FAILED: failed to create resource: Internal error occurred: failed calling webhook "[trust.cert-manager.io](http://trust.cert-manager.io/)": failed to call webhook: Post "[https://trust-manager.helm-charts.svc:443/validate-trust-cert-manager-io-v1alpha1-bundle?timeout=5s](https://trust-manager.helm-charts.svc/validate-trust-cert-manager-io-v1alpha1-bundle?timeout=5s)": service "trust-manager" not found
```

The service actually exists in the `trust-manager` namespace, not the `helm-charts` namespace.

After examining the code it would appear that this is the only place where the namespace is hard-coded to the release namespace. Everywhere else uses the helper function "trust-manager.namespace".

# Fix

Use "trust-manager.namespace" function when creating the validating webhook.

### DIFF
--- a/deploy/charts/trust-manager/templates/webhook.yaml
+++ b/deploy/charts/trust-manager/templates/webhook.yaml
@@ -48,5 +48,5 @@ webhooks:
     clientConfig:
       service:
         name: {{ include "trust-manager.name" . }}
-        namespace: {{ .Release.Namespace | quote }}
+        namespace: {{ include "trust-manager.namespace" . }}
         path: /validate-trust-cert-manager-io-v1alpha1-bundle


erikgb:
Thanks @joemccall86! This looks like a nice bugfix! I will review shortly.

/ok-to-test

erikgb:
This seems like an oversight from https://github.com/cert-manager/trust-manager/pull/198. Thanks!

/lgtm
/cc @inteon

erikgb:
To verify if I got more permissions now, I can try to

/approve

PR #217, opened 2023-10-29
chore: generate applyconfigurations for custom resources

erikgb:
I think the migration to SSA in this project made the code harder to read/understand. 😉 While the support for SSA in the controller-runtime client is limited at present, ref. https://github.com/kubernetes-sigs/controller-runtime/issues/347, I am convinced that we can do better.

Generating applyconfigurations with controller-gen is in the making, ref. https://github.com/kubernetes-sigs/controller-tools/pull/818 and https://github.com/kubernetes-sigs/controller-tools/pull/536, but it's a complex task that still needs some work.

In the meantime, I suggest generating applyconfigurations using the upstream generator that we can use in our controllers. I also suggest exporting the generated applyconfiguration types, even if I think client types should be available without depending on the operator module - creating a potential dependency-hell. This has been a long-standing issue in cert-manager that I hope can get some progress soon.

/cc @inteon @SgtCoDFish


SgtCoDFish (approved):
/lgtm
/approve

This doesn't seem to runtime impact currently - making it very easy to approve. Thanks for looking into this, I'm interested in following it!

Issue #224, opened 2023-11-07
Custom trust namespace - permissions issue


tspearconquest:
Hello,

I've configured trust-manager as a subchart in order to install it alongside my cert-manager deployment, and I added the below values:

```
app:
  namespace: trust-manager
  trust:
    namespace: trust
```

When installing trust-manager, the role and rolebinding for trust-manager to manage its leases is configured against the ` trust` namespace and trust-manager uses the `trust` namespace as per the `--trust-namespace` flag.

However, as per the documentation:

> An ideal deployment would be a fresh namespace dedicated entirely to trust-manager, to minimize the number of actors in your cluster that can modify your trust sources.

To me, this means that the leases would be kept in the namespace where trust manager is installed, and the "trust" namespace will only contain secrets to be read by trust-manager for creating a bundle.

I've created a patch for the helm chart to setup permissions for the leases separately so that trust-manager will be granted privileges to manage the leases in its namespace instead of the trust namespace, but in order to make this patch effective, we also need for trust-manager to not use the trust namespace for its leases, and instead either read the POD_NAMESPACE variable from the environment to set the namespace for the leases, or provide a separate flag to specify the trust-manager namespace to the binary to be able to properly set the namespace for the leases.

tspearconquest:
#225

erikgb:
@tspearconquest Thanks for this issue. This seems like a bug. I would also expect the leader election lease to be located in the same namespace as trust-manager is running, and not the trust namespace. But I think the fix should be quite simple. See: https://github.com/kubernetes-sigs/controller-runtime/blob/9f9f369134d82816648b691c168008c2d0aa9bfe/pkg/leaderelection/leader_election.go#L74-L80.

I think we just have to make a simple change in Go code. Just delete this line: https://github.com/cert-manager/trust-manager/blob/cb1bef1552a8d175e61c11793832661869e3c5fe/cmd/trust-manager/app/app.go#L81

But we need to configure RBAC correctly.

tspearconquest:
#225 should resolve the RBAC issue because it separates the lease permission to its own role and rolebinding, and they are both pointed at the deployment namespace rather than the trust namespace. I will submit a separate MR for the deletion of line 81 in leader_election.go.

tspearconquest:
https://github.com/cert-manager/trust-manager/pull/226

erikgb:
I think the change should go in the same PR. RBAC and controller behavior must be in sync.

[...]

PR #225, opened 2023-11-07
Allow permissions to put the leases in the trust-manager namespace, not the trust namespace

tspearconquest:
This pull request separates the lease permissions to their own role and role binding in the namespace where trust-manager will be installed. See #224 for more details.

### DIFF
--- a/deploy/charts/trust-manager/templates/role.yaml
+++ b/deploy/charts/trust-manager/templates/role.yaml
@@ -14,6 +14,14 @@ rules:
   - "get"
   - "list"
   - "watch"
+---
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: {{ include "trust-manager.name" . }}
+  namespace: {{ include "trust-manager.namespace" . }}
+  labels:
+{{ include "trust-manager.labels" . | indent 4 }}
+rules:
 - apiGroups:
   - "coordination.k8s.io"
   resources:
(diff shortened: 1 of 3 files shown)


deploy/charts/trust-manager/templates/role.yaml
```diff
@@ -14,6 +14,14 @@ rules:
   - "get"
   - "list"
   - "watch"
+---
+apiVersion: rbac.authorization.k8s.io/v1
```
erikgb, line 18:
```suggestion
kind: Role
apiVersion: rbac.authorization.k8s.io/v1
```

erikgb (changes requested):
Thanks @tspearconquest. I think we should do this, but it probably deserves special attention in the release notes for our next release. As we change namespace of the leader election lease for some users, that might allow multiple trust-manager controllers to run for a short while.

tspearconquest:
Sounds good. Is there anything I need to do in order to mention this in the release notes, or will it be taken care of in the release?

erikgb:
/cc @wallrj @SgtCoDFish @inteon 

I am tempted to approve this as well, but since my approver status is very fresh, I don't want to do anything wrong. 👼

SgtCoDFish (approved):
/lgtm
/approve

Thanks for this - it makes sense to me. Really appreciate your involvement 😁

Issue #199, opened 2023-10-17
Support of setting arbitrary password for PKCS12 truststore


gsustek:
We are using cert-manager to issue certificates to various-servers running in our Kubernetes cluster. Currently we use  truststore.p12 in the placed in kubernetes Secret to establish trust to the client java applications (sTLS). 
Our java applications are running on java11 and [java17](https://bugs.openjdk.org/browse/JDK-8275252) which does not supports password-less pkcs12 container as JVM argument.

To fix this, we were hoping to use this cert-manager component to distribute the trust. But since a all of our clients are Java-based, we need to be able to provide the trust bundle in password protected PKCS12 format. 

_-Djavax.net.ssl.trustStore=/configuration/p12/truststore/trustStore.p12 -Djavax.net.ssl.trustStorePassword=xxxxxx_

Would it be possible to add an opt-in to provide the arbitrary password to protected PKCS12?

erikgb:
Oh, doesn't even Java 17 support password-less PKCS#12 truststores? Yes, then I think it makes sense to add a controller config to optionally set a password for all PKCS#12 truststores created by trust-manager. Note: I suggest making this a configuration of the controller globally for the cluster, and NOT per Bundle. Please submit a PR and assign me for review.

arsenalzp:
This issue is reasonable.
trust-manager exports bundles in PEM format by default.
Do we have to add feature to export bundles in PKCS12 format optionally, with arbitrary password option as well?

erikgb:
Opt-in to export in JKS/PKCS12 is there, but with hard-coded password (JKS) and password-less (PKCS12).

arsenalzp:
Ah, I found ’AdditionalFormats’ structure.

arsenalzp:
Hello,
I'm thinking about the best way providing arbitrary password to the _trust-manager_ controller.
As it cluster scoped, it's possible to set it by using _trust-manager_ CLI arguments or ether by using environment variables for _trust-manager_ Deployment resource.
Of course we can user _Secret_ resource which is retrieved by _trust-manager_ controller, however this way adding some complication to the controller logic; this way is more secure.

erikgb:
> Of course we can user Secret resource which is retrieved by trust-manager controller, however this way adding some complication to the controller logic; this way is more secure.

Passwords for truststores are not about security, but integrity. And all consumers of the truststore need to know the agreed password. So I would not consider it a secret. The truststore contains only public information.

[...]

PR #229, opened 2023-11-09
Add pkcs12 password feature

arsenalzp:
This PR is related to the issue #199 
It add PKCS12 arbitrary password feature which allows to set a password for PKCS12 bundle as well as don't remove password-less PKCS12 feature.


erikgb (changes requested):
@arsenalzp This looks like a good start! I have added some comments after my first review pass. Some of them are just suggestions - so feel free to disagree. But please explain why. 😉 

/ok-to-test

deploy/charts/trust-manager/values.yaml
```diff
@@ -50,6 +50,8 @@ defaultPackageImage:
 app:
   # -- Verbosity of trust-manager logging; takes a value from 1-5, with higher being more verbose
   logLevel: 1
+  # -- Password for JKS and PKCS12 bundles pretection
+  pkcs12Password: defaultPKCS12Passwd
```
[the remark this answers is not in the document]
arsenalzp: Hello,
Thank you for you feedback, this PR is gonna be the 2nd one for this project!
Yes, I agree with you, I change it, however related tests were added, so there are two tests now: password-less and arbitrary password for both Secret and ConfiMap resources.

pkg/bundle/sync.go
```diff
@@ -244,12 +249,10 @@ func certAlias(derData []byte, friendlyName string) string {
 	return certHash[:8] + "|" + friendlyName
 }
 
-type pkcs12Encoder struct {
-	password string
-}
+func (e pkcs12Encoder) encode() ([]byte, error) {
+	var encoder = pkcs12.Modern2023 // user moder PKCS#12 encoder
```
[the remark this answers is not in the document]
arsenalzp: Default PKCS12 encoder structure represent legacy algorithms, so it's not possible to use encrypted pkcs12 by using openssl 3, for example.  However it was possible to use password-less pkcs12 with modern openssl 3.
Please take into account that pkcs12.EncodeTrustStoreEntries is marked as deprecated.
So, based on the documentation it is recommended to use modern encoder, which provides the following algorithms: 
```
var Modern2023 = &Encoder{
	macAlgorithm:         oidSHA256,
	certAlgorithm:        oidPBES2,
	keyAlgorithm:         oidPBES2,
	macIterations:        2048,
	encryptionIterations: 2048,
	saltLen:              16,
        rand:                 rand.Reader,
}
```

pkg/bundle/sync.go
```diff
@@ -482,18 +486,20 @@ func (b *bundle) syncSecretTarget(
 	return true, nil
 }
 
-func populateAdditionalFormatData(data string, target trustapi.BundleTarget, targetMap map[string][]byte) error {
+// Populates additional data formats which were defined in Bundle resource: JKS and/or PKCS12.
+// Both JSK and PKCS12 containers might be encrypted by password.
+func populateAdditionalFormatData(bundleData string, bundlePasswd string, target trustapi.BundleTarget, targetMap map[string][]byte) error {
```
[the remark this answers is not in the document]
arsenalzp: I was thinking to re-write it to use encoder interface and and provide all necessary arguments in the related structure then call .encode() method based on the casting results: either PKCS12 type or JKS, however it makes code more a little bit complicated due to casting.
Nevertheless, I'll think out about that, let's approve the basis of the solution first.

pkg/bundle/sync.go
```diff
@@ -64,6 +63,16 @@ const (
 	fieldManager             = "trust-manager"
 )
 
+type pkcs12Encoder struct {
+	bundlePasswd string
```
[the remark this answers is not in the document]
arsenalzp: Agree, I will change it.

pkg/bundle/sync.go
```diff
@@ -64,6 +63,16 @@ const (
...
+	bundleData   string
+}
+
+type jksEncoder struct {
+	bundlePasswd []byte
```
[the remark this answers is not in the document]
arsenalzp: s.Store() accept second argument  (password) as []byte, so I wanted to avoid type conversion, of course it is not consume much CPU/Memory resources in our case, despite of it is running in container.

deploy/charts/trust-manager/values.yaml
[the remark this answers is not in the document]
erikgb: >  related tests were added, so there are two tests now: password-less and arbitrary password for both Secret and ConfiMap resources.

Ahhh, my review comment was not good - it could easily be misinterpreted. What I meant, is that to get the Helm Deployment template correct, you probably have to try a few rounds of local rendering with different Helm values.

pkg/bundle/sync.go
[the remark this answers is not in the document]
erikgb: I don't think we need to do a refactoring to interfaces - at least not now. :-)

pkg/bundle/sync.go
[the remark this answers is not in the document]
erikgb: Ok, makes sense. I would prefer readability of code in this case.

arsenalzp:
Hello,
Your remarks were fixed.
Could you please so kind to review PR again?

deploy/charts/trust-manager/README.md
```diff
@@ -31,6 +31,7 @@ Kubernetes: `>= 1.25.0-0`
 | app.metrics.service.enabled | bool | `true` | Create a Service resource to expose metrics endpoint. |
 | app.metrics.service.servicemonitor | object | `{"enabled":false,"interval":"10s","labels":{},"prometheusInstance":"default","scrapeTimeout":"5s"}` | ServiceMonitor resource for this Service. |
 | app.metrics.service.type | string | `"ClusterIP"` | Service type to expose metrics. |
+| app.pkcs12Password | string | `nil` | Password for JKS and PKCS12 bundles pretection |
```
inteon, line 34:
```suggestion
| app.pkcs12Password | string | `nil` | Password for JKS and PKCS12 bundles protection |
```

pkg/bundle/sync.go
```diff
@@ -173,15 +185,11 @@ func (b *bundle) secretBundle(ctx context.Context, ref *trustapi.SourceObjectKey
...
-type jksEncoder struct {
-	password []byte
-}
-
 // encodeJKS creates a binary JKS file from the given PEM-encoded trust bundle and password.
 // Note that the password is not treated securely; JKS files generally seem to expect a password
 // to exist and so we have the option for one.
-func (e jksEncoder) encode(trustBundle string) ([]byte, error) {
-	cas, err := util.DecodeX509CertificateChainBytes([]byte(trustBundle))
+func (e jksEncoder) encode() ([]byte, error) {
```
inteon, line 191: Why are these signatures being changed?

cmd/trust-manager/app/options/options.go
```diff
@@ -156,6 +156,10 @@ func (o *Options) addBundleFlags(fs *pflag.FlagSet) {
 	fs.BoolVar(&o.Bundle.SecretTargetsEnabled,
 		"secret-targets-enabled", false,
 		"Controls if secret targets are enabled in the Bundle API.")
+
+	fs.StringVar(&o.Bundle.PKCS12Password,
```
inteon, line 160: @erikgb if this option is set at the controller-level, changing this value will cause all the newly reconciled bundles to have a password (not the existing ones). Only when an existing store is changed, the password will be set too.

inteon: I would vote for an additional output-format (eg. PKCS12WithPassword)

erikgb: > if this option is set at the controller-level, changing this value will cause all the newly reconciled bundles to have a password (not the existing ones). Only when an existing store is changed, the password will be set too.

Why? We could make it happen. You get events for all watched resource types on startup.

pkg/bundle/sync.go
arsenalzp: Hello,
Thank you for a review.
_jksEncoder_ and _pkcs12Encoder_ structures bring all necessary information in corresponding properties:
```
type encoderData struct {
	storePasswd string
	bundleData  string
}

type pkcs12Encoder struct {
	encoderData
}

type jksEncoder struct {
	encoderData
}
```
In the future it might be reasonable to rewrite functions like _populateAdditionalFormatData_ to either accept encoder interface and based on the casting result invoke corresponding `encode()` method or use generics, something like that.

cmd/trust-manager/app/options/options.go
inteon: Could it be clarified in the flag description that this will change the passwords of all existing PKCS12 stores (and thus break existing setups)?
Also, let's fix any issues with mixed passwords in this PR.

erikgb: > let's fix any issues with mixed passwords in this PR.

Which mix?

inteon: If you set the password to "password1" and change it to "password2" later, some stores will still have the "password1" and only newly reconciled stores will have "password2".

arsenalzp: > If you set the password to "password1" and change it to "password2" later, some stores will still have the "password1" and only newly reconciled stores will have "password2".

You can't update password without re-starting controller pod, once restarted, all stores will be reassembled with new password reflection.

inteon: Did you test that all stores were updated?

arsenalzp: > Did you test that all stores were updated?

Yes, I did.
I've just checked once again by changing related _trust-manager_ Deployment resource property.

deploy/charts/trust-manager/README.md
```diff
@@ -31,6 +31,7 @@ Kubernetes: `>= 1.25.0-0`
 | app.metrics.service.enabled | bool | `true` | Create a Service resource to expose metrics endpoint. |
 | app.metrics.service.servicemonitor | object | `{"enabled":false,"interval":"10s","labels":{},"prometheusInstance":"default","scrapeTimeout":"5s"}` | ServiceMonitor resource for this Service. |
 | app.metrics.service.type | string | `"ClusterIP"` | Service type to expose metrics. |
+| app.pkcs12Password | string | `nil` | Password for JKS and PKCS12 stores protection |
```
erikgb, line 34: I think these docs are a bit misleading. The default should be `""` (empty string) creating a password-less PKC12 truststore.

pkg/bundle/sync.go
erikgb: I also don't think you need to change the signature of this function. That makes a struct instance locked to the bundle data. Why do you need to change it?

deploy/charts/trust-manager/README.md
arsenalzp: > I think these docs are a bit misleading. The default should be `""` (empty string) creating a password-less PKC12 truststore.

I guess it makes code more clear and avoid of copying bundle data between function/method. 
The bundle data is tightly coupled to its encoder; imagine you would like to add `decode()`, `transform()` methods in the future, then it leads that the bundle data should be returned from one encoder method and immediately consumed by another one.

arsenalzp:
I guess it makes code more clear and avoid of copying bundle data between function/method.
The bundle data is tightly coupled to its encoder; imagine you would like to add decode(), transform() methods in the future, then it leads that the bundle data should be returned from one encoder method and immediately consumed by another one.

deploy/charts/trust-manager/README.md
arsenalzp: I agree, it will be changed

pkg/bundle/sync.go
arsenalzp: I guess it makes code more clear and avoid of copying bundle data between function/method.
The bundle data is tightly coupled to its encoder; imagine you would like to add decode(), transform() methods in the future, then it leads that the bundle data should be returned from one encoder method and immediately consumed by another one.
Let's look how is the bundle data was being used before: put to populate function just only to be put to corresponding encoder's method again.

erikgb: That's pretty common in decorators. When will you use function parameters with your view on this?

arsenalzp: I agree with you, but at this moment encoder structure consists of 2 fields: password one and data one, what if we need more parameters for encoder, more methods. Ephemeral example - push bundle data into some destination, then encode, so that data should be put as argument twice: to some ’push(data)’ as well as to ’encode(data)’ functions. 
My proposal look pretty good, we operate by encoder structure, it might be pushed anyway, with all necessary parameters. Cool!

arsenalzp: Colleagues, if you insist to use bundle data as argument to `encode` method, I will rewrite, of course.

arsenalzp:
Hello colleagues,
Sorry that I'm disturbing you again.
Any decision regarding this PR? Should I change/fix something?

erikgb:
> Should I change/fix something?

Fields or function params is not a blocker for me. I prefer function params, but. Maybe you can fix (and close) the other open discussions? So we are left with just that point to land?

SgtCoDFish:
This is a drive-by comment and I don't have time to fully review but I think I'm negative on adding this as a command-line flag.

This maps neatly to being added in the resource, which is where I'd expect to be able to set it:

```json
{
  "apiVersion": "trust.cert-manager.io/v1alpha1",
  "kind": "Bundle",
  "metadata": {
    "name": "testing"
  },
  "spec": {
    "sources": [
      {
        "useDefaultCAs": true
      }
    ],
    "target": {
      "additionalFormats": {
        "jks": {
          "key": "my-bundle.jks",
          "password": "super_secret",
        }
      },
      "configMap": {
        "key": "mybundle.pem"
      }
    }
  }
}
```

Putting myself in the shoes of a user, I'd be familar with the concept that each JKS file can have a different password. I'd expect trust-manager to work the same way as that.

Obviously today we have one global setting for the password, but that's just because the JKS files need _some_ password, not because the setting should be global to all bundles.

I can't really see a reason for this to be a CLI flag.

arsenalzp:
> This is a drive-by comment and I don't have time to fully review but I think I'm negative on adding this as a command-line flag.
> 
> This maps neatly to being added in the resource, which is where I'd expect to be able to set it:
> 
> ```json
> {
>   "apiVersion": "trust.cert-manager.io/v1alpha1",
>   "kind": "Bundle",
>   "metadata": {
>     "name": "testing"
>   },
>   "spec": {
>     "sources": [
>       {
>         "useDefaultCAs": true
>       }
>     ],
>     "target": {
>       "additionalFormats": {
>         "jks": {
>           "key": "my-bundle.jks",
>           "password": "super_secret",
>         }
>       },
>       "configMap": {
>         "key": "mybundle.pem"
>       }
>     }
>   }
> }
> ```
> 
> Putting myself in the shoes of a user, I'd be familar with the concept that each JKS file can have a different password. I'd expect trust-manager to work the same way as that.
> 
> Obviously today we have one global setting for the password, but that's just because the JKS files need _some_ password, not because the setting should be global to all bundles.
> 
> I can't really see a reason for this to be a CLI flag.

Hello,
It was proposed to do it globally for entire cluster.
However, if someone has different point of view, let's discuss this topic; it is no so hard to rewrite this solution.

arsenalzp:
Another PR #233  related to @SgtCoDFish was submitted.

Issue #100, opened 2023-01-19
Modifying Bundle target can result in CA certs not being available for a while


irbekrm:
At the moment if  `bundle.spec.target` gets modified, trust manager will apply two updates in separate reconciles- [one where the old CA data is deleted from the target configmaps](https://github.com/cert-manager/trust-manager/blob/main/pkg/bundle/bundle.go#L116-L152), another where the new CA data is written to the target configmaps. This will happen even when the CA data does not need to change, for example if namespace selector is removed.

In practice this means that there is a period of time where the CA data is not available and in some cases it could cause issues:
- if there is a latency for calls to kube apiserver either because there is a large number of targets that trust-manager needs to update or because the apiserver is busy, the period when the CA cert is not available will be longer
- if user applies a change to the Bundle that both changes the target and also references a source that cannot be used (is an invalid CA cert), the old CA data will be removed and the configmaps will remain empty till the user fixes the issue

Assuming that we think that CA data should be 'always' available, I think it would be preferable if target updates were atomic and we didn't have a period of time where the target configmaps are empty.

I ran an experiment with 1000 namespaces and observed that the after a target namespace selector change a target configmap had its CA cert deleted and remained empty for 5 minutes till the same CA data got added back:
```
irbe@trust-manager$ k logs deploy/trust-manager -ncert-manager | grep test-1000
I0119 13:24:53.280164       1 bundle.go:151] trust/bundle "msg"="deleted old target key" "bundle"="my-org.com" "namespace"="test-1000" "old_target"={"configMap":{"key":"ca.crt"},"namespaceSelector":{"matchLabels":{"foo":"bar"}}}
I0119 13:29:15.481179       1 sync.go:230] trust/bundle "msg"="synced bundle to namespace" "bundle"="my-org.com" "namespace"="test-1000"
```

Related (wrt whether we should prioritize CA data being always available): #99

erikgb:
I also noted this behavior when preparing a pull request for trust-manager. I wonder if there is a particular reason for doing it this way - as a two-step process? Maybe @JoshVanL has some insights?

inteon:
As mentioned by @erikgb. This should be fixed on master thanks to #89.

PR #228, opened 2023-11-09
Tweak readme to encourage use

SgtCoDFish:
Follows from something mentioned at the cert-manager booth at KubeCon!

PR #230, opened 2023-11-11
chore!: remove .status.target field from Bundle API

erikgb:
... to simplify Bundle reconciler after migration to SSA. The now complex nested structure, inherited from .spec.target is almost not in effective use at present. I think maintaining a copy of the target field made sense before we migrated to SSA, since we had to do reconciliation in two steps, ref. https://github.com/cert-manager/trust-manager/issues/100. Many new features have also been added, making the `BundletTarget` struct grow.

Yes, I know this is a breaking change, but we are still on `v1lpha1` version of the API. Also the API will be broken in under status (observed state), and I don't think many users will be affected by it.


SgtCoDFish:
(I'm on holiday today so I probably won't be able to respond today beyond this comment)

As you can probably guess, I'm nervous about making any breaking change here - even for v1alpha1 😂 

I spoke to a bunch of people at KubeCon who wanted to use more trust-manager but were worried about production-readiness. #228 was sparked by those discussions. I'm of the opinion that trust-manager is ready for prod and I want to encourage that as much as possible.

One of the things that was suggested at KubeCon was to create v1beta1 soon. I'd be quite comfortable with making this change in the migration to v1beta1. Would it be fine to create v1beta1 in the near future and leave v1alpha1 unbroken?

If we have to change v1alpha1, would it be possible to make `.status.target` an empty struct or a static dummy or something? Or is that going to be too cumbersome?

I'm not saying "no" to this PR, just exploring the alternatives. I recognise what you're aiming for I think!

erikgb:
> As you can probably guess, I'm nervous about making any breaking change here - even for v1alpha1 😂

In the spec part, I would expect that. But this is the status part - which I don't think many users really use. 😉 

> I spoke to a bunch of people at KubeCon who wanted to use more trust-manager but were worried about production-readiness. #228 was sparked by those discussions. I'm of the opinion that trust-manager is ready for prod and I want to encourage that as much as possible.

Cool! I agree it's approaching production readiness - even if we use it in production without any issues already.

> One of the things that was suggested at KubeCon was to create v1beta1 soon. I'd be quite comfortable with making this change in the migration to v1beta1. Would it be fine to create v1beta1 in the near future and leave v1alpha1 unbroken?

By adding a beta version, we send a signal about a more stable API. What if we create v1beta1, but make it unserved for now? I think we should improve the API before approaching GA.

> If we have to change v1alpha1, would it be possible to make `.status.target` an empty struct or a static dummy or something? Or is that going to be too cumbersome?

Why? We can drop using `.status.target` in the controller, but that's still breaking - not in a more sophisticated manner. 😆 

> I'm not saying "no" to this PR, just exploring the alternatives. I recognise what you're aiming for I think!

My goal is to simplify things. I am big fan of simplicity.

pkg/bundle/bundle_test.go
```diff
@@ -1215,14 +1160,17 @@ func Test_Reconcile(t *testing.T) {
 			},
 			expEvent: `Normal Synced Successfully synced Bundle to all namespaces`,
 		},
-		"if Bundle had a Secret target, and Secret targets are disabled, return an error": {
+		"if Bundle has Secret target, and Secret targets are disabled, return an error": {
```
inteon, line 1163: @erikgb This test was testing that all secrets are removed if the resource had a Secret target.
Now, Bundles that pointed to Secrets will not cleanup ConfigMaps if the secrets are being disabled.
Not sure if this is a problem, but I just want to make sure you note this.

erikgb: So a user will disable secret targets and modify the Bundle _at the same time_? If you disable secret targets, but leave the  bundle unchanged, it will error out. If you update the bundle and then disable secret targets, the secrets should already be deleted? Or are you thinking of a scenario I don't see? In that case, we could reintroduce some state under status, but I would prefer to avoid it.

pkg/bundle/sync.go
```diff
@@ -546,9 +546,9 @@ func (b *bundle) needsUpdate(ctx context.Context, log logr.Logger, obj *metav1.P
 		}
 
 		if targetType == "ConfigMap" {
-			if bundle.Status.Target != nil && bundle.Status.Target.ConfigMap != nil {
+			if bundle.Spec.Target.ConfigMap != nil {
```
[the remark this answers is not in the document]
erikgb: What is old/new in this context? Isn't this part of the configmap reconcile?

PR #232, opened 2023-11-14
Add two new variables to the Helm chart

bodgit:
* Allow customising the `revisionHistoryLimit` field in the Deployment, defaulting to the Kubernetes default of 10.
* Allow customising the volume used for storing the default trust package, defaulting to the original `emptyDir: {}` value.

### DIFF
--- a/deploy/charts/trust-manager/templates/deployment.yaml
+++ b/deploy/charts/trust-manager/templates/deployment.yaml
@@ -7,6 +7,9 @@ metadata:
 {{ include "trust-manager.labels" . | indent 4 }}
 spec:
   replicas: {{ .Values.replicaCount }}
+  {{- if ne (quote .Values.revisionHistoryLimit) (quote "") }}
+  revisionHistoryLimit: {{ .Values.revisionHistoryLimit }}
+  {{- end }}
   selector:
     matchLabels:
       app: {{ include "trust-manager.name" . }}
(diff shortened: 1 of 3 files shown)


deploy/charts/trust-manager/templates/deployment.yaml
```diff
@@ -120,7 +121,7 @@ spec:
       {{- end }}
       volumes:
       - name: packages
-        emptyDir: {}
+        {{- toYaml .Values.packagesVolume | nindent 8 }}
```
inteon, line 124: Could you give an example of how and why you would change this value?

bodgit:
@inteon did you have any further feedback?

deploy/charts/trust-manager/templates/deployment.yaml
SgtCoDFish: Hey, just seen this PR - I think we'll go in a different direction with this by setting a size limit unilaterally in #308  - appreciate your involvement. If you wanted to reword this to only focus on revisionHistoryLimit (along with maybe some justification for why we should change that) we'd be happy to take a look!

SgtCoDFish:
Gonna close this since we went a different way with it! Thanks so much @bodgit for raising something - appreciate you getting involved even if we didn't merge this ❤️

bodgit:
> Gonna close this since we went a different way with it! Thanks so much @bodgit for raising something - appreciate you getting involved even if we didn't merge this ❤️

I will redo the PR for just the `revisionHistoryLimit` settings.

SgtCoDFish:
Ah sorry - I missed that part entirely!

Yeah if you raise a PR I'll take a look at that. Please feel free to assign it to me and ping me for a review!

PR #235, opened 2023-11-16
Add pkcs12 password feature into Bundle resource

arsenalzp:
It is another PR related to #199. It adds PKCS12 arbitrary password feature into Bundle resource.


arsenalzp:
All conflicts were fixed, all work was re-based.

deploy/charts/trust-manager/templates/trust.cert-manager.io_bundles.yaml
```diff
@@ -185,6 +188,62 @@ spec:
...
+                            type: string
+                    secret:
+                      description: Secret is the target Secret that all Bundle source data will be synced to. Using Secrets as targets is only supported if enabled at trust-manager startup. By default, trust-manager has no permissions for writing to secrets and can only read secrets in the trust namespace.
+                      type: object
+                      required:
+                        - key
+                      properties:
+                        key:
+                          description: Key is the key of the entry in the object's `data` field to be used.
+                          type: string
```
erikgb, line 246: I don't think this section should be added.

pkg/apis/trust/v1alpha1/types_bundle.go
```diff
@@ -122,6 +122,8 @@ type AdditionalFormats struct {
 	// PKCS12 requests a PKCS12-formatted binary trust bundle to be written to the target.
 	// The bundle is created without a password.
```
erikgb, line 123: Must be updated.

pkg/apis/trust/v1alpha1/types_bundle.go
```diff
@@ -122,6 +122,8 @@ type AdditionalFormats struct {
...
 	PKCS12 *KeySelector `json:"pkcs12,omitempty"`
+	// Arbitrary password for bundle stores
+	Password string `json:"password,omitempty"`
```
erikgb, line 126: Are we forcing the same password on JKS and PKCS12? I am not sure it's the most flexible solution.

erikgb (changes requested):

pkg/apis/trust/v1alpha1/types_bundle.go
arsenalzp: > Are we forcing the same password on JKS and PKCS12? I am not sure it's the most flexible solution.
Yes, exactly, because pksc12 and jks are two formats of the same bundle, I don't expect customer wants to have different password for each format of the same bundle.
Does it make sense?

erikgb: > Does it make sense?

It really depends on how you look at truststore passwords: If you think the password is a sensitive secret that needs to be protected , it could make sense. But those will not be satisfied with what we are suggesting in this PR anyway. 😄 As commented in https://github.com/cert-manager/trust-manager/issues/199#issuecomment-1806384505 they think we need to put the password/passwords in secrets.

Then there are others, like me, who just consider these binary truststore formats as troublesome and want to use the more standardized PEM format. But are forced to use JKS or PKCS12 by tools that don't want to support PEM. We want it to be as simple as possible to use. And for Java, the situation is like this:

- Older Java Runtime Environments are delivered with a default JKS truststore, where the password is `changeit`. I would like to replace this default JKS truststore with as little effort as possible, which means that I want the password to be `changeit` - to avoid reconfiguration of all applications running on the JRE.
- Newer Java Runtime Environments have migrated to password-less PKCS12 truststore, ref. https://bugs.openjdk.org/browse/JDK-8275252. In this case I want the truststore to be password-less, for the same reasons as described for the JKS format.

So forcing the same password on both truststore formats will block users like me, who just want this to be as simple as possible and don't think the truststore password is a sensible thing. A workaround is to create two bundle resources, but that will waste a lot of space in etcd - since there is currently no way to turn off the PEM format in bundles.

Does this makes sense?

erikgb: @arsenalzp If you have time, please join our daily stand-up today at 11:30 CET. I thought this would be a "good first issue", but it has turned out to be a lot more challenging than anticipated. Thanks for the stamina!

arsenalzp: Hello,
Yes, it does make sense for me. I implemented a new structures with password parameter for each one, as was proposed by you and your colleague, it wasn't so complicated from the technical perspective, because code is quite clear, however some tests are not so clear for me, so they were falling, but I fixed all issues. I need more time to trim the code.
Thank you for the invitation, I would like to join, but today is Public holidays in Czech Republic, so I'm far away from my laptop and reliable Internet.

pkg/bundle/sync.go
```diff
@@ -244,12 +252,10 @@ func certAlias(derData []byte, friendlyName string) string {
 	return certHash[:8] + "|" + friendlyName
 }
 
-type pkcs12Encoder struct {
-	password string
-}
+func (e pkcs12Encoder) encode() ([]byte, error) {
+	var encoder = pkcs12.Modern2023 // user modern PKCS#12 encoder
```
erikgb, line 256:
```suggestion
	var encoder = pkcs12.LegacyDES
```

When working on https://github.com/cert-manager/cert-manager/pull/6515 I realize that we prefer compatibility over creating a more secure trust store. PKCS12 and JKS are insecure anyway and should be protected by other means than the password.

erikgb:
@arsenalzp A tip: Try to use rebase instead of merge. The merge commits creates a lot of issues with DCO.

pkg/bundle/sync.go
arsenalzp: I took you remark into account in a new PR #244 . I decided to re-create it due to mess of remarks and uncertain decisions.

PR #237, opened 2023-11-18
refactor: simplify patching resources using generics and Unstructured

erikgb:
This PR makes use of the Bundle API apply configurations introduced in https://github.com/cert-manager/trust-manager/pull/217.  The PR also suggests an alternative approach to do SSA using the controller-runtime client. You might have good reasons for keeping the existing approach, but we are using the alternative approach proposed in this PR with success. And it is at least more compact in code measures. Please let me know what you think.

I am planning to simplify even more, by just using `BundleStatusApplyConfiguration`, after/if https://github.com/cert-manager/trust-manager/pull/236 is merged. The utility functions to handle `needsUpdate` is blocking a well-written SSA reconciler IMO. 😉 

/cc @inteon @SgtCoDFish


pkg/bundle/ssa.go
```diff
@@ -0,0 +1,73 @@
...
+// so following the established pattern to use bundle as receiver does not work here.
+
+// applyConfiguration is a type constraint for apply configurations in use in this package.
+type applyConfiguration interface {
+	*trustapiac.BundleApplyConfiguration | *corev1.ConfigMapApplyConfiguration | *corev1.SecretApplyConfiguration
+}
+
+// patchResource patches the resource using SSA forcing ownership of fields
+func patchResource[AC applyConfiguration](ctx context.Context, w client.Writer, ac AC) error {
+	o, err := runtime.DefaultUnstructuredConverter.ToUnstructured(ac)
```
inteon, line 44: I would like to keep using the simple json encode here instead.
I think this "ToUnstructured" call adds another json round-trip.

erikgb: I am not sure if the same thing is possible with this approach, but I will take a look. I agree it would probably make a slight performance enhancement, but to me simple and readable code is more important. This not time-critical code. But I will take another look...

erikgb: No, I don't think it's possible. Created an alternative PR: https://github.com/cert-manager/trust-manager/pull/239

PR #239, opened 2023-11-20
refactor: use generated Bundle API apply configurations

erikgb:
This PR is an alternative to https://github.com/cert-manager/trust-manager/pull/237 that also uses the generated Bundle API apply configurations introduced in https://github.com/cert-manager/trust-manager/pull/217, but uses the established pattern for SSA-patching resources from apply configurations.

When this PR is hopefully merged, I suggest moving the generated apply configurations to an exported package - which will make them available to any user pulling in this project as a dependency.


pkg/bundle/bundle.go
```diff
@@ -94,13 +95,27 @@ type bundle struct {
 func (b *bundle) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	result, statusPatch, resultErr := b.reconcileBundle(ctx, req)
 	if statusPatch != nil {
-		con, patch, err := ssa_client.GenerateBundleStatusPatch(req.Name, req.Namespace, statusPatch)
+		applyConfig := trustapiac.Bundle(req.Name, req.Namespace).
+			WithStatus(trustapiac.BundleStatus().WithConditions())
```
inteon, line 99:
```suggestion
			WithStatus(trustapiac.BundleStatus())
```
Is there a reason you added `.WithConditions()` here?

pkg/bundle/bundle.go
```diff
@@ -94,13 +95,27 @@ type bundle struct {
...
+		applyConfig.Status.DefaultCAPackageVersion = statusPatch.DefaultCAPackageVersion
+		for _, condition := range statusPatch.Conditions {
+			applyConfig.Status.Conditions = append(applyConfig.Status.Conditions,
+				*trustapiac.BundleCondition().
+					WithType(condition.Type).
+					WithStatus(condition.Status).
+					WithLastTransitionTime(*condition.LastTransitionTime).
+					WithReason(condition.Reason).
+					WithMessage(condition.Message).
+					WithObservedGeneration(condition.ObservedGeneration))
```
inteon, line 109:
```suggestion
			applyConfig.Status.WithConditions(
				trustapiac.BundleCondition().
					WithType(condition.Type).
					WithStatus(condition.Status).
					WithLastTransitionTime(*condition.LastTransitionTime).
					WithReason(condition.Reason).
					WithMessage(condition.Message).
					WithObservedGeneration(condition.ObservedGeneration))
```

pkg/bundle/bundle.go
erikgb: No, thanks - fixed. 😉

pkg/bundle/bundle.go
erikgb: Thanks, fixed.

erikgb:
@inteon The failing test now seems to be a real flake. I got it a couple of times yesterday as well. Any idea what might be the root cause?

SgtCoDFish:
Capturing an important point of discussion from the standup on 2023-11-21: This PR would be a no-op in practice (i.e., this shouldn't change any functionality for anyone running trust-manager).

The motivation here is dogfooding - we think clients should use the apply configurations when interacting with trust-manager resources, and so there's value in us doing that (so everyone's working with the same API)

pkg/bundle/internal/ssa_client/patch.go
```diff
@@ -17,10 +17,86 @@ limitations under the License.
...
 	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	trustapi "github.com/cert-manager/trust-manager/pkg/apis/trust/v1alpha1"
+	trustapiac "github.com/cert-manager/trust-manager/pkg/applyconfigurations/trust/v1alpha1"
 )
 
+func GenerateBundlePatch(
+	bundlePatch *trustapiac.BundleApplyConfiguration,
+) (*trustapi.Bundle, client.Patch, error) {
+	if bundlePatch == nil || bundlePatch.Name == nil {
```
inteon, line 35: @erikgb can you also check that the patch does not have a namespace?

inteon: Also, can we change this back to `GenerateBundleStatusPatch` and only patch the status?

pkg/bundle/util.go
```diff
@@ -18,25 +18,27 @@ package bundle
...
+	"k8s.io/utils/ptr"
 
 	trustapi "github.com/cert-manager/trust-manager/pkg/apis/trust/v1alpha1"
+	trustapiac "github.com/cert-manager/trust-manager/pkg/applyconfigurations/trust/v1alpha1"
 )
 
 // bundleHasCondition returns true if the bundle has an exact matching condition.
 // The given condition will have the ObservedGeneration set to the bundle Generation.
 // The LastTransitionTime is ignored.
 func bundleHasCondition(
```
inteon, line 30: I feel like using the ApplyConfiguration object in this function really adds a lot of unnecessary logic.
I would like to keep using trustapi.BundleCondition here.
In general, I think there is no reason to have `trustapiac.BundleConditionApplyConfiguration`, since we will always atomically edit such a struct (and not its individual fields).

pkg/bundle/internal/ssa_client/patch.go
erikgb: > can you also check that the patch does not have a namespace?

Yes, we can do that, makes sense.

> Also, can we change this back to GenerateBundleStatusPatch and only patch the status?

Why? A patch is a patch.

pkg/bundle/util.go
erikgb: > I feel like using the ApplyConfiguration object in this function really adds a lot of unnecessary logic.

Which logic is added? The logic is the same - we just have to be more explicit about zero-values, mainly because of the tests that don't set mandatory fields. I think this function should not exist, and instead patch unconditionally or have a more "technical" check if a patch is needed or not. IMO the established pattern of `needsUpdate` is fragile in general and adds unnecessary complexity - both code and tests.

> I would like to keep using trustapi.BundleCondition here.

Where would you then put the conversion logic? I feel this comment is more about your general objection to apply configurations commented below....

> In general, I think there is no reason to have trustapiac.BundleConditionApplyConfiguration, since we will always atomically edit such a struct (and not its individual fields).

I notice you are not a big fan of apply configurations, and that's a fair point of view. 😸 If we don't believe that's the way forward, we should probably remove the generation of them. I agree that AC's are not ideal, but I think it gives us some more freedom when modeling the resources Go API. Without them, we have to make sure we **always** use a pointer type - if not we risk breaking the SSA semantics.

erikgb:
It seems like there is no desire to use applyconfigurations in the maintainer team, closing. I'll create a PR to remove the generation of the ACs, as there is no point in having them if we're not going to use them. :-)

PR #240, opened 2023-11-21
refactor: replace BundleCondition with upstream metav1.Condition

erikgb:
As discussed in today's stand-up, here is a PR to replace our custom `BundleCondition` with the upstream standard `metav1.Condition`. Only downside of this PR, which might be a reason for not merging it, is the condition API docs that are now inherited from upstream. This goes into the CRD, and IMO the wording was better before. 😆

Note to reviewers: I am removing the webhook validation of duplicate bundle conditions, as I think it's not needed anymore. We have set the `+listType` and `+listMapKey` markers on the `.status.conditions` field, so the api-server should handle eventual duplicates.


pkg/apis/trust/v1alpha1/types_bundle.go
```diff
@@ -165,43 +165,11 @@ type BundleStatus struct {
...
-	// set based upon.
-	// For instance, if .metadata.generation is currently 12, but the
-	// .status.condition[x].observedGeneration is 9, the condition is out of date
-	// with respect to the current state of the Bundle.
-	// +optional
-	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
-}
-
 // BundleConditionType represents a Bundle condition value.
 type BundleConditionType string
```
SgtCoDFish, line 169: suggestion: Is this type now redundant? If so I'd be comfortable removing it

erikgb: Yes, I just forgot to remove it. Thanks! Fix incoming.

SgtCoDFish (commented):
Soft LGTM from me, I think this looks like a great simplification! Don't have time to review in depth right now hence the soft LGTM but I took a quick look and I like it 😁 Thank you!

pkg/webhook/validation.go
```diff
@@ -216,16 +216,6 @@ func (v *validator) validate(ctx context.Context, obj runtime.Object) (admission
 		}
 	}
 
-	path = field.NewPath("status")
-
-	conditionTypes := make(map[trustapi.BundleConditionType]struct{})
-	for i, condition := range bundle.Status.Conditions {
```
inteon, line 222: Why are we removing this check?

erikgb: Please read the PR description. Do you think it's still needed? Any duplicates will be resolved by the map semantics on the list.

erikgb: I just noticed it when I had to change the map key type to `string`. But I can leave it if you insist. 😉 But I still think it's not required to have it.

pkg/bundle/util.go
```diff
@@ -26,64 +27,42 @@ import (
...
-		(*patchConditions)[idx] = newCondition
-
-		return newCondition
-	}
-
-	// If we've not found an existing condition of this type, we simply insert
-	// the new condition into the slice.
-	*patchConditions = append(*patchConditions, newCondition)
-
+	meta.SetStatusCondition(patchConditions, newCondition)
```
inteon, line 65: `meta.SetStatusCondition` contains some code paths that we won't use (eg. setting LastTransitionTime).
I think that explicitly defining only the necessary logic here will make this function less error-prone.
The underlying reason is that `meta.SetStatusCondition` was created for CSA, not SSA.

erikgb: WDYT about removing the whole thing, and just append the new condition unconditionally? It's a bit counter-intuitive, at least to me, to account for overriding conditions when building up a SSA patch. To me, that's a code smell...

pkg/webhook/validation.go
erikgb: The change making this webhook validation obsolete was introduced in https://github.com/cert-manager/trust-manager/pull/173/files.

pkg/bundle/util.go
inteon: I don't think we should silently allow you to create an invalid set of conditions.
We could throw an error if the condition is already set, but that code might be equally complex.

erikgb: That is actually a good idea! I will add the code to validate this, and I think it will be very simple.

erikgb: Ok, I have added the validation. Now with a panic, but I could change the signature to return an error.

pkg/bundle/util.go
```diff
@@ -51,6 +52,11 @@ func (b *bundle) setBundleCondition(
 	patchConditions *[]metav1.Condition,
 	newCondition metav1.Condition,
 ) metav1.Condition {
+	cond := meta.FindStatusCondition(*patchConditions, newCondition.Type)
+	if cond != nil {
```
inteon, line 56:
```suggestion
	if cond := meta.FindStatusCondition(*patchConditions, newCondition.Type); cond != nil {
```

deploy/charts/trust-manager/templates/trust.cert-manager.io_bundles.yaml
```diff
@@ -153,32 +153,46 @@ spec:
                   description: List of status conditions to indicate the status of the Bundle. Known condition types are `Bundle`.
                   type: array
                   items:
-                    description: BundleCondition contains condition information for a Bundle.
+                    description: "Condition contains details for one aspect of the current state of this API Resource. --- This struct is intended for direct use as an array at the field path .status.conditions.  For example, \n type FooStatus struct{ // Represents the observations of a foo's current state. // Known .status.conditions.type are: \"Available\", \"Progressing\", and \"Degraded\" // +patchMergeKey=type // +patchStrategy=merge // +listType=map // +listMapKey=type Conditions []metav1.Condition `json:\"conditions,omitempty\" patchStrategy:\"merge\" patchMergeKey:\"type\" protobuf:\"bytes,1,rep,name=conditions\"` \n // other fields }"
```
inteon, line 156: We should find a way to fix this description.

erikgb: Yes, I will look into that now.

inteon: > I could ask on sig-api-machinery Slack
That would be awesome! Also, maybe we can learn from other projects that do this?

erikgb: I think most projects just take the easy way out and lives with the problem. If you do a quick search in indexed GH projects, there are many hits for this "nonsense": https://github.com/search?q=%22Represents+the+observations+of+a+foo%27s+current+state.%22&type=code

erikgb: Asked a question on [Slack](https://kubernetes.slack.com/archives/C0EG7JC6T/p1700754470441259).

SgtCoDFish:
Adding context to this: checked this PR out locally, then patched the makefile to use the external-applyconfigurations arg first, which produced the following diff:

```diff
diff --git a/Makefile b/Makefile
index 98ec359..e580808 100644
--- a/Makefile
+++ b/Makefile
@@ -115,6 +115,7 @@ generate-applyconfigurations: | $(BINDIR)/code-generator-$(CODE_GENERATOR_VERSIO
 	@echo ">> generating pkg/applyconfigurations..."
 	$(BINDIR)/code-generator-$(CODE_GENERATOR_VERSION)/applyconfiguration-gen \
 		--go-header-file 	hack/boilerplate/boilerplate.go.txt \
+		--external-applyconfigurations k8s.io/apimachinery/pkg/apis/meta/v1.Condition:k8s.io/client-go/applyconfigurations/meta/v1 \
 		--input-dirs		"$(API_DIRS)" \
 		--output-package  	"$(GO_MODULE)/pkg/applyconfigurations" \
 		--trim-path-prefix 	"$(GO_MODULE)" \
diff --git a/pkg/applyconfigurations/trust/v1alpha1/bundlestatus.go b/pkg/applyconfigurations/trust/v1alpha1/bundlestatus.go
index 4ca4f22..8f86afd 100644
--- a/pkg/applyconfigurations/trust/v1alpha1/bundlestatus.go
+++ b/pkg/applyconfigurations/trust/v1alpha1/bundlestatus.go
@@ -19,14 +19,14 @@ limitations under the License.
 package v1alpha1
 
 import (
-	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	v1 "k8s.io/client-go/applyconfigurations/meta/v1"
 )
 
 // BundleStatusApplyConfiguration represents an declarative configuration of the BundleStatus type for use
 // with apply.
 type BundleStatusApplyConfiguration struct {
-	Conditions              []v1.Condition `json:"conditions,omitempty"`
-	DefaultCAPackageVersion *string        `json:"defaultCAVersion,omitempty"`
+	Conditions              []v1.ConditionApplyConfiguration `json:"conditions,omitempty"`
+	DefaultCAPackageVersion *string                          `json:"defaultCAVersion,omitempty"`
 }
 
 // BundleStatusApplyConfiguration constructs an declarative configuration of the BundleStatus type for use with
@@ -38,9 +38,12 @@ func BundleStatus() *BundleStatusApplyConfiguration {
 // WithConditions adds the given value to the Conditions field in the declarative configuration
 // and returns the receiver, so that objects can be build by chaining "With" function invocations.
 // If called multiple times, values provided by each call will be appended to the Conditions field.
-func (b *BundleStatusApplyConfiguration) WithConditions(values ...v1.Condition) *BundleStatusApplyConfiguration {
+func (b *BundleStatusApplyConfiguration) WithConditions(values ...*v1.ConditionApplyConfiguration) *BundleStatusApplyConfiguration {
 	for i := range values {
-		b.Conditions = append(b.Conditions, values[i])
+		if values[i] == nil {
+			panic("nil value passed to WithConditions")
+		}
+		b.Conditions = append(b.Conditions, *values[i])
 	}
 	return b
 }
```

erikgb:
@SgtCoDFish I think your finding could be very useful in another context, since it allows you to avoid re-generating apply configurations (e.g. generated in another project). But I think the goal here is to make `applyconfiguration-gen` treat our own `BundleCondition` like the upstream `metav1.Condition` - NOT generate an apply configuration struct, but just use the API struct. And I don't see how this flag can be used to achieve that. Am I missing something here? 🤔

SgtCoDFish:
> Am I missing something here? 🤔

No, you're not! It's my poor communication, sorry; I was working out why it wasn't generating apply configs for the Condition type, and I think I answered that (it's not generating any because it's an external type, I think).

I don't think the flag helps us to achieve your goal. I guess if I'm right about why it's not generating apply configs, we could create a `github.com/cert-manager/trust-manager-condition` module containing just the `BundleCondition` - but that seems crazy and I don't think we'd actually do that!

erikgb:
Superseded by https://github.com/cert-manager/trust-manager/pull/249.

PR #244, opened 2023-11-25
add PKCS12 arbitrary password feature

arsenalzp:
It is another PR related to #199 . It adds PKCS12 arbitrary password feature into Bundle resource for each store: JKS and PKCS12.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,6 +45,8 @@ CONTAINER_REGISTRY_API_URL ?= https://quay.io/v2/jetstack/cert-manager-package-d
 
 GOPROXY ?= https://proxy.golang.org,direct
 
+GO_SOURCES := $(shell find . -name "*.go") go.mod go.sum
+
 CI ?=
 
 # can't use a comma in an argument to a make function, so define a variable instead
@@ -94,7 +96,9 @@ vet:
 	go vet ./...
 
 .PHONY: build
-build: | $(BINDIR) ## build trust-manager
+build: $(BINDIR)/trust-manager | $(BINDIR) ## build trust-manager for the host system architecture
+
+$(BINDIR)/trust-manager: $(GO_SOURCES) | $(BINDIR)
 	CGO_ENABLED=0 go build -o $(BINDIR)/trust-manager ./cmd/trust-manager
 
 .PHONY: generate
--- a/make/trust-manager-build.mk
+++ b/make/trust-manager-build.mk
@@ -28,7 +28,7 @@ trust-manager-save:
 
 .PHONY: trust-manager-load
 trust-manager-load:
-	$(call build_trust_manager,type=docker,latest,linux/amd64)
+	$(call build_trust_manager,type=docker,latest,linux/$(ARCH))
 
 .PHONY: trust-manager-push
 trust-manager-push:
--- a/make/trust-package-debian.mk
+++ b/make/trust-package-debian.mk
@@ -37,7 +37,7 @@ endif
 
 .PHONY: trust-package-debian-load
 trust-package-debian-load:
-	$(call build_debian_trust_package,type=docker,latest,linux/amd64)
+	$(call build_debian_trust_package,type=docker,latest,linux/$(ARCH))
 
 .PHONY: trust-package-debian-push
 trust-package-debian-push:


pkg/apis/trust/v1alpha1/types_bundle.go
```diff
@@ -146,6 +146,8 @@ type SourceObjectKeySelector struct {
 type KeySelector struct {
 	// Key is the key of the entry in the object's `data` field to be used.
 	Key string `json:"key"`
+	// Arbitrary password for particular store (JKS or PKCS12)
+	Password string `json:"password,omitempty"`
```
erikgb, line 150: I don't think you can add this field to the `KeySelector` struct, as it is also used by sources and PEM encoding - that don't have passwords.

arsenalzp: That was taken into account, if no password was set, then default one is used for JKS, and no one is used for PKCS12. Or do you mean that JKS and pkcs12 are uset as sourses for othe Bundle resources?
If so, what is your proposal?

erikgb: Look at the changes in the generated API spec! You are adding new fields multiple places in the API. I think this new `password` field should only be added to `.spec.target.additionalFormats.jks` and `.spec.target.additionalFormats.pkcs12`. I think you should create a new struct in the API and inline the `KeySelector`:

````go
type Keystore struct {
	KeySelector `json:",inline"`

	Password string `json:"password,omitempty"`
}
````

Then use this struct for the listed fields.

Comment (arsenalzp) [reply pkg/apis/trust/v1alpha1/types_bundle.go:150]: OMG, that's true password was many times in the different places, it's my mistake, fixing it.

Comment (arsenalzp) [reply pkg/apis/trust/v1alpha1/types_bundle.go:156]: Issue was fixed! Indeed, structure `KeySelector` is used in many places, I didn't take it into account, that was my mistake!

Comment (erikgb) [reply pkg/apis/trust/v1alpha1/types_bundle.go:158]: Can you please rebase the PR, @arsenalzp. I will try to review tonight.

File: pkg/apis/trust/v1alpha1/types_bundle.go [hunk 2]
```diff
@@ -118,10 +118,12 @@ type BundleTarget struct {
 type AdditionalFormats struct {
 	// JKS requests a JKS-formatted binary trust bundle to be written to the target.
 	// The bundle is created with the hardcoded password "changeit".
```
erikgb, line 120:
```suggestion
```

This is no longer true.

pkg/apis/trust/v1alpha1/types_bundle.go
```diff
@@ -118,10 +118,12 @@ type BundleTarget struct {
...
-	JKS *KeySelector `json:"jks,omitempty"`
+	JKS *Truststore `json:"jks,omitempty"`
+	/* JKS *KeySelector `json:"jks,omitempty"` */
```
erikgb, line 122:
```suggestion
```

pkg/apis/trust/v1alpha1/types_bundle.go
```diff
@@ -118,10 +118,12 @@ type BundleTarget struct {
...
 	// PKCS12 requests a PKCS12-formatted binary trust bundle to be written to the target.
 	// The bundle is created without a password.
```
erikgb, line 124:
```suggestion
```

pkg/apis/trust/v1alpha1/types_bundle.go
```diff
@@ -118,10 +118,12 @@ type BundleTarget struct {
...
-	PKCS12 *KeySelector `json:"pkcs12,omitempty"`
+	PKCS12 *Truststore `json:"pkcs12,omitempty"`
+	/* PKCS12 *KeySelector `json:"pkcs12,omitempty"` */
```
erikgb, line 126:
```suggestion
```

pkg/apis/trust/v1alpha1/types_bundle.go
```diff
@@ -148,6 +150,14 @@ type KeySelector struct {
 	Key string `json:"key"`
 }
 
+// Trust store is used to enrich AdditionalFormats structure with
+// arbitrary password feature.
+type Truststore struct {
+	KeySelector `json:",inline"`
+	// Arbitrary password for particular store (JKS or PKCS12)
```
erikgb, line 157:
```suggestion
	// Password to the truststore
```

I also think we should create two structs, one for JKS and another one for PKCS12. This will allow us to configure a static API default for the password. It might look like this for JKS (equivalent for PKCS12).

````go
type JKS struct {
    KeySelector `json:",inline"`

    // Password to JKS trust store
    //+kubebuilder:default=changeit
    Password *string `json:"password,omitempty"`
}
````

Note the pointer used for the optional field.

erikgb (changes requested):
I did a pass of the API changes, and I think we need more work before this eventually can be merged. On second thought, I think this issue was not a "good first issue" after all, especially after we decided to include this change in the API. Sorry about that. But please try to finish, @arsenalzp. I can try to assist. Let's keep focus on the API changes first, as that is the most important part to get right.

PR #247, opened 2023-11-27
Improve package CI error handling

### LINKED ISSUE #246: Incorrect error handling in cert-manager-package-debian updater
We have a CI process which regularly tries to keep our Debian package up to date by setting up a fresh debian container and checking the latest version of `ca-certificates` against the latest tag available for our package bundles.

The logic is in [this file](https://github.com/cert-manager/trust-manager/blob/620be4f7184d2f5cea8fcae604a4c1e1edceff06/hack/update-debian-trust-package.sh).

The offending line is:

https://github.com/cert-manager/trust-manager/blob/620be4f7184d2f5cea8fcae604a4c1e1edceff06/hack/update-debian-trust-package.sh#L93

This line treats a nonzero exit code as evidence that the upstream version doesn't exist. However, that's only true if crane got a 404. If quay.io is down when we try to check the latest version, we'll see (say) a 504 error but the exit code will still be interpreted as the package not existing.

This happened on 2023-11-27. Crane got the following 504 error:

```text
2023/11/27 16:35:18 HEAD request failed, falling back on GET: GET https://quay.io/v2/: unexpected status code 504 Gateway Timeout: <html>
<head><title>504 Gateway Time-out</title></head>
<body>
<center><h1>504 Gateway Time-out</h1></center>
<hr><center>nginx/1.20.1</center>
</body>
</html>
Error: GET https://quay.io/v2/: unexpected status code 502 Bad Gateway: <!doctype html>
<html lang=en>
<title>Redirecting...</title>
<h1>Redirecting...</h1>
<p>You should be redirected automatically to the target URL: <a href="https://quay.io/quay-registry/static/502.html/">https://quay.io/quay-registry/static/502.html/</a>. If not, click the link.

+++ latest image appears not to exist; building and pushing quay.io/jetstack/cert-manager-package-debian:20210119.0
```

And following this, the trust package was rebuilt.

The error handling should be improved so that we specifically look for a "not found" response.

SgtCoDFish:
Fixes #246 - see that issue for details!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,6 +41,7 @@ RELEASE_VERSION ?= v0.7.0
 BUILDX_BUILDER ?= trust-manager-builder
 
 CONTAINER_REGISTRY ?= quay.io/jetstack
+CONTAINER_REGISTRY_API_URL ?= https://quay.io/v2/jetstack/cert-manager-package-debian/manifests
 
 GOPROXY ?= https://proxy.golang.org,direct
 
--- a/hack/update-debian-trust-package.sh
+++ b/hack/update-debian-trust-package.sh
@@ -29,6 +29,7 @@ CTR=${CTR:-docker}
 
 REPO=${1:-}
 DEBIAN_TRUST_PACKAGE_SUFFIX=${2:-}
+REGISTRY_API_URL=${3:-}
 
 DEBIAN_IMAGE=docker.io/library/debian:11-slim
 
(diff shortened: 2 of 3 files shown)


SgtCoDFish:
> To me, this seems like some kind of race condition between the controller and the test.

Wouldn't surprise me 🙃  I'd like to rip the whole test suite to pieces and start fresh tbh!

erikgb:
>  I'd like to rip the whole test suite to pieces and start fresh tbh!

Why? What would you do? I think these tests with a simplified control-plane are very valuable. But I know that the cert-manager team prefers unit tests. Yes, unit tests run a little faster, but tests using the control-plane are better IMO. 😉

SgtCoDFish:
> Why? What would you do? I think these tests with a simplified control-plane are very valuable. But I know that the cert-manager team prefers unit tests. Yes, unit tests run a little faster, but tests using the control-plane are better IMO. 😉

Oh on the contrary, we much much prefer integration + e2e tests by a lot!

I'd like to remove Ginkgo because I think it makes tests more confusing to reason about. I've also been working on something (closed source but hoping I can open it for the cert-manager project) which manages creating the control place / kind clusters in the test itself, which is _much_ more efficient, easier to handle, and much more parallellisable!

That's a thing for another time though!

erikgb (commented):
I am not a basher, so I'll let someone else review this - thought the changes look ok to me. But I wonder why we seem so "scared" of pushing the trust image? Even pushing over an existing tag seems fine to me, since the image build should be reproducible IMO. So there must be a reason. 😉

SgtCoDFish:
> But I wonder why we seem so "scared" of pushing the trust image? Even pushing over an existing tag seems fine to me, since the image build should be reproducible IMO. So there must be a reason.

Pushing over the existing tag is what we did yesterday so it does work 😁

The reason we have the check is that we run this check every couple of hours so we can update quickly when something changes upstream, and it takes about 1 min to run if we don't build and it takes about 23.5 mins to run if we build everything and push. So there's a cost savings element (not huge - maybe $1 a day - but it's not nothing) and there's a "not pushing loads to quay.io" element (as quay.io can be flaky and we want to be good citizens!)

inteon:
Looks good to me @SgtCoDFish, as mentioned before, the image overwrite probably is not a big deal, but we want our scripts to behave more predictably anyway.
/approve
/lgtm

PR #248, opened 2023-11-28
Fix apply configuration generation on macOS

SgtCoDFish:
I noticed this while looking into #240

## Commit 1

On at least macos 14.1 (sonoma) the paste utility seems to require a filename to be specified; before this change, the `API_DIRS` variable fails with the following error:

> usage: paste [-s] [-d delimiters] file ...

## Commit 2

This avoids us spinning up a separate shell when evaluating `API_DIRS`, since `GO_MODULE` already contains the value we want.

This change also changes the variable assignments; make variables are very confusing, but `:=` should only evaluate at startup while `=` will evaluate every time the variable is used.

See [these docs](https://www.gnu.org/software/make/manual/html_node/Recursive-Assignment.html) for how `=` works, and [these](https://www.gnu.org/software/make/manual/html_node/Simple-Assignment.html) for `:=`. From the first link:

> Another disadvantage is that any functions referenced in the definition will be executed every time the variable is expanded. This makes make run slower; worse, it causes the wildcard and shell functions to give unpredictable results because you cannot easily control when they are called, or even how many times.


erikgb:
Thanks Ash! I am just dreaming of a Mac! 😉 I have also verified that this still works on my Windows-Ubuntu.

/lgtm
/approve

PR #249, opened 2023-11-28
BREAKING: chore: align BundleCondition with upstream metav1.Condition

erikgb:
This is a PR that replaces https://github.com/cert-manager/trust-manager/pull/240. Since we decided to not migrate to the upstream `metav1.Condition`, I have tried to align our own `BundleCondition` as much as possible the the current upstream struct. Personally, I think it's an improvement that first and foremost will add more API spec validations better protecting us from controller bugs. This is technically breaking changes, but since the trust-manager controller should be the only actor modifying `Bundle` status, I think it's fine.

Also removing the obsolete validation in the webhook for duplicate conditions. This was code is not needed after https://github.com/cert-manager/trust-manager/pull/173.


deploy/charts/trust-manager/templates/trust.cert-manager.io_bundles.yaml
```diff
@@ -156,6 +156,8 @@ spec:
                     description: BundleCondition contains condition information for a Bundle.
                     type: object
                     required:
+                      - lastTransitionTime
+                      - reason
```
SgtCoDFish, line 160: question: (I recognise that this would also have been added by #240) I guess this is a breaking change if anyone is setting custom conditions on their bundles, right? I'm not really scared by that, but I suppose we'd need to mention it in release notes - maybe we should update this PR's title to breaking?

I'd also be fine with leaving these as _not_ required for now, and then making them required in`v1alpha2` / `v1beta1`. I don't feel strongly either way!

SgtCoDFish (approved):
/lgtm
/approve
/hold

This looks like an improvement to me. A lot of these changes are technically breaking if we nitpick, but I suspect that it'd be worthwhile to make the changes anyway - I don't feel like we _need_ to wait for `v1alpha2` or `v1beta1` for these changes.

The hold is either for us to change this PR to be non-breaking (which I absolutely don't expect us to do) or to add an indidator to the PR description that it might be breaking (and thus would be worth mentioning in release notes)

pkg/apis/trust/v1alpha1/types_bundle.go
```diff
@@ -168,24 +168,34 @@ type BundleStatus struct {
 // BundleCondition contains condition information for a Bundle.
 type BundleCondition struct {
 	// Type of the condition, known values are (`Synced`).
-	Type BundleConditionType `json:"type"`
+	// The regex it matches is (dns1123SubdomainFmt/)?(qualifiedNameFmt)
```
munnerz, line 171: Where can I go to understand an example of `(dns1123SubdomainFmt/)?(qualifiedNameFmt)`? What is the qualified name format? Could we maybe rephrase this to use examples rather than a psuedo-regex-y thing?

deploy/charts/trust-manager/templates/trust.cert-manager.io_bundles.yaml
erikgb: These changes are breaking for sure, but only for actors writing to `Bundle` status. For consumers of status, it is actually an improvement - since we are now promising more. I am open to any solution here, but if we want to make this change totally non-breaking the change should probably be postponed to the next API version.

pkg/apis/trust/v1alpha1/types_bundle.go
erikgb: Good point! I just copied the text from the upstream Condition. 😊 Maybe adding these regex'es is something we should not do - even if it's a deviation from the upstream API?

SgtCoDFish: I think it's fine to add the regex - but I agree with James that the upstream text that's been copied here is not super clear. Maybe we just remove this line - anyone who cares what the regex matches can copy it into some regex matcher site. Most probably, nobody will care because it'll only be maintainers who set bundle conditions!

deploy/charts/trust-manager/templates/trust.cert-manager.io_bundles.yaml
erikgb: **BREAKING** looks very scary, and this won't break for users. But I am ok with it.

pkg/apis/trust/v1alpha1/types_bundle.go
erikgb: Removed it now.

deploy/charts/trust-manager/templates/trust.cert-manager.io_bundles.yaml
SgtCoDFish: I think it's mostly to remind the releaser to write a little paragraph on this in the release notes - we shouldn't make a big scary thing out of this!

SgtCoDFish (approved):
/lgtm
/approve

Thanks for this 😁 Took a bit of discussion to get here but I think this is a worthwhile improvement!

SgtCoDFish:
/lgtm
/approve
/unhold

Thanks!

PR #250, opened 2023-11-29
Better handling of local arch differences

SgtCoDFish:
These commits help when developing on non-amd64 architectures.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,6 +45,8 @@ CONTAINER_REGISTRY_API_URL ?= https://quay.io/v2/jetstack/cert-manager-package-d
 
 GOPROXY ?= https://proxy.golang.org,direct
 
+GO_SOURCES := $(shell find . -name "*.go") go.mod go.sum
+
 CI ?=
 
 # can't use a comma in an argument to a make function, so define a variable instead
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -94,7 +94,9 @@ vet:
 	go vet ./...
 
 .PHONY: build
-build: | $(BINDIR) ## build trust-manager
+build: $(BINDIR)/trust-manager | $(BINDIR) ## build trust-manager for the host system architecture; to cross compile for linux, use "make build-linux-$arch"
```
inteon, line 97: I think this comment does not work for this PR yet, the "build-linux-$arch" target does not exist.

Makefile
```diff
@@ -94,7 +94,9 @@ vet:
...
+
+$(BINDIR)/trust-manager: | $(BINDIR)
```
inteon, line 99: I think we have to add the go files as a dependency, otherwise make will think that the binary is up-to-date.

make/trust-manager-build.mk
```diff
@@ -28,7 +28,7 @@ trust-manager-save:
 
 .PHONY: trust-manager-load
 trust-manager-load:
-	$(call build_trust_manager,type=docker,latest,linux/amd64)
+	$(call build_trust_manager,type=docker,latest,linux/$(ARCH))
```
inteon, line 31: 👍

PR #252, opened 2023-12-02
test: fix flaky tests by introducing komega

erikgb:
This change will hopefully fix the flaky integration tests that have been annoying me in several PRs. I like Ginkgo/Gomega, but I know the maintainers are not big fans. 😁 But as long as we are using Ginkgo/Gomega, I think komega (part of controller-runtime) makes the code more compact and avoids attempts on outdated test resource updates.

/cc @inteon @SgtCoDFish

### DIFF
--- a/test/integration/bundle/suite.go
+++ b/test/integration/bundle/suite.go
@@ -35,6 +35,7 @@ import (
 	"k8s.io/utils/ptr"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/envtest/komega"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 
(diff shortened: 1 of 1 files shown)


SgtCoDFish (approved):
/lgtm
/approve

I wasn't aware of komega but it definitely looks like it makes things simpler! Even if this doesn't fix flakes I feel like it makes things easier to reason about.

>  I like Ginkgo/Gomega, but I know the maintainers are not big fans.

I should expand on this: I feel personally that these tools make the tests harder to understand. I've also been burned by the big changes introduced in Ginkgo v2 which meant we had (and have) to do lots of work to port to the new test format.

Specifically, the flakiness of these tests could be caused by many things but I find the flow so confusing with ginkgo that I can't be sure of where the issue is actually introduced.

I'd prefer plain old go tests - I've never really felt like ginkgo adds much for me personally. But if we can get these tests to be non-flaky with ginkgo, I see no reason to rush to move away from it!

erikgb:
Created manual cherry-pick PR: https://github.com/cert-manager/trust-manager/pull/264

PR #253, opened 2023-12-02
add dedicated structures for PKCS12 and JKS stores

### LINKED ISSUE #199: Support of setting arbitrary password for PKCS12 truststore

arsenalzp:
This PR fixes #199  .
As was proposed by @erikgb, two dedicated structures PKCS12 and JKS were added.
Could you please be so kind to review this PR?


pkg/apis/trust/v1alpha1/types_bundle.go
```diff
@@ -118,10 +118,26 @@ type BundleTarget struct {
...
+	//+kubebuilder:default=changeit
+	Password *string `json:"password"`
+}
+
+type PKCS12 struct {
+	KeySelector `json:",inline"`
+
+	// Password to PKCSK12 trust store
+	//+kubebuilder:default=""
+	Password *string `json:"password"`
```
erikgb, line 140:
```suggestion
        // +optional
	//+kubebuilder:validation:MaxLength=30
	//+kubebuilder:default=""
	Password *string `json:"password"`
```

erikgb (changes requested):
@arsenalzp Please try to avoid closing PRs and opening new ones. It makes it harder to track work on this new feature.

I still think the API additions need a bit more work, and again I think this is the most important part. Making a field with a default required does not make sense to me. Though a password IS required, it has a default. So it is not required from the user's point of view. I've put up some suggestions, but I am not sure about the min/max length requirements. It is not easy to find any good specification of these legacy formats.

pkg/apis/trust/v1alpha1/types_bundle.go
```diff
@@ -118,10 +118,26 @@ type BundleTarget struct {
...
 	// The bundle is created without a password.
-	PKCS12 *KeySelector `json:"pkcs12,omitempty"`
+	PKCS12 *PKCS12 `json:"pkcs12,omitempty"`
+}
+
+type JKS struct {
+	KeySelector `json:",inline"`
+
+	// Password to JKS trust store
+	//+kubebuilder:default=changeit
```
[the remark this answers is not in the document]
arsenalzp: Fixed.

pkg/apis/trust/v1alpha1/types_bundle.go
arsenalzp: Fixed

arsenalzp:
Hello colleagues,
Should I put an improvement in current PR? If so, could you be so kind to put remarks 🙂

erikgb:
> Hello colleagues, Should I put an improvement in current PR? If so, could you be so kind put your remarks 🙂

Please rebase the PR and ensure you have run `make generate`. I still see the password field marked as required in the generated OpenAPI schema, and that's not correct. And another tip, at least when I am the reviewer: please resolve conversations when they are fixed! If needed they can be reopened. It just makes the review process more comfortable for all parties.

pkg/bundle/bundle_test.go
```diff
@@ -553,7 +563,14 @@ func Test_Reconcile(t *testing.T) {
 			existingSecrets: []client.Object{sourceSecret},
 			existingBundles: []client.Object{
 				gen.BundleFrom(baseBundle,
-					gen.SetBundleTargetAdditionalFormats(trustapi.AdditionalFormats{JKS: &trustapi.KeySelector{Key: "target.jks"}}),
+					gen.SetBundleTargetAdditionalFormats(trustapi.AdditionalFormats{
+						JKS: &trustapi.JKS{
+							KeySelector: trustapi.KeySelector{
+								Key: "target.jks",
+							},
+							Password: &jksDefaultPasswordPointer,
```
erikgb, line 571:
```suggestion
							Password: ptr.To(DefaultJKSPassword),
```

erikgb (changes requested):
Added a few suggestions. I wonder if we should **not** use the default passwords in tests. Ideally, the defaults should be just be present in the API spec.

pkg/bundle/sync.go
```diff
@@ -76,6 +75,14 @@ type bundleData struct {
...
 }
 
+type jksEncoder struct {
+	password string
+}
+
+type pkcs12Encoder struct {
+	password string
+}
+
```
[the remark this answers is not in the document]
arsenalzp: I guess to have structures, global vars and consts as well as imports in one place rather than somewhere in the middle of code lines.
If you think that was bad decision then I can move it back.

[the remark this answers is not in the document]
erikgb: I think we should avoid moving it in this PR. I personally prefer to have the structs defined together with the methods having the structs as receiver. But the main point here is to not move things just out of personal preference.

arsenalzp:
All remarks were fixed, branch was re-based as well.

pkg/bundle/sync.go
```diff
@@ -265,7 +264,7 @@ func (e pkcs12Encoder) encode(trustBundle string) ([]byte, error) {
 		})
 	}
 
-	return pkcs12.EncodeTrustStoreEntries(rand.Reader, entries, e.password)
+	return pkcs12.LegacyDES.EncodeTrustStoreEntries(entries, e.password)
```
erikgb, line 267:
```suggestion
	return pkcs12.LegacyRC2.EncodeTrustStoreEntries(entries, e.password)
```

I think we should be conservative about changing the crypto in this PR. Ref. https://github.com/cert-manager/cert-manager/pull/6517

erikgb (changes requested):
Nice, thanks for this. Almost LGTM from me. Are we able to add some tests using this new feature? It seems like all tests use the default passwords.

arsenalzp:
> Nice, thanks for this. Almost LGTM from me. Are we able to add some tests using this new feature? It seems like all tests use the default passwords.

Hello,
Yes, I will add tests for both stores, which use arbitrary password.
Is that correct?

arsenalzp:
I spend the whole evening with troubleshooting PKCS12 test case, and I've just found that PKCS12 encoders are not deterministic:
```
$ openssl pkcs12 -nokeys -in test.pem -export -out test.p12 -password pass:qwerty
$ md5sum test.p12
9176d1b36d911766e74810ff0fa3bbcd  test.p12
$ openssl pkcs12 -nokeys -in test.pem -export -out test.p12 -password pass:qwerty
$ md5sum test.p12
```
89b666d7907c0b8420470f1c25a66c2b  test.p12
[arsenal@arsenal-laptop pkcs12]$

erikgb:
> I spent the whole evening with troubleshooting PKCS12 test case, and I've just found that PKCS12 encoders are not deterministic:

Yes, both JKS and PKCS12 adds some salt making them non-deterministic. I think the tests must be created with that in mind. What's wrong with the existing test approach?

https://github.com/cert-manager/trust-manager/blob/01bd331abb8ee071025e2b8989930a2eb3b1d8e9/pkg/bundle/sync_test.go#L1541-L1573

arsenalzp:
> > I spent the whole evening with troubleshooting PKCS12 test case, and I've just found that PKCS12 encoders are not deterministic:
> 
> Yes, both JKS and PKCS12 adds some salt making them non-deterministic. I think the tests must be created with that in mind. What's wrong with the existing test approach?
> 
> https://github.com/cert-manager/trust-manager/blob/01bd331abb8ee071025e2b8989930a2eb3b1d8e9/pkg/bundle/sync_test.go#L1541-L1573

Thank you for a hint!
I was trying to put test in bundle_test.go. 
Now it works!

pkg/bundle/sync_test.go
```diff
@@ -1313,6 +1330,7 @@ func Test_buildSourceBundle(t *testing.T) {
 		expNotFoundError bool
 		expJKS           bool
 		expPKCS12        bool
+		expArbPasswd     bool // should we use arbitrary password instead of default one?
```
erikgb, line 1333:
```suggestion
		expPassword    *string
```

I wonder if this could simplify the test? If JKS/PKCS12 expected: if `expPassword` is set, use the value, otherwise use default password? That would remove the need for a new constant.

erikgb (approved):
Thanks @arsenalzp! Did a quick review now. Just one suggestion. I would like to take a closer look and another review from a maintainer before we merge. 

Fixes #199 

/approve
/hold

pkg/apis/trust/v1alpha1/types_bundle.go
```diff
@@ -118,10 +118,31 @@ type BundleTarget struct {
...
+	//+kubebuilder:validation:MinLength=1
+	//+kubebuilder:validation:MaxLength=30
+	//+kubebuilder:default=changeit
+	Password *string `json:"password"`
+}
+
+type PKCS12 struct {
+	KeySelector `json:",inline"`
+
+	// Password to PKCSK12 trust store
```
inteon, line 141:
```suggestion
	// Password to PKCS12 trust store
```

inteon (commented):
LGTM, just one small typo

arsenalzp:
> @arsenalzp Maybe you can update the PR title and description? It seems like I am not allowed to do it. I suggest;
> 
>     * rewording the title to describe the feature added
> 
>     * change "relates to" to "fixes" [Support of setting arbitrary password for PKCS12 truststore #199](https://github.com/cert-manager/trust-manager/issues/199) (so the issue will be closed on merge of this PR)

Done!

pkg/apis/trust/v1alpha1/types_bundle.go
```diff
@@ -118,10 +118,31 @@ type BundleTarget struct {
 type AdditionalFormats struct {
 	// JKS requests a JKS-formatted binary trust bundle to be written to the target.
 	// The bundle is created with the hardcoded password "changeit".
```
SgtCoDFish, line 120: suggestion: Looks like this comment will need updating. Might be worth linking to https://cert-manager.io/docs/faq/#keystore-passwords as well!

pkg/apis/trust/v1alpha1/types_bundle.go
```diff
@@ -118,10 +118,31 @@ type BundleTarget struct {
...
-	JKS *KeySelector `json:"jks,omitempty"`
+	JKS *JKS `json:"jks,omitempty"`
 	// PKCS12 requests a PKCS12-formatted binary trust bundle to be written to the target.
 	// The bundle is created without a password.
```
SgtCoDFish, line 123: suggestion: Looks like this comment will need updating. Might be worth linking to https://cert-manager.io/docs/faq/#keystore-passwords as well!

pkg/apis/trust/v1alpha1/types_bundle.go
```diff
@@ -118,10 +118,31 @@ type BundleTarget struct {
...
+	PKCS12 *PKCS12 `json:"pkcs12,omitempty"`
+}
+
+type JKS struct {
+	KeySelector `json:",inline"`
+
+	// Password to JKS trust store
+	//+optional
+	//+kubebuilder:validation:MinLength=1
+	//+kubebuilder:validation:MaxLength=30
```
SgtCoDFish, line 133: suggestion: It's good to have _some_ max length, but I think 30 might be a bit small. A reasonable password to use here would be 32 hex encoded bytes (`openssl rand -hex 32`), which would have a length of 64 bytes. I'd suggest setting the max length to 128 so that there is a max but if someone wants to use a really secure password they can.

pkg/apis/trust/v1alpha1/types_bundle.go
```diff
@@ -118,10 +118,31 @@ type BundleTarget struct {
...
+	//+kubebuilder:default=changeit
+	Password *string `json:"password"`
+}
+
+type PKCS12 struct {
+	KeySelector `json:",inline"`
+
+	// Password to PKCSK12 trust store
+	//+optional
+	//+kubebuilder:validation:MaxLength=30
```
SgtCoDFish, line 143: suggestion: As above, I'd suggest increasing this max length!

pkg/apis/trust/v1alpha1/types_bundle.go
```diff
@@ -118,10 +118,31 @@ type BundleTarget struct {
...
 	// PKCS12 requests a PKCS12-formatted binary trust bundle to be written to the target.
 	// The bundle is created without a password.
-	PKCS12 *KeySelector `json:"pkcs12,omitempty"`
+	PKCS12 *PKCS12 `json:"pkcs12,omitempty"`
+}
+
+type JKS struct {
+	KeySelector `json:",inline"`
+
+	// Password to JKS trust store
```
SgtCoDFish, line 130: nitpick:

```suggestion
	// Password for JKS trust store
```

pkg/apis/trust/v1alpha1/types_bundle.go
SgtCoDFish, line 141: nitpick:

```suggestion
	// Password for PKCS12 trust store
```

pkg/apis/trust/v1alpha1/types_bundle.go
arsenalzp: Ooopps, my mistake, will fix!

pkg/apis/trust/v1alpha1/types_bundle.go
arsenalzp: Fixed.

inteon (commented):
/approve
/lgtm
/hold in case someone else wants to review

pkg/apis/trust/v1alpha1/types_bundle.go
```diff
@@ -118,10 +118,32 @@ type BundleTarget struct {
 type AdditionalFormats struct {
 	// JKS requests a JKS-formatted binary trust bundle to be written to the target.
 	// The bundle is created with the hardcoded password "changeit".
```
erikgb, line 120:
```suggestion
	// The bundle has "changeit" as the default password.
```

pkg/apis/trust/v1alpha1/types_bundle.go
```diff
@@ -118,10 +118,32 @@ type BundleTarget struct {
...
-	JKS *KeySelector `json:"jks,omitempty"`
+	// For more information refer to this link https://cert-manager.io/docs/faq/#keystore-passwords
+	JKS *JKS `json:"jks,omitempty"`
 	// PKCS12 requests a PKCS12-formatted binary trust bundle to be written to the target.
 	// The bundle is created without a password.
```
erikgb, line 124:
```suggestion
	// The bundle is by default created without a password.
```

erikgb (commented):
A couple of suggestions to API docs improvements, but

/lgtm

arsenalzp:
> A couple of suggestions to API docs improvements, but
> 
> /lgtm

Fixes were implemented.

pkg/bundle/sync_test.go
```diff
@@ -1522,6 +1605,25 @@ func Test_buildSourceBundle(t *testing.T) {
...
+			}
+			if test.expPKCS12 {
+				if test.expPassword != nil {
+					password = *test.expPassword
+					test.bundle.Spec.Target.AdditionalFormats.PKCS12.Password = test.expPassword
+				} else {
+					password = DefaultPKCS12Password
+				}
+			}
+
```
erikgb, line 1626: If you apply the suggested changes above, I think this password logic could be simplified and added to the existing expJKS/expPKCS12 below. WDYT? I would prefer to not modify the bundle resource in the assert section.

arsenalzp: Ok, I agree with your suggestions

arsenalzp: Suggestions were implemented.

pkg/bundle/sync_test.go
```diff
@@ -1487,20 +1512,78 @@ func Test_buildSourceBundle(t *testing.T) {
...
+				Sources: []trustapi.BundleSource{
+					{ConfigMap: &trustapi.SourceObjectKeySelector{Name: "configmap", KeySelector: trustapi.KeySelector{Key: "key"}}},
+				},
+				Target: trustapi.BundleTarget{
+					AdditionalFormats: &trustapi.AdditionalFormats{
+						PKCS12: &trustapi.PKCS12{
+							KeySelector: trustapi.KeySelector{
+								Key: pkcs12Key,
+							},
+							Password: ptr.To(DefaultPKCS12Password),
```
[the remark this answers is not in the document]
arsenalzp: Suggestions were implemented.

pkg/bundle/sync_test.go
```diff
@@ -1487,20 +1512,78 @@ func Test_buildSourceBundle(t *testing.T) {
...
+				Sources: []trustapi.BundleSource{
+					{ConfigMap: &trustapi.SourceObjectKeySelector{Name: "configmap", KeySelector: trustapi.KeySelector{Key: "key"}}},
+				},
+				Target: trustapi.BundleTarget{
+					AdditionalFormats: &trustapi.AdditionalFormats{
+						JKS: &trustapi.JKS{
+							KeySelector: trustapi.KeySelector{
+								Key: jksKey,
+							},
+							Password: ptr.To(DefaultJKSPassword),
```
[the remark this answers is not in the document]
arsenalzp: Suggestions were implemented.

PR #257, opened 2023-12-20
fix: apply-configuration gen for Bundle (cluster-scoped)

erikgb:
While trying to migrate one of our operators to SSA, I realized a bug in https://github.com/cert-manager/trust-manager/pull/217. Bundle is a cluster-scoped resource, and needs an additional marker to generate a correct AC constructor. Please see https://kubernetes.slack.com/archives/C02MRBMN00Z/p1702911424438589 for details.


inteon (commented):
/approve
/lgtm

PR #258, opened 2023-12-21
Add label selector option for Secret and ConfigMap sources

### LINKED ISSUE #256: Use label selector to add sources to a bundle
The `Bundle` CR allows to specify a list of `Secret` and `ConfigMap` by name in the `spec.sources`. In environments where the list of certificates to add to the `Bundle` is dynamic, it can become complicated to maintain the `Bundle`.

Would it be possible to specify, in a `Bundle` spec, a label selector for `Secret` and `ConfigMap`? That way the `Bundle` could be automatically updated, based on labels on `Secret` and `ConfigMap`, when those resources are created, updated and deleted.

ocampeau:
This PR adds support to use label selector for Secret and ConfigMap sources. 

Given a `Bundle` that looks like this:

```yaml
kind: Bundle
metadata:
  name: some-bundle
spec:
  sources:
  - useDefaultCAs: true
  - configMapLabelSelector:
      key: ca.crt
      selector:
        matchLabels:
          fruit: apple
  - secretMapLabelSelector:
      key: ca.crt
      selector:
        matchLabels:
          fruit: banana
  target:
    configMap:
      key: ca.crt
```
this behavior will happen:

* all `Secret` that has label `fruit: banana` and a key `ca.crt` will have the value of `ca.crt` included in the `Bundle`
* all `ConfigMap` that has label `fruit: apple` and a key `ca.crt` will have the value of `ca.crt` included in the `Bundle`

Multiple `configMapLabelSelector` and `secretMapLabelSelector` are supported.

In a case like this:
```yaml
kind: Bundle
metadata:
  name: some-bundle
spec:
  sources:
  - useDefaultCAs: true
  - configMapLabelSelector:
      key: ca.crt
      selector:
        matchLabels:
          fruit: apple
  - configMapLabelSelector:
      key: ca.crt
      selector:
        matchLabels:
          fruit: banana
  target:
    configMap:
      key: ca.crt
```

then:

* all `ConfigMap` that has label `fruit: banana` and a key `ca.crt` will have the value of `ca.crt` included in the `Bundle`
* all `ConfigMap` that has label `fruit: apple` and a key `ca.crt` will have the value of `ca.crt` included in the `Bundle`

fixes #256


erikgb:
I wonder if we need to extend the API with "new source types"; could we instead just extend the existing source `configMap` and `secret` types to allow **either** `name` **or** `selector`? Making the following bundle valid:

````yaml
kind: Bundle
metadata:
  name: some-bundle
spec:
  sources:
  - useDefaultCAs: true
  - configMap:
      key: ca.crt
      name: my-configmap
  - configMap:
      key: ca.crt
      selector:
        matchLabels:
          fruit: apple
  - secret:
      key: ca.crt
      name: my-secret
  - secret:
      key: ca.crt
      selector:
        matchLabels:
          fruit: banana
  target:
    configMap:
      key: ca.crt
````

**Comment by ocampeau** [comment]
Yeah definitely! I will update my PR to reuse the existing source type like you mention. It's much better like that.

**Comment by ocampeau** [comment]
It's done. Let me know what you think.

File: pkg/webhook/validation.go [hunk 1]
```diff
@@ -102,8 +102,8 @@ func (v *validator) validate(ctx context.Context, obj runtime.Object) (admission
 			sourceCount++
 			unionCount++
 
-			if len(configMap.Name) == 0 {
-				el = append(el, field.Invalid(path.Child("name"), configMap.Name, "source configMap name must be defined"))
+			if len(configMap.Name) == 0 && configMap.Selector == nil {
```
erikgb, line 105: We must also validate that the user hasn't set both `name` and `selector`.

ocampeau: I just added a test for that. Let me know what you think!

pkg/apis/trust/v1alpha1/types_bundle.go
```diff
@@ -66,13 +66,13 @@ type BundleSpec struct {
 // BundleSource is the set of sources whose data will be appended and synced to
 // the BundleTarget in all Namespaces.
 type BundleSource struct {
-	// ConfigMap is a reference to a ConfigMap's `data` key, in the trust
-	// Namespace.
+	// ConfigMap is a reference (by name) to a ConfigMap's `data` key, or to a
+	// list of ConfigMap's `data` key using label selector, in the trust Namespace
```
erikgb, line 70:
```suggestion
	// list of ConfigMap's `data` key using label selector, in the trust Namespace.
```

For consistency.

pkg/apis/trust/v1alpha1/types_bundle.go
```diff
@@ -66,13 +66,13 @@ type BundleSpec struct {
...
 	// +optional
 	ConfigMap *SourceObjectKeySelector `json:"configMap,omitempty"`
 
-	// Secret is a reference to a Secrets's `data` key, in the trust
-	// Namespace.
+	// Secret is a reference (by name) to a Secret's `data` key, or to a
+	// list of Secret's `data` key using label selector, in the trust Namespace
```
erikgb, line 75:
```suggestion
	// list of Secret's `data` key using label selector, in the trust Namespace.
```

For consistency.

pkg/apis/trust/v1alpha1/types_bundle.go
```diff
@@ -158,8 +158,15 @@ type NamespaceSelector struct {
 // in the trust Namespace.
 type SourceObjectKeySelector struct {
 	// Name is the name of the source object in the trust Namespace.
+	// This field must be left empty when `selector` is set
+	//+optional
 	Name string `json:"name"`
```
erikgb, line 163:
```suggestion
	Name string `json:"name",omitempty`
```

pkg/bundle/sync.go
```diff
@@ -23,6 +23,7 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"k8s.io/apimachinery/pkg/labels"
```
erikgb, line 26: This import should be moved to the preceding block. This block is for stdlib imports.

pkg/apis/trust/v1alpha1/types_bundle.go
```diff
@@ -158,8 +158,15 @@ type NamespaceSelector struct {
...
 
+	// Selector is the label selector to use to fetch a list of objects. Must not be set
+	// when `Name` is set.
+	//+optional
+	Selector *metav1.LabelSelector `json:"selector,omitempty"`
```
erikgb, line 168: I notice that we have created a custom type for the namespace selector in this project API: `NamespaceSelector`. I wonder why? Maybe we want to own the API? @inteon WDYT? I am biased here, but leaning towards using the upstream K8s type.

pkg/bundle/sync.go
```diff
@@ -138,41 +139,103 @@ func (b *bundle) buildSourceBundle(ctx context.Context, bundle *trustapi.Bundle)
...
+
+		data, ok := configMap.Data[ref.Key]
+		if !ok {
+			return "", notFoundError{fmt.Errorf("no data found in ConfigMap %s/%s at key %q", b.Namespace, ref.Name, ref.Key)}
+		}
+		return data, nil
+	} else {
+		var configMapList corev1.ConfigMapList
+
+		labelSelectorRequirements, err := labels.ParseToRequirements(metav1.FormatLabelSelector(ref.Selector))
```
erikgb, line 162: I think we should rather use `metav1.LabelSelectorAsSelector` here - to get better/safer error handling. The function `metav1.FormatLabelSelector` seems to return "funny" strings in case of error. Maybe create a small helper to reduce duplication even more?

pkg/webhook/validation.go
```diff
@@ -102,8 +102,11 @@ func (v *validator) validate(ctx context.Context, obj runtime.Object) (admission
...
+				el = append(el, field.Invalid(path.Child("name", "selector"), configMap.Name, "source configMap name or label selector are both empty; one must be set"))
+			}
+			if len(configMap.Name) > 0 && configMap.Selector != nil {
+				el = append(el, field.Invalid(path.Child("name", "selector"), configMap.Name, "source configMap name and label selector are both set; only one must be set"))
```
erikgb, line 109:
```suggestion
				el = append(el, field.Invalid(path, configMap, "source configMap name and label selector are both set; only one must be set"))
```

I think you need to take a look at the error message here. Not sure what would be most user-friendly message. Maybe take a look at the `oneOf` validation? This validation is what we need here, and is supported på OpenAPI, but not yet by controller-runtime.

pkg/webhook/validation.go
erikgb: Looks like a good start, but a bit more work is needed. We could see if there is a way to patch the CRD with a [oneOf](https://swagger.io/docs/specification/data-models/oneof-anyof-allof-not/), which is supported by OpenAPI/Kubernetes, but not yet by c/r, ref. https://github.com/kubernetes-sigs/controller-tools/issues/461.

erikgb (changes requested):
Added some review comments, but mostly nits. I am prepared to move forward with this PR when the code looks good. Well done @ocampeau! 😍

pkg/bundle/sync.go
ocampeau: Done!

pkg/apis/trust/v1alpha1/types_bundle.go
inteon: I think it is very valid to embed the metav1 struct here.
Also, the LabelSelector seems to have good comments that can be used to generate useful OpenAPI descriptions.

pkg/bundle/sync.go
ocampeau: I refactored this part a bit to reduce duplication and used `metav1.LabelSelectorAsSelector`. Let me know what you think!

pkg/webhook/validation.go
ocampeau: When I add the `oneOf` validation manually in the CRD spec, this is the error message that is returned:

```bash
The Bundle "test" is invalid:
* <nil>: Invalid value: "": "spec.sources[1].secret" must validate one and only one schema (oneOf). Found none valid
* spec.sources[1].secret.name: Required value
* spec.sources[1].secret.selector: Required value
* <nil>: Invalid value: "": "spec.sources[2].configMap" must validate one and only one schema (oneOf). Found none valid
* spec.sources[2].configMap.name: Required value
* spec.sources[2].configMap.selector: Required value
```

Unless you says otherwise, I think trying to modify the CRD afterwards to inject the `oneOf` validation can be a fragile solution. So instead, since controller-runtime does not support `oneOf` OpenAPI validation out of the box, I tried to return an error message as close as possible to what the OpenAPI spec returns, with some minor deviations to be as user-friendly as possible.

It goes like that:

```bash
Error from server (Forbidden): error when creating "bundle.yaml": admission webhook "trust.cert-manager.io" denied the request: 
[
spec.sources.[2].configMap: Invalid value: "name: ' ', selector: nil": must validate one and only one schema (oneOf): [name, selector]. Found none valid,
spec.sources.[1].secret: Invalid value: "name: allo, selector: {}": must validate one and only one schema (oneOf): [name, selector]. Found none valid
]
```

1. Let's look at the first error, when both `name` and `selector` are not set:

```text
spec.sources.[2].configMap: Invalid value: "name: ' ', selector: nil": must validate one and only one schema (oneOf): [name, selector]. Found none valid
```

here, I simply hardcoded in the error message an empty string as the value of `name` and `nil` as the value of `selector`.

2. Let's look at the second error, when both `name` and `selector` are set:

```text
spec.sources.[1].secret: Invalid value: "name: allo, selector: {}": must validate one and only one schema (oneOf): [name, selector]. Found none valid
```

here, I parametrized the value of the `name` fields using the value provided by the user, but instead of parametrizing as well the value of `selector`, which turns out to be an ugly stringified version of a Golang map, I instead hardcoded `{}` to indicate that the value is not nil, it is set to some object.


Let me know what you think

pkg/apis/trust/v1alpha1/types_bundle.go
ocampeau: So I guess I can resolve this comment?

pkg/webhook/validation.go
ocampeau: From what I have found Googling around a bit, most people seems to manually add that validation into a webhook instead of patching the CRD afterwards (because it gets overwritten every time we re-generate the CRDs).

See my other comments on the other threads about this issue: https://github.com/cert-manager/trust-manager/pull/258#discussion_r1445222407

inteon: We already have a webhook, so I think there is no need for a oneOf OpenAPI validation.

pkg/bundle/controller.go
```diff
@@ -140,7 +141,12 @@ func AddBundleController(
...
-					// Bundle references this ConfigMap as a source. Add to request.
+					if source.ConfigMap.Selector != nil {
+						if labelsMatchSelector(obj.GetLabels(), source.ConfigMap.Selector) {
+							return true
+						}
+					}
+
 					if source.ConfigMap.Name == obj.GetName() {
 						return true
 					}
```
inteon, line 152:
```suggestion
					} else {
					  if source.ConfigMap.Name == obj.GetName() {
						  return true
					  }
					}
```

pkg/bundle/controller.go
```diff
@@ -215,3 +226,15 @@ func inNamespacePredicate(namespace string) predicate.Predicate {
 		return object.GetNamespace() == namespace
 	})
 }
+
+// labelsMatchSelector returns true is all objLabels matches the label selector
+// and false otherwise
+func labelsMatchSelector(objLabels map[string]string, selector *metav1.LabelSelector) bool {
+	labelRequirements, _ := labels.ParseToRequirements(metav1.FormatLabelSelector(selector))
```
inteon, line 233: I think you can use `metav1.LabelSelectorAsSelector` here.

inteon:
@ocampeau This is an awesome improvement ❤️!
Thanks a lot for contributing this.
/approve
Can be LGTM'd once the two comments above have been resolved ^^

pkg/bundle/controller.go
ocampeau: Indeed! Good catch!

ocampeau:
I have resolved all conversation except this one: https://github.com/cert-manager/trust-manager/pull/258#discussion_r1445222407

Let me know if this is acceptable or not. 

Thanks a lot for accepting this PR by the way, it will really help us. It was a pleasure to contribute to this project.

pkg/webhook/validation.go
erikgb: Thanks for the detailed work on this, @ocampeau! As long as the field reference is correct (example: `spec.sources.[1].secret`) , I wouldn't worry too much about the following text. I like your suggestion of dropping the default string formatting of a Golang struct.

ocampeau: Cool! Thanks I am glad you like it.

erikgb:
> Any ETA on when this could potentially be merged? Do you have documentation on the release velocity of this project?

Sorry, I have been very busy lately. Will try to review this weekend if @inteon hasn't reviewed and merged it before.

We have a couple of unreleased changes already, so I would suggest considering releasing 0.8.0 after this PR is merged. WDYT, @inteon? Sorry for the delay, @ocampeau!

pkg/bundle/controller.go
```diff
@@ -19,6 +19,7 @@ package bundle
 import (
 	"context"
 	"fmt"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
erikgb, line 22: Nit: This should be moved to the next import block. This block is for stdlib imports.

pkg/bundle/controller.go
```diff
@@ -215,3 +222,13 @@ func inNamespacePredicate(namespace string) predicate.Predicate {
 		return object.GetNamespace() == namespace
 	})
 }
+
+// labelsMatchSelector returns true is all objLabels matches the label selector
```
erikgb, line 226:
```suggestion
// labelsMatchSelector returns true if objLabels matches the label selector
```

pkg/webhook/validation.go
```diff
@@ -102,8 +102,11 @@ func (v *validator) validate(ctx context.Context, obj runtime.Object) (admission
 			sourceCount++
 			unionCount++
 
-			if len(configMap.Name) == 0 {
-				el = append(el, field.Invalid(path.Child("name"), configMap.Name, "source configMap name must be defined"))
+			if len(configMap.Name) == 0 && configMap.Selector == nil {
+				el = append(el, field.Invalid(path, "name: ' ', selector: nil", "must validate one and only one schema (oneOf): [name, selector]. Found none valid"))
+			}
+			if len(configMap.Name) > 0 && configMap.Selector != nil {
+				el = append(el, field.Invalid(path, fmt.Sprintf("name: %s, selector: {}", configMap.Name), "must validate one and only one schema (oneOf): [name, selector]. Found none valid"))
```
erikgb, line 109:
```suggestion
				el = append(el, field.Invalid(path, fmt.Sprintf("name: %s, selector: {}", configMap.Name), "must validate one and only one schema (oneOf): [name, selector]. Found both set"))
```

I am not sure about the error message here, but it should indicate that we detected both set - which is invalid.

erikgb (changes requested):
Some minor nits left, but this is close to a LGTM from me!

pkg/bundle/controller.go
ocampeau: I think my IDE shuffles them automatically. Anyway, its fixed now. Good catch.

pkg/webhook/validation.go
ocampeau: I think `Found both set` is good. It makes sens to me at least.

inteon:
/lgtm
Thanks a lot for the feature.
After this gets merged, we can do the v0.8.0 release. Also, @ocampeau could you create a documentation PR on the website to document this new feature?

ocampeau:
Sure I can add documentation. I'll take a look into it

ocampeau:
Documentation added with this PR: https://github.com/cert-manager/website/pull/1391

PR #260, opened 2023-12-25
fix: should reconcile targets consistently

### LINKED ISSUE #259: Bundle is continuously synced when PKCS12 is enabled
When the PKCS12 additional format is enabled trust-managed is continuously syncing the ConfigMaps and ultimately filling up etcd with old versions, compacting and defragging etcd results in it filling back up to several GB size within 15 minutes or so.

Nothing in the logs to suggest why (even with the logging level set to 5) and the logs indicate a successful sync but it fires a sync every seconds or so, stops as soon as I remove the PKCS12 format from the Bundle. Not happening with JKS.

K8s 1.27.7 - RKE2
trust-managed 0.7.0 - Installed from Helm

erikgb:
When checking target keys using managed fields, we must also check `binaryData` field in target configmaps.

Fixes #259

### DIFF
--- a/test/integration/bundle/suite.go
+++ b/test/integration/bundle/suite.go
@@ -447,6 +447,41 @@ var _ = Describe("Integration", func() {
 		}, eventuallyTimeout, eventuallyPollInterval).Should(BeNil(), "checking that bundle was re-added to newly labelled namespace")
 	})
 
+	Context("Reconcile consistency", func() {
+		It("should have stable resourceVersion", func() {
+			var configMap corev1.ConfigMap
+			Expect(cl.Get(ctx, client.ObjectKey{Namespace: "kube-system", Name: testBundle.Name}, &configMap)).To(Succeed())
+			resourceVersion := configMap.ResourceVersion
+			Consistently(komega.Object(&configMap)).Should(HaveField("ObjectMeta.ResourceVersion", Equal(resourceVersion)))
+		})
+
+		It("should have stable resourceVersion for JKS target", func() {
+			Expect(komega.Update(testBundle, func() {
+				testBundle.Spec.Target.AdditionalFormats = &trustapi.AdditionalFormats{
+					JKS: &trustapi.JKS{KeySelector: trustapi.KeySelector{Key: "target.jks"}}}
+			})()).To(Succeed())
+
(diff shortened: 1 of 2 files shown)


inteon:
@erikgb just for context, do you know why the extra reconciliations and patches keep happening? It was expecting the noop patches to not trigger new reconciliations.

erikgb:
@inteon Neither the JKS nor PKCS#12 format is idempotent, so when the check for keys never match it will patch the target with changes. I really don't know why this didn't happen with JKS only. The added tests only fail on PKCS#12 without the fix...

inteon:
@erikgb I think we explicitly try to make the JKS encoded value deterministic: https://github.com/cert-manager/trust-manager/blob/54e8b096c2faf5e2f11454739d2abd4f5a691f75/pkg/bundle/sync.go#L191-L193

inteon:
@erikgb It looks like the pkcs#12 format is using a random salt, which makes it non-idempotent.
see https://github.com/SSLMate/go-pkcs12/blob/v0.4.0/pkcs12.go#L704C5-L706

Now that I fully understand the cause of this issue and why it occurs for pkcs#12 & not for JKS, I'm more than happy to merge this.
/approve
/lgtm

test/integration/bundle/suite.go
```diff
@@ -447,6 +447,41 @@ var _ = Describe("Integration", func() {
...
+			})()).To(Succeed())
+
+			configMap := &corev1.ConfigMap{ObjectMeta: metav1.ObjectMeta{Namespace: "kube-system", Name: testBundle.Name}}
+			Eventually(komega.Object(configMap)).Should(HaveField("BinaryData", HaveKey("target.jks")))
+
+			resourceVersion := configMap.ResourceVersion
+			Consistently(komega.Object(configMap)).Should(HaveField("ObjectMeta.ResourceVersion", Equal(resourceVersion)))
+		})
+
+		It("should have stable resourceVersion for PKCS12 target", func() {
```
inteon, line 471: I really like these tests, feel free to add these to other projects if you have some extra time 😄.

erikgb: > I really like these tests, feel free to also add these to other projects if you have some extra time 😄.

Thanks, I also like them. Reconcile-loop errors like this one can be very annoying, and I have been looking for a good way of testing/detecting them.... I'll see what I have time for. 😉

PR #262, opened 2024-01-02
Upgrade dependencies

inteon:
Result from running "go-mod-upgrade".


erikgb:
@inteon Why is this not picked up by dependabot?

/lgtm

inteon:
> @inteon Why is this not picked up by dependabot?
> 
> /lgtm

Not sure, maybe because it is not versioned?

PR #271, opened 2024-01-10
Update clusterrole permission to resolve errors.

nan-coupa:


nan-coupa:
Will retest with the 0.7.1 chart before opening PR.

PR #273, opened 2024-01-13
Add an option to filter expired certificates from bundle

### LINKED ISSUE #272: Add option to filter out expired certificates
There are [certain tools](https://www.openssl.org/blog/blog/2021/09/13/LetsEncryptRootCertExpire/) which struggle in the face of expired certs in a trust bundle (initially reported [here](https://github.com/cert-manager/trust-manager/issues/242#issuecomment-1883489057)).

These tools might choose an expired root over an unexpired one in the case that two valid chains can be built. For users who rely on these tools, the ability to filter expired certs from their bundles would be valuable since it removes the possibility of the tool picking the wrong cert.

However, it doesn't seem clear that we should _always_ do this or that we should do this _by default_. Having an expired certificate in a bundle does have some value in that it provides clients with more information about the source of a TLS error they receive.

In other words, if I have an expired root in my bundle, I'll get a "certificate expired" error if something chains back to it, rather than a "certificate untrusted" error if there's no matching root in my bundle.

## Proposal

Add some flag (potentially a command-line argument such as `--filter-expired-certificates` if there's no reason for this to be done on a per-bundle basis) which signals to trust-manager that the completed bundles it creates should have expired certs pruned.

Hoega:
Closes #272 

Add an option to skip expired certificates and exclude them from the bundle.

The option is defined as a cli parameter to be passed by the helm charts values:

```
filterExpiredCertificates:
  # -- Whether to filter expired certificates from the trust bundle.
  enabled: false
```


erikgb:
@Hoega Thanks a lot for this PR. I will try to review it soon!

> Please run 'make -s update-helm-docs'

to fix the CI.

pkg/util/cert_pool.go
```diff
@@ -54,7 +55,7 @@ func (cp *certPool) isCertificateDuplicate(certData []byte) bool {
 }
 
 // Append certificate to a pool
-func (cp *certPool) appendCertFromPEM(PEMdata []byte) error {
+func (cp *certPool) appendCertFromPEM(PEMdata []byte, filterExpiredCerts bool) error {
```
SgtCoDFish, line 58: suggestion: I can't imagine a situation where we'd need to configure `filterExpiredCerts` on a per-certificate basis like this.

Do you think it makes sense to add `filterExpiredCerts` as a boolean option on `certPool`, so we'd have `newCertPool(filterExpiredCerts bool) *certPool` instead and `appendCertFromPEM` would keep its current signature?

pkg/util/pem.go
```diff
@@ -43,8 +43,8 @@ import (
 // https://www.rfc-editor.org/rfc/rfc7468
 
 // See also https://github.com/golang/go/blob/5d5ed57b134b7a02259ff070864f753c9e601a18/src/crypto/x509/cert_pool.go#L201-L239
-func ValidateAndSanitizePEMBundle(data []byte) ([]byte, error) {
-	certificates, err := ValidateAndSplitPEMBundle(data)
+func ValidateAndSanitizePEMBundle(data []byte, filterExpiredCerts bool) ([]byte, error) {
+	certificates, err := ValidateAndSplitPEMBundle(data, filterExpiredCerts)
```
SgtCoDFish, line 47: suggestion: I'm uneasy about making a breaking change to an exported function here (and below with `ValidateAndSplitPEMBundle`) - I don't think we need to.

Can we change this so it's not breaking?

SgtCoDFish (commented):
This looks really great - thank you for raising it! I've got a couple of comments - what do you think?

pkg/util/cert_pool.go
Hoega: You are right, that would be better, I will make the change later. Thanks!

Hoega: I have created new functions with a certPool receiver to preserve the signature of the previous function. I am not an expert in Go, so please let me know if you think of a better approach.

pkg/util/cert_pool.go
```diff
@@ -21,19 +21,22 @@ import (
...
 
 // NewCertPool returns a new, empty CertPool.
-func newCertPool() *certPool {
+func NewCertPool(filterExpiredCerts bool) *certPool {
 	return &certPool{
 		certificatesHashes: make(map[[32]byte]struct{}),
 		certificates:       make([]*x509.Certificate, 0),
+		filterExpiredCerts: filterExpiredCerts,
 	}
 }
```
SgtCoDFish, line 41: suggestion: I appreciate you making a change so that we don't break `ValidateAndSanitizePEMBundle`!

I don't think we need to make this specific change - `ValidateAndSanitizePEMBundle` is in the `util` package and so it can call `newCertPool` without an issue. It'd be my preference to keep certPool entirely private so it's easier for us to change later!

I should've suggested this in my first review, but how about keeping `newCertPool` private and adding something like this:

```go
type ValidateAndSanitizeOptions struct {
    FilterExpired bool
}

func ValidateAndSanitizePEMBundleWithOptions(data []byte, opts ValidateAndSanitizeOptions) ([]byte, error) {
    // implementation here ...
}

func ValidateAndSanitizePEMBundle(data []byte) ([]byte, error) {
    opts := ValidateAndSanitizeOptions{
        FilterExpired: false,
    }

    return ValidateAndSanitizePEMBundleWithOptions(data, opts)
}
```

What do you think? 😁

Hoega: Okay, I see. I have made the update.

pkg/util/pem_test.go
```diff
@@ -116,6 +119,44 @@ func TestValidateAndSanitizePEMBundle(t *testing.T) {
...
+	expiredCertCases := map[string]struct {
+		parts          []string
+		expectedOutput []string
+		expectErr      bool
+	}{
+		"valid bundle with valid certs and an expired cert": {
+			parts:          []string{dummy.TestCertificate1, dummy.TestExpiredCertificate, dummy.TestCertificate3},
+			expectedOutput: []string{dummy.TestCertificate1, dummy.TestCertificate3},
+			expectErr:      false,
+		},
```
SgtCoDFish, line 132: comment: I think it'd be good to have a test that the expired cert is _not_ removed if FilterExpired is false, but I don't think it should block this PR: I'll raise a followup PR to add that after this merges :)

SgtCoDFish (approved):
/lgtm
/approve

This looks great - really appreciate you raising this and getting involved in the community ❤️ 

I'll try to get around to releasing a new trust-manager version soon which will include this PR 😁

pkg/util/pem_test.go
SgtCoDFish: Oh wait, the cases above test that - you already have it tested 😁 Thank you!

PR #275, opened 2024-01-17
Test tweaks following #273

SgtCoDFish:
This adds some checks on the new dummy expired certificate, and changes how we test filtering of expired certs.

I'd rather not check the exact output of `ValidateAndSanitizePEMBundleWithOptions` to check for cert expiry, since that'll be brittle if we change what that function outputs. This test instead explicitly checks for expired certs.

### DIFF
--- a/pkg/util/pem_test.go
+++ b/pkg/util/pem_test.go
@@ -18,6 +18,7 @@ package util
 
 import (
 	"bytes"
+	"crypto/x509"
 	"strings"
 	"testing"
 	"unicode/utf8"
@@ -34,8 +35,10 @@ func TestValidateAndSanitizePEMBundle(t *testing.T) {
 	}
 
 	cases := map[string]struct {
-		parts     []string
-		expectErr bool
+		parts              []string
+		filterExpiredCerts bool
+		expectExpiredCerts bool
+		expectErr          bool
 	}{
 		"valid bundle with all types of cert and no comments succeeds": {
 			parts:     []string{dummy.TestCertificate1, dummy.TestCertificate2, dummy.TestCertificate3},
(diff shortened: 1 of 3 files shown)


pkg/util/pem_test.go
```diff
@@ -49,9 +52,10 @@ func TestValidateAndSanitizePEMBundle(t *testing.T) {
 			parts:     []string{dummy.TestCertificate1, string(poisonComment), dummy.TestCertificate2, randomComment, dummy.TestCertificate3, string(poisonComment)},
 			expectErr: false,
 		},
-		"valid bundle with expired cert succeeds": {
-			parts:     []string{dummy.TestCertificate1, dummy.TestExpiredCertificate},
-			expectErr: false,
+		"valid bundle with expired cert succeeds with the expired cert intact": {
+			parts:              []string{dummy.TestCertificate1, dummy.TestExpiredCertificate},
+			expectExpiredCerts: true,
+			expectErr:          false,
```
erikgb, line 58: Could it make sense to omit fields with the correct value for the test case? I makes things a bit more compact.

erikgb (approved):
Thanks @SgtCoDFish! This looks good to me. Just a minor question/comment. Feel free to ignore it.

/approve
/lgtm
/hold

pkg/util/pem_test.go
SgtCoDFish: Could you rephrase that for me? I don't think my morning coffee is kicking in yet and I'm not sure what you mean!

erikgb: Sorry! I just meant that the zero value of a `bool` in Golang is `false`. When do we explicitly set at field to the zero value in the test, and when do we not do it? I think the most common is to omit setting the field.

SgtCoDFish: Ahhh, I'm with you! It's a fair point.

In tests I like to be explicit about them because that makes the test cases easier to copy+paste and change (although I've not been explicit here).

I should be consistent, but I'm gonna merge this since it won't break anything.

SgtCoDFish:
/unhold

Thanks for the review Erik 😁

PR #277, opened 2024-01-18
Bump versions for release, bump deps

SgtCoDFish:
Also adds some vars (which shouldn't change anything) to allow us to tweak CGO_ENABLED and GOEXPERIMENT if we want. That could be a separate PR, but it's a no-op so I figured it'd be OK to include here.

This should be the last step before we can release v0.8.0


erikgb (approved):
/lgtm
/approve

PR #290, opened 2024-02-14
Generate values.schema.json

inteon:


Makefile
```diff
@@ -218,7 +221,7 @@ ensure-trust-manager: ensure-kind kind-load ensure-cert-manager | $(BINDIR)/helm
 	$(BINDIR)/helm-$(HELM_VERSION)/helm uninstall --kubeconfig $(BINDIR)/kubeconfig.yaml -n cert-manager trust-manager || :
 	$(BINDIR)/helm-$(HELM_VERSION)/helm upgrade --kubeconfig $(BINDIR)/kubeconfig.yaml -i -n cert-manager trust-manager deploy/charts/trust-manager/. \
 		--set image.tag=latest \
-		--set defaultTrustPackage.tag=latest$(DEBIAN_TRUST_PACKAGE_SUFFIX) \
+		--set defaultPackageImage.tag=latest$(DEBIAN_TRUST_PACKAGE_SUFFIX) \
```
inteon, line 224: This is actually an error that went unnoticed until now.
Thanks to the values.schema.json validation, we we able to detect that this value contained a typo.

PR #295, opened 2024-02-16
Add optional resources block for init container

niklastanner:
Add optional resources block for the init container of the default package.

### DIFF
--- a/deploy/charts/trust-manager/templates/deployment.yaml
+++ b/deploy/charts/trust-manager/templates/deployment.yaml
@@ -47,6 +47,10 @@ spec:
           seccompProfile:
             type: RuntimeDefault
           {{- end }}
+        {{- if .Values.defaultPackage.resources }}
+        resources:
+          {{- toYaml .Values.defaultPackage.resources | nindent 10 }}
+        {{- end }}
       {{- end }}
       containers:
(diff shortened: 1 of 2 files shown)


SgtCoDFish:
/test pull-trust-manager-verify

(pod was killed during the test I think)

SgtCoDFish:
Hey @niklastanner thanks for raising this (and sorry it took a while to look into)! It looks like your PR is falling foul of the new values schema that was added. Are you able to also update the schema?

niklastanner:
@SgtCoDFish What do you mean by "update the schema"? It's the exact same schema already used by the main image. Building the template locally using `helm template` does generate a valid YAML file.

SgtCoDFish:
My apologies - it looks like running `make generate` will solve the issue. I think the tests hiccuped when I looked - sorry!

SgtCoDFish:
@niklastanner Are you able to run `make generate` locally quickly and commit the result? I might be able to squeeze this into the next trust-manager release this week if we're super quick

SgtCoDFish:
@niklastanner I went ahead and did `make generate` in #316 just so we could include this in trust-manager v0.9.0

Thank you for your contribution, really appreciate you!

PR #304, opened 2024-02-24
Add support for PodMonitor

shlomitubul:
Add support for using PodMonitor


erikgb:
Hi @shlomitubul! Can you explain why you cannot use a `ServiceMonitor`? This PR adds a lot of additional complexity, and `ServiceMonitor` and `PodMonitor` are quite equivalent in function, ref. https://github.com/prometheus-operator/prometheus-operator/issues/3119.

shlomitubul:
@erikgb we want to reduce the use of k8s service when it's not needed as the metrics service only exists to give ServiceMonitor access

shlomitubul:
@erikgb is there anything from my end that need to be handled ?

erikgb:
> @erikgb is there anything from my end that need to be handled ?

I would personally solve your use case (minimize the number of services) differently. Since trust-manager already has a webhook service, I would use this service to scrape metrics also. And keep the `ServiceMonitor` - to reduce Helm chart complexities. But I'll leave the decision to the maintainers. Maybe @inteon has an opinion?

hawksight:
/cc @hawksight 

I will try this out manually to verify, hopefully and review accordingly.
It looks good from a glance, but I'd like to verify with some testing.

PR #306, opened 2024-02-26
Helm: Uniformize all label include statements & add labels to pod template

inteon:
- use the same include label statements everywhere & same as cert-manager uses
- adds the common labels to the pod template
- adds the common labels to resource that were missing them


deploy/charts/trust-manager/templates/deployment.yaml
```diff
@@ -18,6 +18,7 @@ spec:
       {{- end }}
       labels:
         app: {{ include "trust-manager.name" . }}
+        {{- include "trust-manager.labels" . | nindent 8 }}
```
inteon, line 21: We forgot to add the csi-driver-spiffe labels to the pod template.

PR #308, opened 2024-02-27
Set a size limit on emptyDir

SgtCoDFish:
This helps to avoid disk exhaustion attack; 50MB seems to be a reasonable level to start at


inteon (commented):
/approve
/lgtm
Thanks for adding this limit!

PR #319, opened 2024-03-07
Simplify managed fields upgrade from CSA to SSA

erikgb:
This PR simplifies the upgrade from CSA to SSA by using the helper provided by upstream Kubernetes and also used by `kubectl`. There is a theoretical change in behavior when upgrading after this PR, since the helper does not take the target configmap key into account when upgrading. Which means there could potentially be another operator/controller using the c/r regression field manager. I still think the simplification of code and RBAC is worth this minor risk.

### DIFF
--- a/pkg/bundle/bundle.go
+++ b/pkg/bundle/bundle.go
@@ -28,7 +28,9 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/tools/record"
+	"k8s.io/client-go/util/csaupgrade"
 	"k8s.io/utils/clock"
 	"k8s.io/utils/ptr"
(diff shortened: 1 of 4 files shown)


inteon:
Thanks @erikgb.
Great work upstreaming support for sub-resources.
/approve
/lgtm

pkg/bundle/bundle.go
```diff
@@ -454,32 +453,13 @@ func (b *bundle) bundleTargetNamespaceSelector(bundleObj *trustapi.Bundle) (labe
...
-		managedFields[i].Manager = fieldManager
+	patch, err := csaupgrade.UpgradeManagedFieldsPatch(obj, sets.New(fieldManager, crRegressionFieldManager), fieldManager, csaupgrade.Subresource("status"))
+	if err != nil {
+		return false, err
 	}
-
-	if !needsUpdate {
-		return false, nil
+	if patch != nil {
+		return true, b.client.Patch(ctx, obj, client.RawPatch(types.JSONPatchType, patch))
```
inteon, line 461: This is quite a neat trick: instead of update, patch (JSON instead of ssa) is used so there is no extra rbac necessary.

PR #330, opened 2024-03-25
Helm chart - document and add to schema nameOverride

DrFaust92:
This fixes the helm chart validation and adds documentation for this param


SgtCoDFish (approved):
/lgtm
/approve

Thanks - I didn't even know we supported this but it makes sense to document properly and add it to the schema!

PR #353, opened 2024-05-17
CMD new TLS options to manage cipher suites and TLS version

kumirska:
Added CMD options to manage TLC options
- tls-cipher-suites
- min-tls-version

### DIFF
--- a/cmd/trust-manager/app/options/options.go
+++ b/cmd/trust-manager/app/options/options.go
@@ -17,6 +17,7 @@ limitations under the License.
 package options
 
 import (
+	"crypto/tls"
 	"flag"
 	"fmt"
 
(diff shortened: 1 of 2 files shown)


cmd/trust-manager/app/options/options.go
```diff
@@ -174,4 +189,10 @@ func (o *Options) addWebhookFlags(fs *pflag.FlagSet) {
 		"Directory where the Webhook certificate and private key are located. "+
 			"Certificate and private key must be named 'tls.crt' and 'tls.key' "+
 			"respectively.")
+	fs.IntVar(&o.Webhook.MinTlsVersion,
+		"min-tls-version", tls.VersionTLS12,
+		"Minimal TLS version to serve webhook.")
+	fs.StringVar(&o.Webhook.TlsCipherSuites,
+		"tls-cipher-suites", defaultCipherSuites,
```
erikgb, line 196: I don't think we should default the cipher suites explicity. Instead we should use the default in Go if no flag given - as we do right now.

erikgb (changes requested):
Please fix the DCO, ref. https://github.com/cert-manager/trust-manager/blob/main/CONTRIBUTING.md#dco-sign-off

In general, I think we should use the Go defaults if no value provided by user for these new flags. Mainly to avoid maintaining the default ciphers and min TLS version.

Issue #282, opened 2024-01-24
No flag to set structured logging format, e.g. JSON? 


smec42:
Describe the bug:
In https://github.com/cert-manager/cert-manager/pull/5828 the option for setting the log-format to JSON was introduced to cert-manager. Is there a plan to enable this also for trust-manager?
Structured logging can be enabled in the pod spec.

SgtCoDFish:
This is definitely something we'd like - I'm personally very enthusiasic about programs having solid logging with JSON support. Definitely consider this on the "roadmap"!

My gut instinct is for us to aggressively change to using `slog` wherever possible, because I like using the stdlib where we can. But I'm not dogmatic about that.

No timeline for this currently, but I think this should be a thing we do!

arsenalzp:
I can tackle this once decision is approved, if you wish.

PR #354, opened 2024-05-17
Add support for JSON logging format

erikgb:
In this PR I propose to introduce slog.Handler as a backend for logr (and klog) - as documented [here](https://github.com/go-logr/logr?tab=readme-ov-file#using-a-sloghandler-as-backend-for-logr).

I am doing a minimal change in this PR where my primary goal is to add support for a structured log format (ref.  https://github.com/cert-manager/trust-manager/issues/282).

**NOTE**: This might break for some users, depending on if/how they parse trust-manager log output. I cannot imagine the slog text format (the new default) is the same as the deprecated klog format. We always try to avoid breaking changes, but I think a change is required here anyway - since the currently used klog format is deprecated.

Fixes https://github.com/cert-manager/trust-manager/issues/282

### DIFF
--- a/cmd/trust-manager/app/options/options.go
+++ b/cmd/trust-manager/app/options/options.go
@@ -17,8 +17,10 @@ limitations under the License.
 package options
 
 import (
-	"flag"
+	"errors"
 	"fmt"
+	"log/slog"
+	"os"
 
 	"github.com/go-logr/logr"
 	"github.com/spf13/cobra"
@@ -28,14 +30,12 @@ import (
 	"k8s.io/client-go/rest"
 	cliflag "k8s.io/component-base/cli/flag"
 	"k8s.io/klog/v2"
-	"k8s.io/klog/v2/klogr"
 
 	"github.com/cert-manager/trust-manager/pkg/bundle"
 )
 
 // Options is a struct to hold options for trust-manager
 type Options struct {
-	logLevel        string
 	kubeConfigFlags *genericclioptions.ConfigFlags
 
 	// ReadyzPort if the port used to expose Prometheus metrics.
@@ -59,6 +59,45 @@ type Options struct {
 
 	// Bundle are options specific to the Bundle controller.
 	Bundle bundle.Options
+
+	// log are options controlling logging
+	log logOptions
+}
+
+type logOptions struct {
+	format logFormat
+	level  int
+}
+
+const (
+	logFormatText logFormat = "text"
+	logFormatJSON logFormat = "json"
+)
+
+type logFormat string
+
+// String is used both by fmt.Print and by Cobra in help text
+func (e *logFormat) String() string {
+	if len(*e) == 0 {
+		return string(logFormatText)
+	}
+	return string(*e)
+}
+
+// Set must have pointer receiver to avoid changing the value of a copy
+func (e *logFormat) Set(v string) error {
+	switch v {
+	case "text", "json":
+		*e = logFormat(v)
+		return nil
+	default:
+		return errors.New(`must be one of "text" or "json"`)
+	}
+}
+
+// Type is only used in help text
+func (e *logFormat) Type() string {
+	return "string"
 }
 
 // Webhook holds options specific to running the trust Webhook service.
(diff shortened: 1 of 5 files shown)


cmd/trust-manager/app/options/options.go
```diff
@@ -59,6 +59,45 @@ type Options struct {
...
+	logFormatText logFormat = "text"
+	logFormatJSON logFormat = "json"
+)
+
+type logFormat string
+
+// String is used both by fmt.Print and by Cobra in help text
+func (e *logFormat) String() string {
+	if len(*e) == 0 {
+		return "text"
```
erikgb, line 82:
```suggestion
		return string(logFormatText)
```

hawksight (approved):
Had a bit of a read through of the PR and of the linked documentation. This seems to make sense to me. The notion for using slog is that this is in the Go standard library now? I can't approve and can't comment explicitly on the code, but if it works 👍 

Can I run this? just need to branch checkout and build the image?

erikgb:
> Had a bit of a read through of the PR and of the linked documentation. This seems to make sense to me. The notion for using slog is that this is in the Go standard library now? I can't approve and can't comment explicitly on the code, but if it works 👍
> 
> Can I run this? just need to branch checkout and build the image?

Thanks @hawksight! You should be able to just checkout the branch and install it. My workstation environment is really borked for working with open source.

cmd/trust-manager/app/options/options.go
```diff
@@ -82,9 +121,20 @@ func (o *Options) Prepare(cmd *cobra.Command) *Options {
 // Complete will populate the remaining Options from the CLI flags. Must be run
 // before consuming Options.
 func (o *Options) Complete() error {
-	klog.InitFlags(nil)
-	log := klogr.New()
-	flag.Set("v", o.logLevel)
```
SgtCoDFish, line 87: comment: Unrelated to this PR but I'd never actually stopped to consider how bonkers it is that `flag.Set("v", o.logLevel)` is how a verbosity level _should_ be set according to klogr!

cmd/trust-manager/app/options/options.go
```diff
@@ -162,6 +209,16 @@ func (o *Options) addBundleFlags(fs *pflag.FlagSet) {
...
 
+func (o *Options) addLoggingFlags(fs *pflag.FlagSet) {
+	fs.Var(&o.log.format,
+		"log-format",
+		"Log format (text or json)")
+
+	fs.IntVarP(&o.log.level,
+		"log-level", "v", 1,
+		"Log level (1-5).")
+}
```
SgtCoDFish, line 220: suggestion (non-blocking): I think changing the type of this arg is a breaking change.

Currently with the latest release of trust-manager I can pass `--log-level=foo5` and it'll silently ignore the invalid number and default to log level 1.

On this branch, if I pass the same arg it fails:

```text
error: invalid argument "boo1" for "-v, --log-level" flag: strconv.ParseInt: parsing "foo1": invalid syntax
```

I think one option to avoid the break would be to keep this as a string and add `func (lo *logOptions) LogLevel() (int, error)` for getting the log level, so we could handle the error ourselves (and probably just default to log level 1 and log a non-fatal error). I don't think that's required but I think it'd be nice. Your call!

SgtCoDFish (approved):
This looks great, thank you for doing it!

I've added a suggestion but it's not a blocker. Adding approval with a hold in case you want to address the suggestion!

Thanks also to @hawksight for testing this yesterday!

/lgtm
/approve
/hold

cmd/trust-manager/app/options/options.go
erikgb: Agree, but that's history after this PR. 😉

cmd/trust-manager/app/options/options.go
erikgb: Thanks! Good catch! I will revert the change of type.

erikgb: Fixed now, but I think we should accept this breaking change. Who would ever set `--log-level=foo5`? 😉 I consider changing the type as a bugfix. Was the existing behavior (accepting any string) intentional?

erikgb: This was discussed in todays stand-up, and we reached a consensus of considering the switch of `log-level` CLI arg type from string to int a bugfix.

SgtCoDFish (approved):
/lgtm
/approve
/unhold

I really like this, and I think if we document the change from string to int for the log level then that's totally reasonable 👍 

Thanks for doing this Erik!

PR #355, opened 2024-05-20
fix use of system trust bundle when building package

SgtCoDFish:
I missed this during the review of #195 . I shouldn't have missed such a huge bug - my bad!

Today this script uses the _system's_ trust-bundle, not the one from the container.

It breaks development on all platforms which aren't Debian-based linux (e.g. macOS, RHEL, etc)

It shouldn't lead to an incorrect bundle being pushed upstream and we're safe there I think, but it needs fixing as a matter of urgency in case our workflows or the bundle changes in the future.


erikgb (commented):
/lgtm

PR #356, opened 2024-05-21
Deduplicate code for syncing target configmaps and secrets

erikgb:
Currently, the code for syncing target configmaps and secrets is separate - even if the logic should be the same. This doesn't make it tempting to add new features to bundle targets. In this PR I made an attempt to duplicate some of the code. We could probably do more, but let me know what you think before I spend more time on this! ☺️

### DIFF
--- a/pkg/bundle/bundle.go
+++ b/pkg/bundle/bundle.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
@@ -201,12 +202,9 @@ func (b *bundle) reconcileBundle(ctx context.Context, req ctrl.Request) (result 
 		secretTarget    targetKind = "Secret"
 	)
 
-	type targetResource struct {
-		Kind targetKind
-		types.NamespacedName
-	}
-
-	targetResources := map[targetResource]bool{}
+	targetResources := map[targetKind]map[types.NamespacedName]bool{}
+	targetResources[configMapTarget] = map[types.NamespacedName]bool{}
+	targetResources[secretTarget] = map[types.NamespacedName]bool{}
 
 	namespaceSelector, err := b.bundleTargetNamespaceSelector(&bundle)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/bundle/bundle.go
```diff
@@ -239,169 +237,107 @@ func (b *bundle) reconcileBundle(ctx context.Context, req ctrl.Request) (result
...
 			}
 		}
 	}
 
-	// Find all old existing ConfigMap targetResources.
-	{
-		configMapList := &metav1.PartialObjectMetadataList{
+	// Find all old existing target resources.
+	for _, kind := range []targetKind{configMapTarget, secretTarget} {
+		targetLog := log.WithValues("kind", kind)
```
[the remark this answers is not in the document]
inteon: ( I think )

pkg/bundle/bundle.go
```diff
@@ -201,12 +202,9 @@ func (b *bundle) reconcileBundle(ctx context.Context, req ctrl.Request) (result
...
 
-	type targetResource struct {
-		Kind targetKind
-		types.NamespacedName
-	}
-
-	targetResources := map[targetResource]bool{}
+	targetResources := map[targetKind]map[types.NamespacedName]bool{}
+	targetResources[configMapTarget] = map[types.NamespacedName]bool{}
+	targetResources[secretTarget] = map[types.NamespacedName]bool{}
```
inteon, line 207: Why did you change this?

erikgb: Because I now use nested maps. I think it made a lot of sense for the rest of the code, but that requires initialization of the nested maps.

pkg/bundle/bundle.go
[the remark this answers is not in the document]
erikgb: Hmmm I am not sure, but it is doable with a condition. This code is somehow protected by https://github.com/cert-manager/trust-manager/blob/main/pkg/bundle/bundle.go#L178-L196, but there is a slight difference now.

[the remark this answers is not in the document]
erikgb: I think you are right. I added a condition now.

pkg/bundle/bundle.go
inteon: Sorry, that is what I meant: why are we using a nested map?

erikgb: I tried to answer that already: It does not require creating a custom type and I think it makes the loops more compact and easier to read. But I can revert to the current structure.

erikgb: Reverted the attempt to use nested maps.

pkg/bundle/bundle.go
```diff
@@ -351,54 +306,40 @@ func (b *bundle) reconcileBundle(ctx context.Context, req ctrl.Request) (result
...
+			syncFunc = func(targetLog logr.Logger, target targetResource, shouldExist bool) (bool, error) {
+				return b.syncSecretTarget(ctx, targetLog, &bundle, target.Name, target.Namespace, resolvedBundle, shouldExist)
 			}
 		}
 
-		if cmSynced || secretSynced {
+		synced, err := syncFunc(targetLog, target, shouldExist)
+		if err != nil {
+			targetLog.Error(err, "failed sync bundle to target namespace")
+			b.recorder.Eventf(&bundle, corev1.EventTypeWarning, fmt.Sprintf("Sync%sTargetFailed", target.Kind), "Failed to sync target %s in Namespace %q: %s", target.Kind, target.Namespace, err)
```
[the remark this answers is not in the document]
inteon: Probably for another PR. Will approve and merge this PR first.

inteon:
Thanks @erikgb. I would like you to continue making improvements, let's do it in separate PRs.

/approve
/lgtm

PR #364, opened 2024-05-29
Use the go version specified in the Makefile tools module

inteon:
We noticed that the go version used to release trust-manager v0.10.0 did not match the go version specified in the `tools` Makefile module.

This is due to an oversight in the release.yaml GH action definition. This PR updates those yaml files to match the logic we are using in other projects (eg. approver-policy, see https://github.com/cert-manager/approver-policy/blob/main/.github/workflows/release.yaml).

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -19,6 +19,10 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
+      - id: go-version
+        run: |
+          make print-go-version >> "$GITHUB_OUTPUT"
+
       - uses: docker/login-action@v3
         with:
           registry: quay.io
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yaml
```diff
@@ -27,15 +31,16 @@ jobs:
...
-          go-version-file: go.mod
+          go-version: ${{ steps.go-version.outputs.result }}
 
       - id: release
         run: make release
 
       - uses: actions/upload-artifact@v4
         with:
-          name: trust-manager-${{ env.VERSION }}.tgz
+          name: ${{ steps.release.outputs.RELEASE_HELM_CHART_NAME }}-${{ steps.release.outputs.RELEASE_HELM_CHART_VERSION }}.tgz
```
inteon, line 41: This change is also based on https://github.com/cert-manager/approver-policy/blob/main/.github/workflows/release.yaml.

SgtCoDFish (approved):
/lgtm
/approve

Waving through, looks good! Can review any followups if needed!

PR #370, opened 2024-06-09
refactor: split bundle sync code into source and target

erikgb:
I find the current sync code (including test) too large to work effectively with improvements. In this PR I have tried to separate the source and target part of the sync. The split is probably not perfect, but it's at least a start.


inteon:
@erikgb I agree with the change.
I'm assuming you only moved code and did not make changes? Feel free to unhold if this is the case.
/approve
/lgtm
/hold

erikgb:
> I'm assuming you only moved code and did not make changes? Feel free to unhold if this is the case.

The only thing I changed was to add an import alias for `meta/v1` package.

/unhold

PR #377, opened 2024-07-07
refactor: build bundle data from relevant spec

erikgb:
This is a minor refactoring changing the input when building bundle data. Main benefit now is just simpler tests, but I am considering future changes that will benefit from this initial change.

/cc @inteon


inteon (approved):
Looking forward to the rest of the refactor.
/approve
/lgtm

Issue #222, opened 2023-11-03
[Feature] - Ability to inject a CA cert into a cert-manager managed secret resource


hawksight:
After reading through this cert-manager issue: https://github.com/cert-manager/cert-manager/issues/1571 I had a temporary bright spark, although the actual solution might not be the one suggested here.

## tl;dr

A single secret where cert-manager controlls `tls.crt` and `tls.key`, but `ca.crt` is injected from trust-manager.

## Problems

1) The pattern of a single secret containing `tls.crt`, `tls.key` and `ca.crt` seems to have become a standard for many tools mounting certificates in kubernetes.
2) There is an inconsistency that cert-manager issuers don't always provide a ca.crt and the current position is that this is no a pattern we want to promote.
3) A CA might not pass back a CA.crt, and the CA.crt might not be what the user actually wants anyway.
4) Using both cert-manager and trust-manager with a Secret + ConfigMap might sounds like a complete solution, but refer back to no1 where any number of k8s operators might not only work with 1 secret with defined keys. (redis operator, opensearch etc..)

## Goals / Themes

1) We want user to be able to adopt cert-manager and trust-manager broadly across their k8s applications
2) We want to maintain certificate issuance and trust distribution via separate processes.
3) Ease of use and automation, eg. avoid having to manually or have custom scripts to combine secrets, rename fields etc..

## Solution(s)

The main idea that struck me was off the back of [this comment](https://github.com/cert-manager/cert-manager/issues/1571#issuecomment-784979272), specifcially:

> Or even just a flag so we can manually set the ca.crt contents and cert-manager will include this in the secret as ca.crt

This seems a valid solution, could you manually insert the `ca.crt` into the secret itself. But why do this manually, trust-manager is about automating trust distribution. So what if trust-manager had a watch / trigger for `Secrets` that required a certain `Bundle` output to be inserted.

Just an idea but what about:

```yaml
kind: Secret
metadata:
  annotations:
    trust.cert-manager.io/ca-inject: "true"
    trust.cert-manager.io/bundle: "my-bundle"
    cert-manager.io/alt-names: alpha.dev.peter-fiddes-gcp.jetstacker.net
    cert-manager.io/certificate-name: alpha-acme
    ...
```

[...]

Issue #58, opened 2022-09-30
Support injection pem into an existing configmap


bakito:
It would be great if it was supported to have certificates injected into existing ConfigMaps if they are annotated, similar to the approach that ist available in OpenShift https://docs.openshift.com/container-platform/4.11/networking/configuring-a-custom-pki.html#certificate-injection-using-operators_configuring-a-custom-pki

Something like the snippet belo would inject the bundle to configmaps with the provided annotation.
```
target:
  configMapAnnotation:
    trust.https://cert-manager.io/inject-bundle: "true"
```

erikgb:
I like this idea and would love to see it implemented. Suggesting a small adjustment to the proposed API:

```yaml
spec:
  target:
    configMapSelector:
      my-label-key: my-label-value
```

The type of the selector should be `map[string]string` as any other Kubernetes label selector.

erikgb:
I have a suggestion to the UX here, which I think can increase the usability/flexibility. What do you think of just having a well-known annotation/label key that you could add to any configmap or secret. Assuming the name of the trust-manager Bundle is `my-ca-bundle`, this could be as easy as:

````yaml
apiVersion: v1
kind: ConfigMap
metadata:
  labels:
    trust.cert-manager.io/inject-bundle: my-ca-bundle
  name: ca-inject 
  namespace: apache
````

erikgb:
I would like to see this feature implemented, and I think I know how it could be done.

/assign

erikgb:
I discussed this issue with the cert-manager maintainers on today's stand-up, and if ever implemented this will probably require a design.

/remove-help
/remove-good-first-issue

PR #378, opened 2024-07-07
refactor: establish target.Reconciler

erikgb:
This is a major refactor of the bundle target area. The main motivation is to prepare the grounds for https://github.com/cert-manager/trust-manager/issues/58 and https://github.com/cert-manager/trust-manager/issues/222.

~~The refactoring became larger because I had to move some fields and functions to avoid circular package dependencies. I would be happy to split these changes out in pre-PRs if desired.~~ Multiple pre-PRs merged now.

### DIFF
--- a/pkg/bundle/internal/target/target.go
+++ b/pkg/bundle/internal/target/target.go
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package bundle
+package target
 
 import (
 	"bytes"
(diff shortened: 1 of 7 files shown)


pkg/bundle/internal/target/target.go
```diff
@@ -354,31 +367,31 @@ func prepareTargetPatch[T targetApplyConfiguration[T]](target T, bundle trustapi
 		)
 }
 
-type targetData struct {
-	data       string
-	binaryData map[string][]byte
+type Data struct {
+	Data       string
+	BinaryData map[string][]byte
 }
```
SgtCoDFish, line 373: comment: Looking at this now, I think we could use better names. Obviously it's not related to this PR, but `Data.PEM` and `Data.Binary` or something would be more descriptive and avoid echoing `Data.Data` 🤔

SgtCoDFish (approved):
/lgtm
/approve

This looks a good bit cleaner to me. I like the API boundary you've established with target.Reconciler!

Thanks for this 😁

pkg/bundle/internal/target/target.go
erikgb: Agreed! I can try to fix this in one of my follow-up PRs.

inteon: Ah, the data struct is unlike the non-data struct. 😄

PR #375, opened 2024-06-30
Introduce certificate pool structure and remove multiple encode/decode process

arsenalzp:
This PR introduce certificate pool structure which is mainly used to hold Certificates.
I want to decouple PEM operations from Certificates pool ones.
Deduplication option for certificate pool was also introduced - deduplication process can be optionally disabled.
It is related to #305

PR #380, opened 2024-07-09
Sort certificates in bundles to ensure deterministic behaviour

### LINKED ISSUE #310: Provide deterministic bundle
The generated trust bundle from trust-manager should provide deterministic bundle in the sense that if the source of the CA order changes, or there are multiple CA sources from multiple secret and they are shuffled in the spec. The generated final trust bundle should always be the same with the same ordering. Mayby by alphabetic order or expiration time order or something.

jabdoa2:
fix #310

### DIFF
--- a/pkg/bundle/source.go
+++ b/pkg/bundle/source.go
@@ -23,6 +23,7 @@ import (
 	"encoding/hex"
 	"encoding/pem"
 	"fmt"
+	"slices"
 	"strings"
 
 	jks "github.com/pavlo-v-chernykh/keystore-go/v4"
@@ -114,7 +115,7 @@ func (b *bundle) buildSourceBundle(ctx context.Context, sources []trustapi.Bundl
 		return bundleData{}, fmt.Errorf("couldn't find any valid certificates in bundle")
 	}
 
-	deduplicatedBundles, err := deduplicateBundles(bundles)
+	deduplicatedBundles, err := deduplicateAndSortBundles(bundles)
 	if err != nil {
 		return bundleData{}, err
 	}
@@ -343,21 +344,19 @@ func (b *bundleData) populateData(bundles []string, formats *trustapi.Additional
 	return nil
 }
 
-// remove duplicate certificates from bundles
-func deduplicateBundles(bundles []string) ([]string, error) {
+// remove duplicate certificates from bundles and sort certificates by hash
+func deduplicateAndSortBundles(bundles []string) ([]string, error) {
 	var block *pem.Block
 
-	var certificatesHashes = make(map[[32]byte]struct{})
-	var dedupCerts []string
+	var certificatesHashes = make(map[[32]byte]string)
 
 	for _, cert := range bundles {
 		certBytes := []byte(cert)
 
-	LOOP:
 		for {
 			block, certBytes = pem.Decode(certBytes)
 			if block == nil {
-				break LOOP
+				break
 			}
 
 			if block.Type != "CERTIFICATE" {
@@ -369,12 +368,23 @@ func deduplicateBundles(bundles []string) ([]string, error) {
 			// check existence of the hash
 			if _, ok := certificatesHashes[hash]; !ok {
 				// neew to trim a newline which is added by Encoder
-				dedupCerts = append(dedupCerts, string(bytes.Trim(pem.EncodeToMemory(block), "\n")))
-				certificatesHashes[hash] = struct{}{}
+				certificatesHashes[hash] = string(bytes.Trim(pem.EncodeToMemory(block), "\n"))
 			}
 		}
-
 	}
 
-	return dedupCerts, nil
+	var orderedKeys [][32]byte
+	for key := range certificatesHashes {
+		orderedKeys = append(orderedKeys, key)
+	}
+	slices.SortFunc(orderedKeys, func(a, b [32]byte) int {
+		return bytes.Compare(a[:], b[:])
+	})
+
+	var sortedDeduplicatedCerts []string
+	for _, key := range orderedKeys {
+		sortedDeduplicatedCerts = append(sortedDeduplicatedCerts, certificatesHashes[key])
+	}
+
+	return sortedDeduplicatedCerts, nil
 }
--- a/pkg/bundle/source_test.go
+++ b/pkg/bundle/source_test.go
@@ -61,7 +61,7 @@ func Test_buildSourceBundle(t *testing.T) {
 				{InLine: ptr.To(dummy.TestCertificate1 + "\n" + dummy.TestCertificate2 + "\n\n")},
 			},
 			objects:          []runtime.Object{},
-			expData:          dummy.JoinCerts(dummy.TestCertificate1, dummy.TestCertificate2),
+			expData:          dummy.JoinCerts(dummy.TestCertificate2, dummy.TestCertificate1),
 			expError:         false,
 			expNotFoundError: false,
 		},
@@ -98,7 +98,20 @@ func Test_buildSourceBundle(t *testing.T) {
 				ObjectMeta: metav1.ObjectMeta{Name: "configmap"},
 				Data:       map[string]string{"key": dummy.TestCertificate1 + "\n" + dummy.TestCertificate2},
 			}},
-			expData:          dummy.JoinCerts(dummy.TestCertificate1, dummy.TestCertificate2),
+			expData:          dummy.JoinCerts(dummy.TestCertificate2, dummy.TestCertificate1),
+			expError:         false,
+			expNotFoundError: false,
+		},
+		"if single ConfigMap source, return data even when order changes": {
+			// Test uses the same data as the previous one but with different order
+			sources: []trustapi.BundleSource{
+				{ConfigMap: &trustapi.SourceObjectKeySelector{Name: "configmap", KeySelector: trustapi.KeySelector{Key: "key"}}},
+			},
+			objects: []runtime.Object{&corev1.ConfigMap{
+				ObjectMeta: metav1.ObjectMeta{Name: "configmap"},
+				Data:       map[string]string{"key": dummy.TestCertificate2 + "\n" + dummy.TestCertificate1},
+			}},
+			expData:          dummy.JoinCerts(dummy.TestCertificate2, dummy.TestCertificate1),
 			expError:         false,
 			expNotFoundError: false,
 		},
@@ -111,7 +124,7 @@ func Test_buildSourceBundle(t *testing.T) {
 				ObjectMeta: metav1.ObjectMeta{Name: "configmap"},
 				Data:       map[string]string{"key": dummy.TestCertificate1},
 			}},
-			expData:          dummy.JoinCerts(dummy.TestCertificate1, dummy.TestCertificate2),
+			expData:          dummy.JoinCerts(dummy.TestCertificate2, dummy.TestCertificate1),
 			expError:         false,
 			expNotFoundError: false,
 		},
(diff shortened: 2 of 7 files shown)


jabdoa2:
Should be good to go now. I adjusted the certificate order in smoke and integration test.

jabdoa2:
@erikgb is there a chance to get this merged & released soonith? we would prefer to forward fix our issue instead of rolling everything back.

erikgb:
> @erikgb is there a chance to get this merged & released soonith? we would prefer to forward fix our issue instead of rolling everything back.

@jabdoa2 Yes, it should be possible to merge and release this improvement soonish. I am going to review it today, but I think we should get https://github.com/cert-manager/trust-manager/pull/375 merged first. The other PR changes overlapping area of code to this PR, and has been open for a while. So you'd have to be prepared for a rebase. 😉

pkg/bundle/source.go
```diff
@@ -369,12 +387,21 @@ func deduplicateBundles(bundles []string) ([]string, error) {
...
+				certificatesHashes[hash] = string(bytes.Trim(pem.EncodeToMemory(block), "\n"))
 			}
 		}
+	}
+
+	var orderedKeys [][32]byte
+	for key := range certificatesHashes {
+		orderedKeys = append(orderedKeys, key)
+	}
+	sort.Sort(ByHash(orderedKeys))
```
erikgb, line 399: Wouldn't it be simpler to just sort the slice now named `sortedDeduplicatedCerts` after it's constructed? This will avoid the custom `ByHash` sorter, as we can just use https://pkg.go.dev/sort#Strings. We don't require a particular order, it just has to be consistent.

erikgb (commented):
Thanks for looking into this, @jabdoa2! Looks (almost) good to me. 😉 As already commented, we should probably wait for https://github.com/cert-manager/trust-manager/pull/375 to get merged.

arsenalzp:
> Thanks for looking into this, @jabdoa2! Looks (almost) good to me. 😉 As already commented, we should probably wait for #375 to get merged.

You can merge it as it is then I can re-write code to remediate all remarks. I'm on vacation right now, thus no chance to get laptop until Tuesday.

erikgb:
> You can merge it as it is then I can re-write code to remediate all remarks as well as introducing `crypto/x509 CertPool` I'm on vacation right now, thus no chance to get laptop until Tuesday. I would like to stay the course.

Thanks @arsenalzp!

arsenalzp:
> > You can merge it as it is then I can re-write code to remediate all remarks as well as introducing `crypto/x509 CertPool` I'm on vacation right now, thus no chance to get laptop until Tuesday. I would like to stay the course.
> 
> Thanks @arsenalzp!

How it will look like: my PR is the first one then #380 is the second one?
Or just merge #380 then I will re-write me PR?

erikgb:
> How it will look like: my PR is the first one then #380 is the second one? Or just merge #380 then I will re-write me PR?

I think https://github.com/cert-manager/trust-manager/pull/380 (this PR) can be merged first, and it also fixes a serious issue. It looks almost GTM to me.

test/env/data.go
```diff
@@ -209,21 +210,50 @@ func CheckBundleSynced(ctx context.Context, cl client.Client, bundleName string,
...
-		remaining := strings.TrimPrefix(got, startingData)
-
+		certBytes = []byte(got)
 		// check that there are a nonzero number of valid certs remaining
+		found := false
+	LOOP2:
+		for {
+			block, certBytes = pem.Decode(certBytes)
+			if block == nil {
+				break LOOP2
```
maelvls, line 243: Can't this be replaced with `continue`? It seems like the purpose of this `break` is to jump to the next iteration of the loop. I don't see why `break LOOP2` is needed.

pkg/bundle/source_test.go
```diff
@@ -499,7 +512,7 @@ func TestBundlesDeduplication(t *testing.T) {
 		t.Run(name, func(t *testing.T) {
 			t.Parallel()
 
-			resultBundle, err := deduplicateBundles(test.bundle)
+			resultBundle, err := deduplicateAndSortBundles(test.bundle)
 
 			assert.Nil(t, err)
```
maelvls, line 518: Just below, I've noticed we don't check that the deduplicated bundle uses the same order as the expected bundle:

```go
assert.ElementsMatch(t, test.testBundle, resultBundle)
```

Could we change this to check the order?

```go
assert.Equal(t, test.testBundle, resultBundle)
```

pkg/bundle/source.go
jabdoa2: That would potentially move more memory around. This way we keep the amount of memcopy to a minimum. We would also have to rehash for every compare. My code is inspired by the golang book on how to iterate a hashmap in a deterministic order. Personally I do not like those custom sorters either but that is how go imagines how this should work.

I can change it but it will probably perform worse.

test/env/data.go
jabdoa2: Probably. I just copied the almost identical deduplication code.

pkg/bundle/source_test.go
jabdoa2: Yeah I noticed that. That is why I put my new test in the test suite above that which actually asserts the order.

test/env/data.go
jabdoa2: I just checked. This needs to be break. This will consume the buffer and the loop has to stop once it ran out of bytes.

pkg/bundle/source.go
jabdoa2: Ah I think I now understand what you mean. You want to sort the buffer without hashing. Just the strings. Guess that might work as well. Since we hash anyway for deduplication the current way is quite cheap. But yeah I see.

jabdoa2: This would be the alternative implementation:

```
// remove duplicate certificates from bundles and sort certificates by hash
func deduplicateAndSortBundles(bundles []string) ([]string, error) {
	var block *pem.Block

	var certificatesHashes = make(map[[32]byte]struct{})
	var dedupCerts []string

	for _, cert := range bundles {
		certBytes := []byte(cert)

	LOOP:
		for {
			block, certBytes = pem.Decode(certBytes)
			if block == nil {
				break LOOP
			}

			if block.Type != "CERTIFICATE" {
				return nil, fmt.Errorf("couldn't decode PEM block containing certificate")
			}

			// calculate hash sum of the given certificate
			hash := sha256.Sum256(block.Bytes)
			// check existence of the hash
			if _, ok := certificatesHashes[hash]; !ok {
				// neew to trim a newline which is added by Encoder
				dedupCerts = append(dedupCerts, string(bytes.Trim(pem.EncodeToMemory(block), "\n")))
				certificatesHashes[hash] = struct{}{}
			}
		}

	}

	sort.Strings(dedupCerts)

	return dedupCerts, nil
}
```

Interestingly, the order is the same as with the hashed version. It breaks two tests though and I do not understand exactly why.

test/env/data.go
jabdoa2: It probably is. I will change that.

jabdoa2: Changed it in my (copied) version and in the original place where I got it from.

pkg/bundle/source.go
erikgb: I think it's worth it - for the sake of readability. But I'll leave it up to maintainers to decide. 😉

pkg/bundle/source_test.go
erikgb: I agree with @maelvls that we should assert the order here as well - if it does require too many additional changes.

jabdoa2: ok i changed that test as well

pkg/bundle/source.go
```diff
@@ -343,21 +344,37 @@ func (b *bundleData) populateData(bundles []string, formats *trustapi.Additional
 	return nil
 }
 
-// remove duplicate certificates from bundles
-func deduplicateBundles(bundles []string) ([]string, error) {
+// ByHash is a type alias for sorting the slice of [32]byte arrays
+type ByHash [][32]byte
```
erikgb, line 348:
```suggestion
type byHash [][32]byte
```

I don't like this type, but if we decide to keep it, I don't think it needs to be exported.

pkg/bundle/source.go
erikgb: WDYT @inteon?

pkg/bundle/source.go
jabdoa2: done

pkg/bundle/source.go
inteon: Can we keep the current code and use `slices.SortFunc(result, bytes.Compare)`?

inteon: see https://github.com/cert-manager/trust-manager/pull/380/commits/a61c915f7aed998df005700e68c6779e73ba5ebb

erikgb:
/lgtm

We should maybe squash commits before/on merge?

pkg/bundle/source.go
jabdoa2: nice solution!

jabdoa2:
> /lgtm
> 
> We should maybe squash commits before/on merge?

I squashed everything into one commit

erikgb (approved):
Thanks for fixing this serious issue, @jabdoa2! 🚀 

/lgtm
/approve

maelvls:
I spent Friday afternoon manually testing the bug with label selectors. I was able to reproduce it, and I was able to test that your PR fixes the issue. I have documented the testing I did in the Hackmd page [PR 380: Manually Testing that ConfigMaps are No Longer Unexpectedly Updated By The ConfigMap Label Selector Bug](https://hackmd.io/@maelvls/BJ1ow_0P0).

By the way, would there be by any chance someone relying the fact that the ordering in `sources` is identical in `targets`?

erikgb:
> By the way, would there be by any chance someone relying the fact that the ordering in `sources` is identical in `targets`?

Order doesn't matter here IMO. The order must be consistent, but I don't think a particular order is required.

jabdoa2:
> > By the way, would there be by any chance someone relying the fact that the ordering in `sources` is identical in `targets`?
> 
> Order doesn't matter here IMO. The order must be consistent, but I don't think a particular order is required.

This should be very unlikely for trust bundles. Those are not usually ordered and order should not matter. 

The story would be different for PEMs containing a certificate chain. For instance, OpenSSL is well known for behaving badly if the first cert is not the leaf cert. Most other libraries will behave well though. Also that is not the usecase for trust-manager.

maelvls:
> Order doesn't matter here IMO. The order must be consistent, but I don't think a particular order is required.

We just found out that order matters for Java clients when there are intermediate certificates in the bundle: https://github.com/cert-manager/trust-manager/issues/419

sagarmujumale:
> > By the way, would there be by any chance someone relying the fact that the ordering in `sources` is identical in `targets`?
> 
> Order doesn't matter here IMO. The order must be consistent, but I don't think a particular order is required.

The Validation Process

When a client verifies a server's certificate, it follows these steps:

    Starts with the leaf certificate: The client uses the public key in the leaf certificate to verify the signature of the intermediate certificate.
    Verifies intermediate certificates: The client iterates through the chain, using each certificate's public key to verify the signature of the next certificate.
    Trusts the root certificate: The final step is to verify the signature of the last intermediate certificate using the public key of the trusted root certificate.

Consequences of Incorrect Order

If the certificate chain is not in the correct order, the validation process will fail:

    Invalid signature: The client will be unable to verify the signature of a certificate using the wrong public key.
    Broken trust chain: The chain of trust will be incomplete, leading to certificate validation failure.

In summary, the correct order of certificates in a chain is essential for establishing trust and ensuring secure communication.

erikgb:
@sagarmujumale I think you are mixing certificate chains and trust bundles.

jabdoa2:
This absolutely matters for certificate chains in some TLS libs (i.e. OpenSSL). However, trust-manager is designed for root of trust (instead of chains).

PR #385, opened 2024-07-15
Add flags to configure LeaseDuration and RenewDeadline

jabdoa2:
This change will prevent trust-manager from restart looping when kube API server is (over-)loaded. Most kubernetes operators increase this value as the default (15s) is too low under load.

### DIFF
--- a/cmd/trust-manager/app/app.go
+++ b/cmd/trust-manager/app/app.go
@@ -80,6 +80,8 @@ func NewCommand() *cobra.Command {
 				LeaderElection:                true,
 				LeaderElectionID:              "trust-manager-leader-election",
 				LeaderElectionReleaseOnCancel: true,
+				LeaseDuration:                 &opts.LeaseDuration,
+				RenewDeadline:                 &opts.RenewDeadline,
 				ReadinessEndpointName:         opts.ReadyzPath,
 				HealthProbeBindAddress:        fmt.Sprintf("0.0.0.0:%d", opts.ReadyzPort),
 				WebhookServer: ctrlwebhook.NewServer(ctrlwebhook.Options{
--- a/cmd/trust-manager/app/options/options.go
+++ b/cmd/trust-manager/app/options/options.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"log/slog"
 	"os"
+	"time"
 
 	"github.com/go-logr/logr"
 	"github.com/spf13/cobra"
@@ -63,6 +64,12 @@ type Options struct {
 
 	// log are options controlling logging
 	log logOptions
+
+	// Leader election lease duration
+	LeaseDuration time.Duration
+
+	// Leader election lease renew duration
+	RenewDeadline time.Duration
 }
 
 type logOptions struct {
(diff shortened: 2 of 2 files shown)


erikgb:
Thanks, @jabdoa2! I think it makes sense to allow these values to be configured. WDYT?

jabdoa2:
> Thanks, @jabdoa2! I think it makes sense to allow these values to be configured. WDYT?

done.

cmd/trust-manager/app/options/options.go
```diff
@@ -187,6 +194,14 @@ func (o *Options) addAppFlags(fs *pflag.FlagSet) {
 		"readiness-probe-path", "/readyz",
 		"HTTP path to expose the readiness probe server.")
 
+	fs.DurationVar(o.LeaseDuration,
+		"lease-duration", time.Second*60,
```
erikgb, line 198:
```suggestion
		"leader-election-lease-duration", time.Second*60,
```

To give more context to user-facing flags.

cmd/trust-manager/app/options/options.go
```diff
@@ -187,6 +194,14 @@ func (o *Options) addAppFlags(fs *pflag.FlagSet) {
...
+		"Lease duration for leader election")
+
+	fs.DurationVar(o.RenewDeadline,
+		"renew-deadline", time.Second*40,
```
erikgb, line 202:
```suggestion
		"leader-election-renew-deadline", time.Second*40,
```

To give more context to user-facing flags.

erikgb (changes requested):
I think the flags should be given a bit more context. The suggested flag names are consistent with the equivalent flags in cert-manager.

cmd/trust-manager/app/options/options.go
inteon, line 198: Let's start with making these durations configurable without changing the defaults.
In another PR, we can increase these values and discuss if that makes sense there.

jabdoa2: Ok will change that. Those defaults are really bad. Even very mainstream controller (i.e. nginx-ingress) increase this by default (nginx-ingress does 30s renew for example).

cmd/trust-manager/app/options/options.go
```diff
@@ -187,6 +194,14 @@ func (o *Options) addAppFlags(fs *pflag.FlagSet) {
 		"readiness-probe-path", "/readyz",
 		"HTTP path to expose the readiness probe server.")
 
+	fs.DurationVar(o.LeaseDuration,
+		"leader-election-lease-duration", time.Second*60,
```
erikgb, line 198:
```suggestion
		"leader-election-lease-duration", time.Second*15,
```

cmd/trust-manager/app/options/options.go
```diff
@@ -187,6 +194,14 @@ func (o *Options) addAppFlags(fs *pflag.FlagSet) {
...
+		"Lease duration for leader election")
+
+	fs.DurationVar(o.RenewDeadline,
+		"leader-election-renew-deadline", time.Second*40,
```
erikgb, line 202:
```suggestion
		"leader-election-renew-deadline", time.Second*10,
```

erikgb (commented):
/approve

Changes looks good to me. Can you try to figure out why CI is failing and squash commits, please @jabdoa2?

jabdoa2:
> /approve
> 
> Changes looks good to me. Can you try to figure out why CI is failing and squash commits, please @jabdoa2?

Test is fixed and commits are squashed. Should be good to go.

cmd/trust-manager/app/app.go
```diff
@@ -74,12 +75,16 @@ func NewCommand() *cobra.Command {
 			eventBroadcaster.StartLogging(func(format string, args ...any) { mlog.V(3).Info(fmt.Sprintf(format, args...)) })
 			eventBroadcaster.StartRecordingToSink(&clientv1.EventSinkImpl{Interface: cl.CoreV1().Events("")})
 
+			renewDeadline := time.Second * 10
+			leaseDuration := time.Second * 15
```
erikgb, line 79: While the tests are now green, we are not using the new flags. 😁 Can you try to fix this please, @jabdoa2?

jabdoa2: Ups. Left over from my tests. Will fix that. Sorry

erikgb (approved):
Thanks, @jabdoa2! 🚀 

/lgtm
/approve

PR #388, opened 2024-07-17
Expose leader election leaseDuration and renewDeadline in helm chart

jabdoa2:


erikgb:
/ok-to-test

@jabdoa2, if you want to avoid the wait to get CI runs approved on each of your PRs, you should consider becoming a member of the cert-manager org! I'll be happy to sponsor your membership, and the process is described in https://github.com/cert-manager/community/blob/main/GOVERNANCE.md#github-member. 😸

inteon:
Please run "make generate" to generate the jsonschema used by the Helm chart.

jabdoa2:
> Please run "make generate" to generate the jsonschema used by the Helm chart.

Done. Thanks for the hint!

deploy/charts/trust-manager/values.yaml
```diff
@@ -182,6 +182,13 @@ app:
   # The verbosity of trust-manager logging. This takes a value from 1-5, with the higher value being more verbose.
   logLevel: 1
 
+  # Leader Election Configuration
```
erikgb, line 185:
```suggestion
```

Nit: Seems a bit superfluous.

deploy/charts/trust-manager/values.yaml
```diff
@@ -182,6 +182,13 @@ app:
...
+  leaderElection:
+    # How long will a lease be valid? You probably want to increase this to 60-120s.
```
erikgb, line 187:
```suggestion
    # The duration that non-leader candidates will wait to force acquire leadership.
    # The default should be sufficient in a healthy cluster but can be slightly increased to prevent trust-manager from restart-looping when the API server is overloaded.
```

deploy/charts/trust-manager/values.yaml
```diff
@@ -182,6 +182,13 @@ app:
...
+    leaseDuration: 15s
+    # How long will a controller try to renew a lease before stepping down? You probably want to increase this to 30-45s.
```
erikgb, line 189:
```suggestion
    # The interval between attempts by the acting leader to renew a leadership slot before it stops leading.
    # This MUST be less than or equal to the lease duration.
    # The default should be sufficient in a healthy cluster but can be slightly increased to prevent trust-manager from restart-looping when the API server is overloaded.
```

erikgb (changes requested):
Thanks for making these new options available to the Helm chart, @jabdoa2! I think some of your value descriptions are a bit biased, so I have proposed some more informal ones.

Can you also please rework the PR title and add a short description?

deploy/charts/trust-manager/values.yaml
SgtCoDFish: I agree with Erik's suggestions here so I've added them to a commit and pushed it 👍

erikgb (approved):
Thanks! 🚀 

/lgtm
/approve

PR #394, opened 2024-07-18
refactor: move default truststore passwords to API

erikgb:
This minor refactoring is extracted from https://github.com/cert-manager/trust-manager/pull/378, and the motivation is to solve circular package dependency issues.

Moving the constants to the API package seems like a good option, as the default passwords used for truststores is highly user-facing.

/cc @inteon

### DIFF
--- a/docs/api/api.md
+++ b/docs/api/api.md
@@ -58,10 +58,19 @@ import "github.com/cert-manager/trust-manager/pkg/apis/trust/v1alpha1"
 
 ## Constants
 
-<a name="BundleConditionSynced"></a>
+<a name="DefaultJKSPassword"></a>
 
 ```go
 const (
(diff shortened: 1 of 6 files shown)


docs/api/api.md
```diff
@@ -58,10 +58,19 @@ import "github.com/cert-manager/trust-manager/pkg/apis/trust/v1alpha1"
 
 ## Constants
 
-<a name="BundleConditionSynced"></a>
```
SgtCoDFish, line 61: I guess this anchor link is just wrong, right? Seems to be a gomarkdoc bug - this should surely be "constants" or something similar, not the name of one of the constants.

Not a blocker obviously, just interesting!

SgtCoDFish (approved):
/lgtm
/approve

Thanks for this, fully agree with these changes!

PR #396, opened 2024-07-19
refactor: extract truststore encoding to internal package

erikgb:
This is another refactoring preparing for https://github.com/cert-manager/trust-manager/pull/378 - to avoid dependency issues. The proposed changes move all types to encode truststores into a dedicated (internal) package.


SgtCoDFish (approved):
/lgtm
/approve

Thanks, I think this'll be the last thing before the release 😁

PR #397, opened 2024-07-20
refactor: move FieldManager to more available package

erikgb:
With some additional improvements that made sense when doing this. The main motivation is to reduce the size of https://github.com/cert-manager/trust-manager/pull/378, as I need the `FieldManager` constant available in a new package without getting circular dependencies.

The package name `ssa_client` annoys me a bit. 😉 I think the name should be just `client`. I was tempted to change this, but let's leave it for now - as it will require us to resolve package name import conflicts. It should be possible to do by creating type aliases in our `client` package. I've seen that c/r does this a lot.

### DIFF
--- a/pkg/bundle/internal/ssa_client/patch.go
+++ b/pkg/bundle/internal/ssa_client/patch.go
@@ -21,6 +21,10 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const (
+	FieldManager = client.FieldOwner("trust-manager")
+)
+
 type applyPatch struct {
 	patch []byte
 }
--- a/pkg/bundle/bundle.go
+++ b/pkg/bundle/bundle.go
@@ -29,7 +29,6 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
-	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/client-go/util/csaupgrade"
 	"k8s.io/utils/clock"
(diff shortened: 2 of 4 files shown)


pkg/bundle/internal/ssa_client/patch.go
```diff
@@ -17,10 +17,101 @@ limitations under the License.
...
 	"k8s.io/apimachinery/pkg/types"
+	coreapplyconfig "k8s.io/client-go/applyconfigurations/core/v1"
+	v1 "k8s.io/client-go/applyconfigurations/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	trustapi "github.com/cert-manager/trust-manager/pkg/apis/trust/v1alpha1"
+)
+
+const (
+	FieldManager = client.FieldOwner("trust-manager")
```
erikgb, line 33: This constant seems a bit "misplaced", but I didn't want to create a new Go-file for just this...

pkg/bundle/internal/ssa_client/patch.go
```diff
@@ -17,10 +17,101 @@ limitations under the License.
...
 )
 
+type bundleStatusApplyConfiguration struct {
+	v1.TypeMetaApplyConfiguration    `json:",inline"`
+	*v1.ObjectMetaApplyConfiguration `json:"metadata,omitempty"`
+	Status                           *trustapi.BundleStatus `json:"status,omitempty"`
+}
+
+func GenerateBundleStatusPatch(
```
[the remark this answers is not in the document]
inteon: I think it already lived in this go module.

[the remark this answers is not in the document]
erikgb: Agreed, done.

pkg/bundle/internal/ssa_client/patch.go
```diff
@@ -17,10 +17,65 @@ limitations under the License.
...
+
+	encodedPatch, err := json.Marshal(configmapPatch)
+	if err != nil {
+		return configmap, nil, err
+	}
+
+	return configmap, applyPatch{encodedPatch}, nil
+}
+
+func GenerateSecretPatch(
```
inteon, line 56: @erikgb same here, can this be moved back to reduce diff?

erikgb: Done

inteon (commented):
/approve
/lgtm

PR #401, opened 2024-07-31
Fix example of topologySpreadConstraints to reference trust-manager

SgtCoDFish:
Spotted by @wallrj in https://github.com/cert-manager/istio-csr/pull/365

### DIFF
--- a/deploy/charts/trust-manager/values.yaml
+++ b/deploy/charts/trust-manager/values.yaml
@@ -167,8 +167,7 @@ tolerations: []
 #     whenUnsatisfiable: ScheduleAnyway
 #     labelSelector:
 #       matchLabels:
-#         app.kubernetes.io/instance: cert-manager
-#         app.kubernetes.io/component: controller
+#         app.kubernetes.io/name: trust-manager
 topologySpreadConstraints: []
 
 filterExpiredCertificates:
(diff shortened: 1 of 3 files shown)


deploy/charts/trust-manager/values.yaml
```diff
@@ -167,8 +167,7 @@ tolerations: []
 #     whenUnsatisfiable: ScheduleAnyway
 #     labelSelector:
 #       matchLabels:
-#         app.kubernetes.io/instance: cert-manager
-#         app.kubernetes.io/component: controller
+#         app.kubernetes.io/name: trust-manager
```
wallrj, line 170: This single label selector is equivalent but not identical to the single label selector in the Deployment selector:
 * https://github.com/cert-manager/trust-manager/blob/ee9a7b61cf66fcef81b0ac4fa4e99b9c7c3f4a62/deploy/charts/trust-manager/templates/deployment.yaml#L11-L12

I guess its better practice to use this namespaced  label.

And I guess the reason for removing the `app.kubernetes.io/instance` label is that there must only be one instance of trust-manager in each cluster.

wallrj (approved):
Thanks!

/approve
/lgtm
/hold so that you can read my comment

deploy/charts/trust-manager/values.yaml
SgtCoDFish: Honestly, I didn't think that deeply on this. Topology spread constraints are pretty deep into K8s magic land, and I reckon anyone wanting to set them is probably pretty comfortable with label selectors and picking what they want - I just wanted an example which mentioned trust-manager, really!

SgtCoDFish:
/unhold

I think the question has been seen!

PR #408, opened 2024-08-14
test: remove dead event recorder code

erikgb:
Noticed some dead code while working with https://github.com/cert-manager/trust-manager/pull/378. Also removing `fakeclient` import alias to avoid clash with variable.

/cc @inteon


pkg/bundle/target_test.go
```diff
@@ -686,13 +682,6 @@ func Test_syncConfigMapTarget(t *testing.T) {
 					assert.Equal(t, pkcs12Data, binData)
 				}
 			}
-
-			var event string
-			select {
-			case event = <-fakerecorder.Events:
-			default:
-			}
-			assert.Equal(t, test.expEvent, event)
```
erikgb, line 695: Or should we keep this to assert no events?

PR #412, opened 2024-08-15
refactor: fix name and import alias collisions

erikgb:
When an identifier name collides with an import alias, my IDE gets confused, making me less effective when coding. This PR fixes all current collisions, but I wonder if golangci-lint might have a linter that can detect this? For one of the collisions, I found it more convenient to fix this by extracting a reusable function avoiding duplicated code.

### DIFF
--- a/pkg/bundle/controller.go
+++ b/pkg/bundle/controller.go
@@ -130,14 +130,8 @@ func AddBundleController(
 		// Reconcile Bundles who reference a modified source ConfigMap.
 		Watches(&corev1.ConfigMap{}, b.enqueueRequestsFromBundleFunc(
 			func(obj client.Object, bundle trustapi.Bundle) bool {
-				for _, source := range bundle.Spec.Sources {
-					if source.ConfigMap == nil {
-						continue
-					}
-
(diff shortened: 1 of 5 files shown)


pkg/bundle/controller.go
```diff
@@ -148,18 +142,8 @@ func AddBundleController(
 		// Reconcile Bundles who reference a modified source Secret.
 		Watches(&corev1.Secret{}, b.enqueueRequestsFromBundleFunc(
 			func(obj client.Object, bundle trustapi.Bundle) bool {
-				for _, source := range bundle.Spec.Sources {
-					if source.Secret == nil {
-						continue
-					}
-
-					if source.Secret.Selector != nil {
```
inteon, line 156: Is it ok to not check `source.Secret.Selector != nil` anymore?

erikgb: Isn't this handled within `sourceSelectsObject` function now? Or have I overlooked something?

SgtCoDFish (approved):
/lgtm
/approve

Thanks, makes sense (and it does look like your function checks it correctly as far as I can tell?)

I would love a linter that checks this, I think this causes a lot of slowdown in `vim-go` 🤔

PR #413, opened 2024-08-15
refactor: move function for creating managed field entries

erikgb:
Extracted from https://github.com/cert-manager/trust-manager/pull/378 - which should show the motivation for this refactoring: I want to extract the code for handling configmap/secret targets out of the huge bundle package. The test utility function moved in this PR is used in tests that I want to move, but also in tests that should remain in the package.

If you think there is a better location (package) for this test utility code, please let me know.

### DIFF
--- a/pkg/bundle/bundle_test.go
+++ b/pkg/bundle/bundle_test.go
@@ -40,6 +40,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 
 	trustapi "github.com/cert-manager/trust-manager/pkg/apis/trust/v1alpha1"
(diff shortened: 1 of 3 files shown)


SgtCoDFish (approved):
/lgtm
/approve

If this is only ever to be used in tests I'd consider putting it under `test/util` or something but since this is in an internal package it really doesn't matter! Thanks!

erikgb:
@SgtCoDFish I have now resolved the conflicts (and added a missing import alias). Could you take another look, please? :-)

SgtCoDFish (approved):
/lgtm
/approve

Thanks 😁

PR #414, opened 2024-08-15
Bump sigs.k8s.io/controller-runtime to 0.19.0

erikgb:
It seems like https://github.com/cert-manager/trust-manager/pull/411 needs some help after controller-runtime added validation of unique controller names in https://github.com/kubernetes-sigs/controller-runtime/pull/2902. I had a look at creating the controller in BeforeSuite, but that would potentially require a lot of changes. So the simplest solution seems to be to skip the new validation in our integration tests, ref. https://github.com/kubernetes-sigs/controller-runtime/pull/2918.


inteon:
/approve
/lgtm
Thanks for the upgrade!

PR #415, opened 2024-08-15
refactor: dedicated target data struct

erikgb:
Extracted from https://github.com/cert-manager/trust-manager/pull/378, and obtaining a cleaner separation of source and target is required to move the target logic to a dedicated package.


SgtCoDFish (approved):
/lgtm
/approve

Seems fine as a stepping stone to something else. Thanks!

Issue #407, opened 2024-08-14
Support other ways to generate service TLS certificate in the Helm chart


anael-l:
Hi,
This project is exactly what I need for managing internal CAs in Openshift clusters.

I don’t use cert-manager, and ideally, I would like to be able to install trust-manager without it.
As of https://github.com/cert-manager/trust-manager/pull/157 , I see this has been addressed, but the doc advises against using this in production.

One of Openshift’s operator allows for generating service certificates in a secret by annotating a secret, signed by the cluster’s CA:
https://docs.openshift.com/container-platform/4.15/security/certificates/service-serving-certificate.html#add-service-certificate_service-serving-certificate
It also can inject a CA bundle into the webhook: https://docs.openshift.com/container-platform/4.15/security/certificates/service-serving-certificate.html#add-service-certificate-validating-webhook_service-serving-certificate

Right now, the Helm chart only allows to either use the Helm generated certificate, or the cert-manager one.
Would you be willing to have another option to use the Openshift managed certificates?

I can understand that it’s not wanted to have a solution that is specific to a platform. In that case, would you consider having an option to not manage the certificate in the Chart, and let the user configure it another way?
I could then add the needed annotations the service and webhook.

Thanks

erikgb:
I prefer avoiding a solution tailored **for** OpenShift, even if we also run OpenShift in our target clusters. But a more generic extension mechanism for webhook certs seems like a valid ask. @anael-l, do you know if OpenShift supports injecting CA certs into CRDs for conversion webhooks? We will probably add new API versions that will require conversion webhooks.

To use the OpenShift mechanism to do this with the current webhook (validating), I think two things are needed:
1. Bring your own secret for the webhook server certificate.
2. Add annotations to webhook configuration resources.

anael-l:
I've never used it, but this seems to be what you are looking for: https://docs.openshift.com/container-platform/4.15/security/certificates/service-serving-certificate.html#add-service-certificate-crd_service-serving-certificate 

1. For the secret, I think it is already possible to do, as the `Deployment` mounts a secret named `{{ include "trust-manager.name" . }}-tls`.
Maybe we could add a variable to specify another secret name.
That could be tied to a condition that would not create a secret with Helm nor a Certificate with cert-manager.

[...]

PR #420, opened 2024-08-19
feat(helm): Allow external TLS provider for webhook

anael-l:
closes: https://github.com/cert-manager/trust-manager/issues/407

Added a new parameter  `app.webhook.tls.certManager.enabled=true` that keeps the default config to create certmanager certificates, but that can be disabled.

Added new parameters `app.webhook.annotations` and `app.webhook.service.annotations` to set annotations for the `ValidatingWebhookConfiguration` and Webhook `Service` respectively.

This allows for having another way to provide TLS certificates like for instance wih the [Openshift service serving certificates](https://docs.openshift.com/container-platform/4.15/security/certificates/service-serving-certificate.html)

Example values:

```yaml
app:
  webhook:
    annotations:
      service.beta.openshift.io/inject-cabundle: "true"
    service:
      annotations:
        service.beta.openshift.io/serving-cert-secret-name: '{{ include "trust-manager.name" . }}-tls'
    tls:
      helmCert:
        enabled: false
      certManager:
        enabled: false
```

### DIFF
--- a/deploy/charts/trust-manager/templates/certificate.yaml
+++ b/deploy/charts/trust-manager/templates/certificate.yaml
@@ -1,4 +1,4 @@
-{{- if not .Values.app.webhook.tls.helmCert.enabled -}}
+{{- if .Values.app.webhook.tls.certManager.enabled -}}
 
 apiVersion: cert-manager.io/v1
 kind: Issuer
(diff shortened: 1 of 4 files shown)


deploy/charts/trust-manager/templates/certificate.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if not .Values.app.webhook.tls.helmCert.enabled -}}
+{{- if .Values.app.webhook.tls.certManager.enabled -}}
```
erikgb, line 1: This appears like a breaking change. It it intentional? If possible, I think we should still support the `app.webhook.tls.helmCert.enabled` Helm value.

anael-l: the `app.webhook.tls.helmCert.enabled` still works, but for for existing installation, the `app.webhook.tls.certManager.enabled` var would have to be set to false, which indeed is a breaking change.

Should we have instead a `app.webhook.tls.externalProvider.enabled=false` var that when enabled, disables the cert-manager certficates creation ?

erikgb: I am not sure if this will work, but could it be an option to add a new optional Helm value to set the name of the webhook certificate secret? And when user sets this value, all use of cert-manager or Helm to generate a webhook certificate is turned off?

PR #424, opened 2024-08-26
add option for deployment annotations to helm chart

a-hat:
Hi!

This PR adds the option to add annotations to the trust-manager deployment. In our specific use case we need it to apply custom annotations to exclude the deployment from kyverno policy validation.

The cert-manager chart allows the same [configuration option](https://github.com/cert-manager/cert-manager/blob/7d797a45d7040eaad420d04dc829e2ae9457bb35/deploy/charts/cert-manager/values.yaml#L325-L327), but for trust-manager it is missing.

I have seen that https://github.com/cert-manager/trust-manager/pull/214 contains the same change, but from what I understand only the certificate annotations are under discussion, so I hope my PR can get merged :)

### DIFF
--- a/deploy/charts/trust-manager/README.md
+++ b/deploy/charts/trust-manager/README.md
@@ -366,6 +366,13 @@ Pod labels to add to trust-manager pods.
 > ```
 
 Pod annotations to add to trust-manager pods.
+#### **app.deploymentAnnotations** ~ `object`
+> Default value:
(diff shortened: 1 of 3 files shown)


a-hat:
@erikgb could you kindly take a look?

erikgb:
@a-hat This new feature seems ok to merge when it passes CI. 😄

Issue #426, opened 2024-08-29
Do not throw error when matchLabels do not find any secret/configmap


Shawcs:
Hi team,

when we specify a selector for configMap or Secret source like this:

```yaml
metadata:
  name: ca-bundle
spec:
  sources:
    - configMap:
        key: ca.crt
        selector:
          matchLabels:
            trust-bundle.certs: includes
target:
    configMap:
      key: ca-bundle.crt
    secret:
      key: ca-bundle.crt
```

If no configMap matches this label selector the operator throw this kind of error:

``` shell
time=2024-08-29T13:40:51.076Z level=ERROR msg="failed to build source bundle" bundle=ca-bundle logger=trust/bundle err="invalid PEM data in source: bundle contains no PEM certificates"
time=2024-08-29T13:40:51.077Z level=ERROR msg="Reconciler error" controller=bundles namespace="" name=ca-bundle reconcileID=687362c5-ada0-456b-b9ff-8e1314938787 logger=trust/manager err="failed to build bundle source: invalid PEM data in source: bundle contains no PEM certificates"
time=2024-08-29T13:40:51.077Z level=DEBUG+3 msg="Failed to build bundle sources: invalid PEM data in source: bundle contains no PEM certificates" logger=trust/manager/events type=Warning object="{Kind:Bundle Namespace: Name:ca-bundle UID:fd490c64-5da0-47ea-add8-fdbb7e756daf APIVersion:trust.cert-manager.io/v1alpha1 ResourceVersion:309025552 FieldPath:}" reason=SourceBuildError
time=2024-08-29T13:40:51.077Z level=DEBUG+3 msg="Failed to build bundle sources: invalid PEM data in source: bundle contains no PEM certificates" logger=trust/manager/events type=Warning object="{Kind:Bundle Namespace: Name:ca-bundle UID:fd490c64-5da0-47ea-add8-fdbb7e756daf APIVersion:trust.cert-manager.io/v1alpha1 ResourceVersion:309025552 FieldPath:}" reason=SourceBuildError
```
---

### Proposition

if we don't find any ConfigMap or Secret matching the label we just ignore it ?

erikgb:
@Shawcs I have created https://github.com/cert-manager/trust-manager/pull/432 to address this issue, but I have a feeling it might not solve your issue. Are you expecting an empty result bundle if no sources are selected by your source selector(s)? It would be nice to know more about your overall use case!

Shawcs:
[...]

PR #432, opened 2024-09-09
fix: don't error if source selector selects no sources

erikgb:
Fixes https://github.com/cert-manager/trust-manager/issues/426?

~~We have a test that asserts an error when the bundle is empty. I doubt if this is correct, and to fix https://github.com/cert-manager/trust-manager/issues/426 properly, I think this behavior must go. We could eventually assert that a Bundle is invalid if it doesn't contain at least one source spec. But that is better done with the webhook (or API CEL expression) IMO.~~

UPDATE: After discussing this PR in today's stand-up, we decided to keep the existing (and tested) behavior where a target bundle needs to contain at least one certificate, at least for now. This means that this PR only fixes a minor bug: before this PR it would error if any source selector selects no sources. Now this is ok, but the resulting bundle must contain at least one certificate.

### DIFF
--- a/pkg/bundle/source.go
+++ b/pkg/bundle/source.go
@@ -18,6 +18,7 @@ package bundle
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strings"
 
@@ -31,6 +32,8 @@ import (
 )
 
 type notFoundError struct{ error }
+
+type selectsNothingError struct{ error }
 
 // bundleData holds the result of a call to buildSourceBundle. It contains the resulting PEM-encoded
 // certificate data from concatenating all the sources together, binary data for any additional formats and
@@ -75,6 +78,12 @@ func (b *bundle) buildSourceBundle(ctx context.Context, sources []trustapi.Bundl
 				sourceData = b.defaultPackage.Bundle
 				resolvedBundle.defaultCAPackageStringID = b.defaultPackage.StringID()
 			}
+		}
+
+		// A source selector may select no configmaps/secrets, and this is not an error.
+		if errors.As(err, &selectsNothingError{}) {
+			b.Log.Info(err.Error())
+			continue
 		}
 
 		if err != nil {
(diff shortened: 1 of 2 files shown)


erikgb:
@SgtCoDFish I think this PR is ready for another review now!

pkg/bundle/source.go
```diff
@@ -81,6 +81,12 @@ func (b *bundle) buildSourceBundle(ctx context.Context, sources []trustapi.Bundl
 			return bundleData{}, fmt.Errorf("failed to retrieve bundle from source: %w", err)
 		}
 
+		// A source selector may select no configmaps/secrets, and this is not an error.
+		// Must continue early since CertPool does not currently support adding no certificates.
+		if len(sourceData) == 0 {
+			continue
+		}
```
SgtCoDFish, line 88: suggestion: Would it be worth logging a warning in this situation? As a platform admin, I'd want to either remove a selector which selects nothing, or fix it.

SgtCoDFish (commented):
One suggestion from me; what do you think?

pkg/bundle/source.go
erikgb: I agree that a platform admin needs to understand what the operator is doing (and not doing). It seems like events are used for similar scenarios in this project. Should we use an event for this too, or just log it? With contextual logging, I am always in doubt what you exactly mean with a "logging a warning". 😉

SgtCoDFish: I'd personally have just logged it, and wouldn't have thought about events. I don't think an event would be bad, but I think the lightweight option of logging is totally fine!

erikgb: I'm now logging if a label selector matches no resources. Not having log levels feels a bit strange to me, but supported by Dave Chaney, I am proposing to log this on a normal info level.

> Eliminate the warning level, it’s either an informational message, or an error condition.

Ref. https://dave.cheney.net/tag/logging

I also introduced a custom error to eliminate the risk of other errors slipping through the empty string conditional. PTAL!

pkg/bundle/source.go
```diff
@@ -171,10 +180,10 @@ func (b *bundle) secretBundle(ctx context.Context, ref *trustapi.SourceObjectKey
 		if selectorErr != nil {
 			return "", fmt.Errorf("failed to parse label selector as Selector for Secret in namespace %s: %w", b.Namespace, selectorErr)
 		}
-		if err := b.client.List(ctx, &sl, client.MatchingLabelsSelector{Selector: selector}); apierrors.IsNotFound(err) {
```
erikgb, line 174: A list call will only return 404 if something is really wrong, ref. https://github.com/kubernetes/client-go/issues/537#issuecomment-450030586

SgtCoDFish: Awesome context, thank you!

erikgb:
> I'm not sure if I missed something but wasn't there a new check to ensure the bundle wasn't empty, so we still required a bundle to resolve to at least one cert?
> 
> Is that missing here? Was it added elsewhere? I might be missing it but it looks like this would permit an empty target?

The check is already there: https://github.com/cert-manager/trust-manager/blob/main/pkg/bundle/source.go#L90-L92, and it is now not changed in this PR. 😉

SgtCoDFish (approved):
/lgtm
/approve

Sorry for the delay, something came up!

I totally missed that check. Thanks very much for this 🙌

PR #436, opened 2024-09-10
refactor: reduce duplication in sync target functions

erikgb:
The main goal of this refactoring is to **move the deletion of target configmap/secret out of the patch functions**. The end goal is to provide functions that can be reused in the new pull-target feature I am working on. When the user creates the target resource, it does not make sense for trust-manager to delete it (as a side effect of patching it).

I have renamed some variables to avoid shadowing in https://github.com/cert-manager/trust-manager/pull/378.

Relates to https://github.com/cert-manager/trust-manager/pull/378

### DIFF
--- a/pkg/bundle/target.go
+++ b/pkg/bundle/target.go
@@ -25,6 +25,7 @@ import (
 	"strings"
 
 	"github.com/go-logr/logr"
+	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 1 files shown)


erikgb:
I have now attempted to reduce duplication even more using generics, but I am not too happy with the result. Go generics is not the generics I am used to. 😉 PTAL @inteon!

inteon (commented):
/approve
/lgtm
Thanks for the simplification!

PR #442, opened 2024-09-22
WIP: refactor: dedicated struct for building source data

erikgb:


erikgb:
@SgtCoDFish Would it be possible for you to review this PR? It's been open for "a while", and I am a bit tired of rebasing it. 😅 So I would like to know if this refactoring might be acceptable.

pkg/bundle/bundle.go
```diff
@@ -35,51 +34,28 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	"github.com/cert-manager/trust-manager/cmd/trust-manager/app/options"
```
SgtCoDFish, line 37: suggestion: I don't love the idea of depending on `./cmd` from `./pkg`.  I think that'll end in a circular dependency down the road - I'd expect `cmd` to depend on `pkg` but not the other way round.

I would vote to keep the options struct where it is in `pkg`.

(I totally accept that package structuring is a bit arbitrary but I do think it's better for the long term to not risk this kind of circular dependency creeping in)

pkg/bundle/internal/target/source.go
```diff
@@ -27,34 +27,62 @@ import (
...
+			return fmt.Errorf("must load default package successfully when default package location is set: %w", err)
+		}
+
+		b.DefaultPackage = &pkg
+
+		b.Options.Log.Info("successfully loaded default package from filesystem", "id", pkg.StringID(), "path", b.Options.DefaultPackageLocation)
+	}
+
+	return nil
+}
```
SgtCoDFish, line 76: suggestion: I think having an `Init` function introduces the risk of an uninitialized `BundleBuilder` struct being used somewhere and blowing things up.

A pattern I really like is to have a `func NewBundleBuilder(client, defaultPackage, options) (*BundleBuilder, error)` - that way, a user can't forget to call `Init` because creating the object and the init checks are done in `NewBundleBuilder`.

(That pattern is more useful when the type is unexported because then it's impossible to create a `bundleBuilder` which hasn't been initialized/validated, but it's not a requirement)

pkg/bundle/bundle.go
```diff
@@ -35,51 +34,28 @@ import (
...
 	trustapi "github.com/cert-manager/trust-manager/pkg/apis/trust/v1alpha1"
 	"github.com/cert-manager/trust-manager/pkg/bundle/internal/ssa_client"
 	"github.com/cert-manager/trust-manager/pkg/bundle/internal/target"
-	"github.com/cert-manager/trust-manager/pkg/fspkg"
 )
 
-// Options hold options for the Bundle controller.
-type Options struct {
-	// Log is the Bundle controller logger.
-	Log logr.Logger
```
SgtCoDFish, line 47: suggestion: I don't think it's required for this PR, but I'd love to see the logger pulled out of `Options` and passed around in the context. But it's not that important.

SgtCoDFish (changes requested):
Got a couple of suggestions, what do you think?

pkg/bundle/bundle.go
erikgb: It was actually a circular dependency that made me move `Options`, but I agree with your comment. Will fix it.

pkg/bundle/bundle.go
erikgb: Funny that you comment on this - as I have always been annoyed with why `Context` is not used to store/fetch a logger in this project. Since you seem open to this change, I will open a PR to fix this - before I make this PR ready for another round of review. 😸

pkg/bundle/bundle.go
erikgb: Would you be ok with moving the options package to `./pkg`? I cannot leave the Bundle options where it currently is located under `./pkg` - as that will create a circular dependency.

pkg/bundle/internal/target/source.go
erikgb: > That pattern is more useful when the type is unexported because then it's impossible to create a bundleBuilder which hasn't been initialized/validated

That will require a new interface, right? I am ok with creating a new interface, but are you?

erikgb:
Closing in favor of https://github.com/cert-manager/trust-manager/pull/648

/close

Issue #433, opened 2024-09-10
Should update target JKS/PKCS when password changed


Shawcs:
Hi !

I noticed some problems around password for generated p12 and JKS file.

### short description: 

For JKS you can alway open it with a no password and the password you setup in the bundle

For p12 and JKS if the password is updated after the bundle creation the created resources in destination namespace are not re generated and keep the old password, even if the bundle update it's generation.

---
### Reproduction of the problem

1. create a bundle with jks and p12 with a password
```yaml
apiVersion: trust.cert-manager.io/v1alpha1
kind: Bundle
metadata:
  name: my-bundle
spec:
  sources:
    - inLine: |
        -----BEGIN CERTIFICATE-----
        MIIDojCCAYqgAwIBAgIQV5ocp05c1d2ULNLEDrdCpTANBgkqhkiG9w0BAQsFADBH
        MQswCQYDVQQGEwJDSDEnMCUGA1UEChMeQmFucXVlIExvbWJhcmQgT2RpZXIgZXQg
        Q2llIFNBMQ8wDQYDVQQDEwZMTyBEUEkwHhcNMjQwMzA3MDAwMDAwWhcNMjUwMzA3
        MjM1OTU5WjAVMRMwEQYDVQQDEwpnaXRodWIuY29tMFkwEwYHKoZIzj0CAQYIKoZI
        zj0DAQcDQgAEQrPuGOisrWzPTzsVzujNAMvKeM1GRDs18c2N5R6LemewOMjO0Ep1
        yESxF/xn4Zj7tlsTeMT5zz4Li1DQN/K1zKOBhjCBgzAdBgNVHQ4EFgQUO2g/NDr1
        RzTK76ZOPZq9Xm56zJ8wDgYDVR0PAQH/BAQDAgeAMAwGA1UdEwEB/wQCMAAwHQYD
        VR0lBBYwFAYIKwYBBQUHAwEGCCsGAQUFBwMCMCUGA1UdEQQeMByCCmdpdGh1Yi5j
        b22CDnd3dy5naXRodWIuY29tMA0GCSqGSIb3DQEBCwUAA4ICAQBEaD1pszjmbtjc
        nE0s43FN2oU/S23Wf331M6Ae72F9B3ceqY/X0oPIHsOtpA7gSDOzjs4sNrHq34qn
        q3SRpmiDepmY4Ba2+gXNg5dWeul2e7ht22L/jYR8oT0pGClWkLiveijBT3Veqbxu
[...]

PR #443, opened 2024-09-22
test: add tests for truststore deterministic

erikgb:
I would still prefer if we could perform an unconditional SSA of target resources and avoid the complex and error-prone target "needs update" logic. This could eventually address https://github.com/cert-manager/trust-manager/issues/433 and in general, ensure trust-manager target resources are more up-to-date.

It seems like trust-manager truststores are now deterministic, except PKCS#12 truststores with a custom (non-empty) password. I attempted to fix this but didn't manage to.

### DIFF
--- a/pkg/bundle/internal/truststore/types_test.go
+++ b/pkg/bundle/internal/truststore/types_test.go
@@ -23,11 +23,63 @@ import (
 	"testing"
 
 	"github.com/pavlo-v-chernykh/keystore-go/v4"
+	"github.com/stretchr/testify/assert"
 
 	"github.com/cert-manager/trust-manager/pkg/apis/trust/v1alpha1"
 	"github.com/cert-manager/trust-manager/pkg/util"
 	"github.com/cert-manager/trust-manager/test/dummy"
 )
+
+func Test_Encoder_Deterministic(t *testing.T) {
+	tests := map[string]struct {
+		encoder             Encoder
+		expNonDeterministic bool
+	}{
+		"JKS default password": {
+			encoder: NewJKSEncoder(v1alpha1.DefaultJKSPassword),
+		},
+		"JKS custom password": {
+			encoder: NewJKSEncoder("my-password"),
+		},
+		"PKCS#12 default password": {
+			encoder: NewPKCS12Encoder(v1alpha1.DefaultPKCS12Password),
+		},
+		"PKCS#12 custom password": {
+			encoder: NewPKCS12Encoder("my-password"),
+			// FIXME: We should try to make all encoders deterministic
+			expNonDeterministic: true,
+		},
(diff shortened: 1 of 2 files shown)


pkg/bundle/internal/truststore/types_test.go
```diff
@@ -23,12 +23,65 @@ import (
...
+		"JKS default password": {
+			encoder: NewJKSEncoder(v1alpha1.DefaultJKSPassword),
+		},
+		"JKS custom password": {
+			encoder: NewJKSEncoder("my-password"),
+		},
+		"PKCS default password": {
+			encoder: NewPKCS12Encoder(v1alpha1.DefaultPKCS12Password),
+		},
+		"PKCS custom password": {
```
SgtCoDFish, line 47: nitpick: There are other things called "PKCS", could we explicitly write `PKCS#12` in these test names instead?

pkg/bundle/internal/truststore/types_test.go
```diff
@@ -23,12 +23,65 @@ import (
...
+			encoder: NewPKCS12Encoder("my-password"),
+			// FIXME: We should try to make all encoders deterministic
+			expNonDeterministic: true,
+		},
+	}
+
+	for name, test := range tests {
+		// test := test // NOTE: uncomment for Go < 1.22, see /doc/faq#closures_and_goroutines
```
SgtCoDFish, line 55: suggestion: I'd be happy to remove this commented line. Go versions below 1.22 aren't supported by Go any more, so I don't think we need to help people using them!

erikgb: But we require Go > 1.22, ref. https://github.com/cert-manager/trust-manager/blob/main/go.mod#L3

SgtCoDFish: I might be missing something here 😅 

Go 1.22 is the [version](https://tip.golang.org/doc/go1.22) which removed the shadowing issue, right? We require go 1.22+ as you say and go 1.21 is out of support now.

This comment is only relevant for go 1.21 and below, isn't it? So it can be removed?

Am I missing something?

pkg/bundle/internal/truststore/types_test.go
erikgb: Go catch, updated!

pkg/bundle/internal/truststore/types_test.go
erikgb: Ugh, sorry! 😊 I added this copy-pasted comment. It's removed now. Appologies!

SgtCoDFish (approved):
/lgtm
/approve

Thanks for this, I really like this kind of test 🚀

PR #449, opened 2024-10-06
Fix issue with updating of  JKS/PKCS targets when password changes

### LINKED ISSUE #433: Should update target JKS/PKCS when password changed

arsenalzp:
This draft PR is trying to fix #433 

As was advised be @erikgb  I added additional field named `passwordHash`.
However, as user shouldn't care of calculating password hash, I added mutation webhook to set `passwordHash` value during `Bundle` creation.

Let's consider this PR as draft one, therefore I would like to discuss my solution with you, then I'm gonna modify tests, prepare documentation for functions which were added.


erikgb:
But the password hash needs to be added to the targets, and not the bundle resource. 😊

arsenalzp:
> But the password hash needs to be added to the targets, and not the bundle resource. 😊

In our case it is configMap which holds additional binary data for jks/pkcs12 formats. 
Do you want to add annotations with hashes for each target formats?
I guess it is possible.

arsenalzp:
I made changes as was discussed. However, it was for `ConfigMap` target only. Once agreed all changes then I will make changes for `Secret` target and tests as well.

pkg/bundle/internal/target/target.go
```diff
@@ -242,6 +254,15 @@ func (r *Reconciler) needsUpdate(ctx context.Context, kind Kind, log logr.Logger
 		needsUpdate = true
 	}
 
+	// Get old password hash annotations.
+	// Calculate new password hash annotation with changed password.
+	// Compare old password hash annotations with new ones.
+	if bundle.Spec.Target.AdditionalFormats != nil {
+		if !isHashAnnotationEqual(getOldHashAnnotations(obj.GetAnnotations(), bundle.Spec.Target.AdditionalFormats), calcNewHashAnnotations(bundle.Spec.Target.AdditionalFormats)) {
```
erikgb, line 261: Is it required to calculate the desired annotations again? Why can't you just supply the desired annotations as a single map? Possibly bundled with the existing data hash annotation?

pkg/bundle/internal/target/target.go
```diff
@@ -396,3 +417,48 @@ func (b *Data) Populate(pool *util.CertPool, formats *trustapi.AdditionalFormats
...
+	crcTable := crc32.MakeTable(128)
+	crc := crc32.New(crcTable)
+	crc.Write([]byte(passwd))
+
+	return strconv.Itoa(int(crc.Sum32()))
+}
+
+// Retrieve old password hashes from the given annotations
+// Annotation keys are taken from bundle additional formats keys
+func getOldHashAnnotations(annotations map[string]string, bundle *trustapi.AdditionalFormats) map[string]string {
```
erikgb, line 432: I don't think this helper adds any value. If you just loop through the desired annotations in `needsUpdate` and check the desired annotations against the target resource annotations I think the code is easier to read.

pkg/bundle/internal/target/target.go
```diff
@@ -396,3 +417,48 @@ func (b *Data) Populate(pool *util.CertPool, formats *trustapi.AdditionalFormats
...
+	return passwdHashAnnotations
+}
+
+// Calculate new password hash annotations from the given additional formats
+// keys and passwords
+func calcNewHashAnnotations(bundle *trustapi.AdditionalFormats) map[string]string {
+	var passwdHashAnnotations = make(map[string]string)
+
+	if bundle.JKS != nil && bundle.JKS.Password != nil {
+		passwdHashAnnotations[bundle.JKS.Key] = calculateHash(*bundle.JKS.Password)
```
erikgb, line 451:
```suggestion
		passwdHashAnnotations["trust.cert-manager.io/jks-pwd-hash"] = calculateHash(*bundle.JKS.Password)
```

I don't think these keys should be dynamic. Please use well-known keys. I've added a suggestion, but feel free to suggest something else.

pkg/bundle/internal/target/target.go
```diff
@@ -396,3 +417,48 @@ func (b *Data) Populate(pool *util.CertPool, formats *trustapi.AdditionalFormats
...
+	}
+	if bundle.PKCS12 != nil && bundle.PKCS12.Password != nil {
+		passwdHashAnnotations[bundle.PKCS12.Key] = calculateHash(*bundle.PKCS12.Password)
```
erikgb, line 454: Same as above. I don't think these annotation keys should be dynamic.

erikgb (changes requested):
Thanks for looking into this @arsenalzp! First review pass done. Before this can be merged we need tests for this and a similar implementation for target secrets.

pkg/bundle/internal/target/target.go
```diff
@@ -396,3 +417,48 @@ func (b *Data) Populate(pool *util.CertPool, formats *trustapi.AdditionalFormats
...
 	return nil
 }
+
+// Calculate hash from the given password string.
+func calculateHash(passwd string) string {
+	crcTable := crc32.MakeTable(128)
+	crc := crc32.New(crcTable)
+	crc.Write([]byte(passwd))
+
+	return strconv.Itoa(int(crc.Sum32()))
```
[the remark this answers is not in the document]
arsenalzp: I guess it is excessively to use sha256 for short text patterns like target password.
OK, I will rewrite.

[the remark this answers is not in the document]
erikgb: That's a fair point. I will consider the alternatives again. 😊

pkg/bundle/internal/target/target.go
arsenalzp: Do you want me to make use two static annotation keys like `trust.cert-manager.io/jks-hash` and `trust.cert-manager.io/pksc12-hash` instead using annotation keys based on targets keys?

In that case code would be

```
if obj.GetLabels()[trust.cert-manager.io/jks-hash] != bundle.Name {
		needsUpdate = true
}
if obj.GetLabels()[trust.cert-manager.io/pksc12-hash] != bundle.Name {
		needsUpdate = true
}
```

pkg/bundle/internal/target/target.go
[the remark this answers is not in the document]
arsenalzp: So, do keep it as it is?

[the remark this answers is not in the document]
erikgb: I still think we should use a proper hash algorithm. And to keep it simple and consistent I think SHA 256 might be an ok choice. My main concern when reading about CRC (which is not well known to me):

> Secondly, unlike cryptographic hash functions, CRC is an easily reversible function, which makes it unsuitable for use in digital signatures.

While these truststore passwords are not very sensible secrets, I am pretty sure some users will think we should treat them well - as passwords. I understand your concern about size, but encoded as hex it is not that bad. To avoid calculating the hash for each target, I suggest building the hash in the main Bundle reconcile function. It can be supplied together with the existing data hash.

Maybe @inteon or @SgtCoDFish wants to chime in here?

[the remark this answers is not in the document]
arsenalzp: Exactly! As passwords are stored in insecure fields, no reason to produce secure hash.
One potential issue here is hash collision, when user changed password, but nothing happened because new calculated hash is the same one as for old password.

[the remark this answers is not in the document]
inteon: We already use a sha256 hash here: https://github.com/cert-manager/trust-manager/blob/c32b33f6229eececd19a1457df483da34980a1cb/pkg/bundle/internal/target/target.go#L111

I would try and use the same mechanism (even combine them if possible).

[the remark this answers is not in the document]
arsenalzp: Got it. 
I'm gonna rewrite it to use sha256 hash function.

arsenalzp:
All remarks have been removed.

pkg/bundle/internal/target/target.go
```diff
@@ -242,6 +256,23 @@ func (r *Reconciler) needsUpdate(ctx context.Context, kind Kind, log logr.Logger
...
+			}
+		}
+
+		if bundle.Spec.Target.AdditionalFormats.PKCS12 != nil {
+			if hash, ok := obj.GetAnnotations()[Pkcs12PasswdHashAnnotation]; !ok || hash != fmt.Sprintf("%x", sha256.Sum256([]byte(*bundle.Spec.Target.AdditionalFormats.PKCS12.Password))) {
+				needsUpdate = true
+			}
+		}
+	}
+
```
erikgb, line 275: This block seems overly complicated. Why can't you use the same (simple) approach as for the data hash:

https://github.com/cert-manager/trust-manager/blob/17561571e008ac2775df6ef966bf89dd325a9913/pkg/bundle/internal/target/target.go#L241

pkg/bundle/internal/target/target.go
```diff
@@ -42,6 +42,11 @@ import (
 	"github.com/cert-manager/trust-manager/pkg/util"
 )
 
+const (
+	JksPasswdHashAnnotation    = "trust.cert-manager.io/jks-pwd-hash"
+	Pkcs12PasswdHashAnnotation = "trust.cert-manager.io/pksc12-pwd-hash"
+)
```
erikgb, line 48: For consistency, I think these annotations should be alongside the related data hash:

https://github.com/cert-manager/trust-manager/blob/17561571e008ac2775df6ef966bf89dd325a9913/pkg/apis/trust/v1alpha1/types_bundle.go#L26

pkg/bundle/internal/target/target.go
```diff
@@ -396,3 +427,18 @@ func (b *Data) Populate(pool *util.CertPool, formats *trustapi.AdditionalFormats
 	}
 	return nil
 }
+
+// Calculate new password hash annotations from the given additional formats
+// keys and passwords
+func calcNewHashAnnotations(bundle *trustapi.AdditionalFormats) map[string]string {
```
erikgb, line 433:
```suggestion
func truststorePasswordAnnotations(bundle *trustapi.AdditionalFormats) map[string]string {
```

Just a suggestion. To me, the current name doesn't have the correct context. The term "new" is not relevant, and "HashAnnotations" is too generic.

erikgb (changes requested):
IMO, this is moving in the right direction now. I would still love to see this new stuff better aligned with the existing data hash. I also think it's time to write some tests - before we finally implement this for all target resource types.

pkg/bundle/internal/target/target.go
arsenalzp: I agree. Fixed.

pkg/bundle/internal/target/target.go
arsenalzp: I agree. Fixed

pkg/bundle/internal/target/target.go
arsenalzp: Indeed, it could be simple, however in some conditions when `Password` is nil application is falling into `nil pointer dereference` panic. _Though modesty be a virtue, yet bashfulness is a vice._  :)

erikgb: But why do you need to calculate any hash here? The correct approach is just to check if the annotations are present with the correct value. And the correct value is the hash calculate in the reconcile func - based on the desired state.

arsenalzp: Unfortunately, then no chance to identify that was a password  changed, so no bundle update is done.

arsenalzp: Ah, I guess you are talking about putting hashes values to `SyncConfigMap` function.

erikgb: This should be very similar to what is currently done to detect bundle content changes (data hash). I am trying to argue that what you are working on should be aligned with the existing functionality. 😉 I would probably supply the desired annotations (containing at least the existing data hash annotation, and eventually these new truststore password annotations) to the sync-functions. The code should be fairly easy to check if the target resource already has the correct annotations - with the correct values (hashes).

arsenalzp: Fixed.

pkg/bundle/internal/target/target.go
```diff
@@ -111,10 +111,18 @@ func (r *Reconciler) SyncConfigMap(
...
 	configMapBinData := resolvedBundle.BinaryData
 
+	// Create empty map.
+	// If no additional formats are present then
+	// no annotations will be written into ConfigMap
+	var passwdHashAnnotations = make(map[string]string)
+	if bundle.Spec.Target.AdditionalFormats != nil {
+		passwdHashAnnotations = truststorePasswordAnnotations(bundleTarget.AdditionalFormats)
+	}
+
```
erikgb, line 121: I suggest you consolidate these optional new annotations with the data hash value (and annotation) a few lines above. This will make the code more compact and also easier to reason about, at least for me.

We should probably also move the calculation of the hashes to the Bundle reconcile function, and supply the desired annotations as a parameter to the sync function. This will avoid calculating the same value for each and every target resource. But that can be fixed in a follow-up PR.

pkg/apis/trust/v1alpha1/types_bundle.go
```diff
@@ -20,10 +20,14 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-var BundleKind = "Bundle"
+var (
```
erikgb, line 23:
```suggestion
const (
```

If changing this, why not use constants instead? I don't understand why these values are declared as variables. Seems like constants to me, but there might be a reason.

pkg/bundle/internal/target/target.go
erikgb:
````go
for k, v := range passwdHashAnnotations {
    if obj.GetAnnotations()[k] != v {
      needsUpdate = true
    }
}
````

Isn't something like this simpler?

File: pkg/apis/trust/v1alpha1/types_bundle.go (hunk 10 above)
Comment (arsenalzp) [reply pkg/apis/trust/v1alpha1/types_bundle.go:23]: I was following as it was done by someone else - var declaration. Yes, constant variables are better in this particular case.

File: pkg/bundle/internal/target/target.go (hunk 6 above)
Comment (arsenalzp) [reply pkg/bundle/internal/target/target.go:262]: Well, for many items a loop should be used with not doubt, however we have only 3 items now, so for obviousness I prefer don't use it.
Anyway, to simplify the code I did as you proposed.

File: pkg/bundle/internal/target/target.go (hunk 9 above)
Comment (arsenalzp) [reply pkg/bundle/internal/target/target.go:122]: Yes, we can improve it in a follow-up PRs.

**Comment by arsenalzp** [comment]
Were my changes accepted? Can I proceed with the hardest step - test?

**Comment by erikgb** [comment]
> Were my changes accepted? Can I proceed with the hardest step - test?

Looking good to me, @arsenalzp! Well done and thanks! Please add some testing of this! 😺

**Comment by arsenalzp** [comment]
Existing tests were modified

**Comment by arsenalzp** [comment]
It seems the issue is not on my side...

**Comment by arsenalzp** [comment]
It seems the issue on my side with gomarkdoc.

**Comment by inteon** [comment]
> It seems the issue on my side with gomarkdoc.

If you run `make generate`, does that solve the issue?

**Comment by arsenalzp** [comment]
```
pkg/apis/trust/v1alpha1/types_bundle.go:28:2: G101: Potential hardcoded credentials (gosec)
	BundleJksPasswdHashAnnotation    = "trust.cert-manager.io/jks-pwd-hash"
	^
pkg/apis/trust/v1alpha1/types_bundle.go:29:2: G101: Potential hardcoded credentials (gosec)
	BundlePkcs12PasswdHashAnnotation = "trust.cert-manager.io/pksc12-pwd-hash"
```

It is necessary to exclude these constants.

**Comment by erikgb** [comment]
@arsenalzp I think this is a false positive that you have to suppress to pass gosec linting.

````
pkg/apis/trust/v1alpha1/types_bundle.go:28:2: G101: Potential hardcoded credentials (gosec)
	BundleJksPasswdHashAnnotation    = "trust.cert-manager.io/jks-pwd-hash"
	^
pkg/apis/trust/v1alpha1/types_bundle.go:29:2: G101: Potential hardcoded credentials (gosec)
	BundlePkcs12PasswdHashAnnotation = "trust.cert-manager.io/pksc12-pwd-hash"
	^
make[1]: *** [make/_shared/go/01_mod.mk:112: verify-golangci-lint] Error 1
````

File: docs/api/api.md [hunk 11]
```diff
@@ -58,6 +58,19 @@ import "github.com/cert-manager/trust-manager/pkg/apis/trust/v1alpha1"
...
 
+<a name="BundleKind"></a>
+
+```go
+const (
+    BundleKind = "Bundle"
+
+    BundleLabelKey                   = "trust.cert-manager.io/bundle"
+    BundleHashAnnotationKey          = "trust.cert-manager.io/hash"
+    BundleJksPasswdHashAnnotation    = "trust.cert-manager.io/jks-pwd-hash"
```
erikgb, line 69:
```suggestion
    BundleJksPasswdHashAnnotation    = "trust.cert-manager.io/jks-pwd-hash" // # nosec G101
```

docs/api/api.md
```diff
@@ -58,6 +58,19 @@ import "github.com/cert-manager/trust-manager/pkg/apis/trust/v1alpha1"
...
+    BundlePkcs12PasswdHashAnnotation = "trust.cert-manager.io/pksc12-pwd-hash"
```
erikgb, line 70:
```suggestion
    BundlePkcs12PasswdHashAnnotation = "trust.cert-manager.io/pksc12-pwd-hash" // # nosec G101
```

erikgb:
@arsenalzp, please run `make generate` to fix the linting. As the error output suggests.

arsenalzp:
> @arsenalzp, please run `make generate` to fix the linting, as the error output suggests.

I did it, it replace `#nosec` stanzas:

```
$ diff --git a/docs/api/api.md b/docs/api/api.md
index ee6ebf9..6c09868 100644
--- a/docs/api/api.md
+++ b/docs/api/api.md
@@ -66,8 +66,8 @@ const (
 
     BundleLabelKey                   = "trust.cert-manager.io/bundle"
     BundleHashAnnotationKey          = "trust.cert-manager.io/hash"
-    BundleJksPasswdHashAnnotation    = "trust.cert-manager.io/jks-pwd-hash" # nosec G101
-    BundlePkcs12PasswdHashAnnotation = "trust.cert-manager.io/pksc12-pwd-hash" # nosec G101
+    BundleJksPasswdHashAnnotation    = "trust.cert-manager.io/jks-pwd-hash"
+    BundlePkcs12PasswdHashAnnotation = "trust.cert-manager.io/pksc12-pwd-hash"
 )
$
 ```

```

erikgb:
> I did it, it is replacing `#nosec` stanzas:

But we don't need, and should not have, the gosec suppressions in a markdown file! 😆

pkg/apis/trust/v1alpha1/types_bundle.go
```diff
@@ -20,10 +20,14 @@ import (
...
 
-var BundleKind = "Bundle"
+const (
+	BundleKind = "Bundle"
 
-var BundleLabelKey = "trust.cert-manager.io/bundle"
-var BundleHashAnnotationKey = "trust.cert-manager.io/hash"
+	BundleLabelKey                   = "trust.cert-manager.io/bundle"
+	BundleHashAnnotationKey          = "trust.cert-manager.io/hash"
+	BundleJksPasswdHashAnnotation    = "trust.cert-manager.io/jks-pwd-hash"
```
erikgb, line 28:
```suggestion
	BundleJksPasswdHashAnnotation    = "trust.cert-manager.io/jks-pwd-hash" # nosec G101
```

pkg/apis/trust/v1alpha1/types_bundle.go
```diff
@@ -20,10 +20,14 @@ import (
...
+	BundlePkcs12PasswdHashAnnotation = "trust.cert-manager.io/pksc12-pwd-hash"
```
erikgb, line 29:
```suggestion
	BundlePkcs12PasswdHashAnnotation = "trust.cert-manager.io/pksc12-pwd-hash" # nosec G101
```

erikgb:
My fault! Sorry! I added the suggestions to the markdown file. 🛕

pkg/apis/trust/v1alpha1/types_bundle.go
```diff
@@ -20,10 +20,14 @@ import (
...
-var BundleKind = "Bundle"
+const (
+	BundleKind = "Bundle"
 
-var BundleLabelKey = "trust.cert-manager.io/bundle"
-var BundleHashAnnotationKey = "trust.cert-manager.io/hash"
+	BundleLabelKey                   = "trust.cert-manager.io/bundle"
+	BundleHashAnnotationKey          = "trust.cert-manager.io/hash"
+	BundleJksPasswdHashAnnotation    = "trust.cert-manager.io/jks-pwd-hash"    // # nosec G101
+	BundlePkcs12PasswdHashAnnotation = "trust.cert-manager.io/pksc12-pwd-hash" // # nosec G101
```
erikgb, line 29:
```suggestion
	BundleJksPasswdHashAnnotationKey    = "trust.cert-manager.io/jks-pwd-hash"    // #nosec G101
	BundlePkcs12PasswdHashAnnotationKey = "trust.cert-manager.io/pksc12-pwd-hash" // #nosec G101
```

It seems like gosec does not understand the suppressions. Looking at other suppressions in this project, I think the space matters. Also suggesting to follow the `(...)Key` pattern established just above.

arsenalzp:
Yeah, it seems the issue with a the `space` character, here is example from project doc:

```
To do this, you can list the rule(s) to be suppressed within the #nosec annotation, e.g: /* #nosec G401 */ or //#nosec G201 G202 G203
```

inteon:
FYI: you can locally generate using "make generate" end verify using "make verify-golangci-lint"

Will take a look at the issue myself on Monday.

arsenalzp:
> FYI: you can locally generate using "make generate" end verify using "make verify-golangci-lint"
> 
> Will take a look at the issue myself on Monday.

Hello,
Yes, I did it yesterday, however the same error is still popping up:

```
$ make verify-golangci-lint
Running '_bin/tools/golangci-lint run --go 1.23.2 -c /home/username/go/src/trust-manager/.golangci.yaml' in directory './trust-packages/debian'

Running '_bin/tools/golangci-lint run --go 1.23.2 -c /home/username/go/src/trust-manager/.golangci.yaml' in directory '.'
pkg/apis/trust/v1alpha1/types_bundle.go:26: File is not `gofmt`-ed with `-s` (gofmt)
        BundleLabelKey                   = "trust.cert-manager.io/bundle"
        BundleHashAnnotationKey          = "trust.cert-manager.io/hash"
make: *** [make/_shared/go/01_mod.mk:112: verify-golangci-lint] Ошибка 1
$
```

arsenalzp:
Issue was found and fixed

pkg/bundle/bundle_test.go
```diff
@@ -141,12 +141,21 @@ func Test_Reconcile(t *testing.T) {
...
 		}
 
-		configMapPatch = func(name, namespace string, data map[string]string, binData map[string][]byte, key *string) *coreapplyconfig.ConfigMapApplyConfiguration {
+		configMapPatch = func(name, namespace string, data map[string]string, binData map[string][]byte, key *string, additionaFormats *trustapi.AdditionalFormats) *coreapplyconfig.ConfigMapApplyConfiguration {
 			annotations := map[string]string{}
 			if key != nil {
 				annotations[trustapi.BundleHashAnnotationKey] = fmt.Sprintf("%x", sha256.Sum256([]byte(data[*key])))
 			}
 
+			if additionaFormats != nil {
```
inteon, line 150: Can we use "truststorePasswordAnnotations" here?

arsenalzp: As this particular variable of type `trustapi.AdditionalFormats` I guess it is good to keep a name which reflects its designation.

```
configMapPatch = func(name, namespace string, data map[string]string, binData map[string][]byte, key *string, additionaFormats *trustapi.AdditionalFormats) *coreapplyconfig.ConfigMapApplyConfiguration {
			annotations := map[string]string{}
			if key != nil {
				annotations[trustapi.BundleHashAnnotationKey] = fmt.Sprintf("%x", sha256.Sum256([]byte(data[*key])))
			}

			if additionaFormats != nil {
				if additionaFormats.JKS != nil {
...
```

arsenalzp: I guess it leads to a confusion, as `annotation` variable consist hashes not for passwords only, there is annotation for `trustapi.BundleHashAnnotationKey` as well. 
So, I would keep the name as it is.

inteon: Sorry, I was unclear. What I meant to ask is if we can use the existing function "truststorePasswordAnnotations" here instead of the logic that follows.

arsenalzp: @erikgb asked me to do use logic statements instead of function

erikgb: @arsenalzp, my comment was addressing the `needsUpdate` logic, and not the patch assert here in the test..

arsenalzp: Hello @erikgb @inteon,
Any objections regarding my statement in previous message?

erikgb: I am planning to review this PR again, but my week has been busy so far.

pkg/bundle/internal/target/target.go
```diff
@@ -196,20 +205,24 @@ func (r *Reconciler) SyncSecret(
 		secretData[k] = v
 	}
 
+	// If no additional formats are present then
+	// no additional annotations will be written into ConfigMap
```
erikgb, line 209:
```suggestion
	// no additional annotations will be written into Secret
```

pkg/bundle/bundle_test.go
```diff
@@ -201,6 +210,10 @@ func Test_Reconcile(t *testing.T) {
 				annotations[trustapi.BundleHashAnnotationKey] = fmt.Sprintf("%x", sha256.Sum256([]byte(data[*key])))
 			}
 
+			// These annotation are used for password hash sum testing
+			annotations[trustapi.BundleJksPasswdHashAnnotationKey] = "annotation will be changed during reconciliation"
+			annotations[trustapi.BundlePkcs12PasswdHashAnnotationKey] = "annotation will be changed reconciliation"
+
```
erikgb, line 216:
```suggestion
```

This does not seem required. If you think I am wrong, please explain.

pkg/bundle/bundle_test.go
erikgb: I must say I agree with @arsenalzp on this. Exporting the internal function seems wrong to me too. It might be my previous refactoring that complicates this a bit, and I can try to improve this in a follow-up PR.

erikgb (changes requested):
There are still things to do before this PR LGTM, but I agree with the general approach taken here. 👍 First and foremost, I am missing a test for the reported issue: changing the password should update the target(s).

The PR needs rebase now. Please let me know if you need help with this, @arsenalzp!

pkg/bundle/bundle_test.go
arsenalzp: Those annotation are just fake passwords hashes which are changed after the reconcile process. Of course it is possible to put something "blablabla", however I would like to keep meaningful text. Just in case.

arsenalzp:
Hm, rebase was done locally, however it is still present.

pkg/bundle/bundle_test.go
erikgb: If we set annotations, I agree that the values should be meaningful. But do we gain anything from setting annotations here?

pkg/bundle/internal/target/target.go
```diff
@@ -105,26 +106,32 @@ func (r *Reconciler) SyncConfigMap(
 
 	// Generated PKCS #12 is not deterministic - best we can do here is update if the pem cert has
 	// changed (hence not checking if PKCS #12 matches)
-	dataHash := fmt.Sprintf("%x", sha256.Sum256([]byte(resolvedBundle.Data)))
+	annotations := make(map[string]string)
+	annotations[trustapi.BundleHashAnnotationKey] = fmt.Sprintf("%x", sha256.Sum256([]byte(resolvedBundle.Data)))
```
erikgb, line 110:
```suggestion
	annotations := map[string]string{
		trustapi.BundleHashAnnotationKey:  fmt.Sprintf("%x", sha256.Sum256([]byte(resolvedBundle.Data))),
	}
```

pkg/bundle/bundle_test.go
arsenalzp: Of course it is possible to use zero string `""`, however I would like to see much meaningful values here.

erikgb:
image

@arsenalzp Are you giving up? 🤠 There should be no need for a new PR. Just fix the branch. 😺

arsenalzp:
> image
> 
> @arsenalzp Are you giving up? 🤠 There should be no need for a new PR. Just fix the branch. 😺

No, it is not so easy to get rid of me :-D

arsenalzp:
Reopen.

arsenalzp:
The button "Reopen and comment" was enabled and I was open to re-open this PR by clicking on that button, however there is still no any commit in the list. It looks weird.

arsenalzp:
изображение
It seems, it is possible to re-open, however no commits in the commits list.

erikgb:
@arsenalzp, please ping me for another review when you have addressed all the present remarks!

arsenalzp:
It seems all remarks have already been resolved...

erikgb:
> It seems all remarks have already been resolved...

I still can't find any test of the issue this PR is solving. I think it will fix it, but it would be nice with a test verifying it's fixed.

arsenalzp:
> > It seems all remarks have already been resolved...
> 
> I still can't find any test of the issue this PR is solving. I think it will fix it, but it would be nice with a test verifying it's fixed.

I agree. 
As temporary solution it is possible to use the Bundle tests which were modified by me to reflect changes I made into the code. There is a comparison of annotations so it is possible to asses result indirectly by using hash annotations.

erikgb:
> Hello colleagues, Do you have any feedback on this solution?

@arsenalzp The fix looks good, but I would still love to see a test verifying that the reported issue is fixed. However I tend to accept your proposed changes and leave the test improvement for a follow-up PR - since this PR appears to fix an annoying bug and has been open for a while. 🤠 But can you please fix the commit message on your single commit?

arsenalzp:
Fixed.

erikgb (commented):
/lgtm
/approve
/hold

Looks good to me, but since @inteon had some comments on this PR I am putting a hold for now. Tim, could you please take a look? @arsenalzp has been working on this for a long time now.

I am going to prepare a PR for deduplicating the target logic more after this is merged.

arsenalzp:
> /lgtm /approve /hold
> 
> Looks good to me, but since @inteon had some comments on this PR I am putting a hold for now. Tim, could you please take a look? @arsenalzp has been working on this for a long time now.
> 
> I am going to prepare a PR for deduplicating the target logic more after this is merged.

What happened with deduplication? Those changes were introduced by me last year.

erikgb:
> > /lgtm /approve /hold
> > Looks good to me, but since @inteon had some comments on this PR I am putting a hold for now. Tim, could you please take a look? @arsenalzp has been working on this for a long time now.
> > I am going to prepare a PR for deduplicating the target logic more after this is merged.
> 
> What happened with deduplication? Those changes were introduced by me last year.

I meant deduplication of source code. Not certs in trust bundle. 😉 You fixed the latter, yes.

Issue #428, opened 2024-08-30
About Target JKS/PKCS12 password - Not in the generated target configMap/Secret


Ceddaerrix:
I have recently discovered trust-manager and started experimenting with it.
To me, the part about JKS/PKCS12 password lacks some clarity:

- In https://cert-manager.io/docs/trust/trust-manager/#targets, `spec.target.additionalFormats.jks.password` and `spec.target.additionalFormats.pkcs12.password` are introduced as a mean to overwrite the default 'changeit' password.
- In https://cert-manager.io/docs/trust/trust-manager/api-reference/#bundlespectargetadditionalformats, no mention of the password parameter or their purposes (=> documentation needs to be updated)

Looking to use the bundle outcome (as a secret) with Jenkins for the HTTPS keystore (see https://github.com/jenkinsci/helm-charts/blob/e60a42b6e11f41328f73c4a90bcbcce14f535165/charts/jenkins/values.yaml#L836-L846), I was under the impression that the JKS/PKCS12 password would be included into the generated secret.
Unfortunately, tests showed that it is not the case...

1. Would it possible to set `spec.target.additionalFormats.jks.password`/`spec.target.additionalFormats.pkcs12.password` from a secret reference and key?
2. Would it be possible to have that password included in to the generated secret, or a separate one?

erikgb:
Hi @Ceddaerrix, this seems to be a multi-purpose issue. 😉 

It seems like the trust-manager API documentation is out-of-date. Thanks for letting us know! I will look into what's happened here.

> Would it possible to set spec.target.additionalFormats.jks.password/spec.target.additionalFormats.pkcs12.password from a secret reference and key?

Why do you need this? It would make the controller mechanics considerably more complex. Passwords on JKS/PKCS#12 keystores/truststores do not make things more secure and should be considered legacy. There is a note about them in our FAQ: https://cert-manager.io/docs/faq/#why-are-passwords-on-jks-or-pkcs12-files-not-helpful

> Would it be possible to have that password included in to the generated secret, or a separate one?

Why? If you think the keystore/truststore password represent anything useful, you should keep it separate from the keystore/password and manage it independently. 😸

PR #455, opened 2024-10-12
fix: make PKCS#12 truststores deterministic

erikgb:
This PR might appear a bit controversial, since it short-circuits the random generator for our PKCS#12 truststores. 🤠 But there is a reason! My end goal is to greatly simplify the Bundle controller by removing the `needsUpdate`-logic and replacing it with well-tuned, but unconditional, SSA requests.

### DIFF
--- a/pkg/bundle/internal/truststore/types.go
+++ b/pkg/bundle/internal/truststore/types.go
@@ -25,6 +25,7 @@ import (
 	"github.com/pavlo-v-chernykh/keystore-go/v4"
 	"software.sslmate.com/src/go-pkcs12"
 
+	"github.com/cert-manager/trust-manager/pkg/bundle/internal/insecurerand"
 	"github.com/cert-manager/trust-manager/pkg/util"
 )
 
@@ -98,7 +99,11 @@ func (e pkcs12Encoder) Encode(trustBundle *util.CertPool) ([]byte, error) {
 		})
 	}
 
-	encoder := pkcs12.LegacyRC2
+	encoder := pkcs12.LegacyRC2.
+		// Short-circuiting the rand generator to make our PKCS#12 truststores deterministic.
+		// This should allow use of unconditional SSA requests from controller.
+		// See: https://cert-manager.io/docs/faq/#why-are-passwords-on-jks-or-pkcs12-files-not-helpful
+		WithRand(insecurerand.Rand)
 
 	if e.password == "" {
 		encoder = pkcs12.Passwordless
--- a/pkg/bundle/internal/insecurerand/insecurerand.go
+++ b/pkg/bundle/internal/insecurerand/insecurerand.go
@@ -1,0 +1,32 @@
+/*
+Copyright 2021 The cert-manager Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package insecurerand
+
+import "io"
+
+var Rand io.Reader = &insecureRand{}
+
+type insecureRand struct {
+}
+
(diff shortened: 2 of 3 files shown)


pkg/bundle/internal/truststore/types.go
```diff
@@ -98,10 +99,14 @@ func (e pkcs12Encoder) Encode(trustBundle *util.CertPool) ([]byte, error) {
...
+	encoder := pkcs12.Passwordless
 
-	if e.password == "" {
-		encoder = pkcs12.Passwordless
+	if e.password != "" {
+		encoder = pkcs12.LegacyRC2.
+			// Short-circuiting the rand generator to make our PKCS#12 truststores deterministic.
+			// This should allow use of unconditional SSA requests from controller.
+			// See: https://cert-manager.io/docs/faq/#why-are-passwords-on-jks-or-pkcs12-files-not-helpful
+			WithRand(rand.New(rand.NewSource(1))) //#nosec G404
```
munnerz, line 109: Can this behaviour be made optional/tunable by the user/administrator? I imagine some environments may require use of a secure randomness source even if the underlying algorithm is considered insecure, though I'm not sure. What does other software in this space do?

erikgb: > Can this behaviour be made optional/tunable by the user/administrator?

No, not without losing the opportunity to greatly simplify the controller (drop needsUpdate-logic). Dropping a secure random source here is for sure controversial. But I think it could be worth it to simplify our code. Some additional arguments:

- Our password-less PKCS#12 (the default) is deterministic.
- We already stated (in our FAQ) that we support keystores/truststores for legacy reasons, and not because we consider them the recommended approach.
- We already got [a suggestion](https://github.com/cert-manager/trust-manager/issues/428) to include the truststore password just next to the truststore in the target configmap/secret. If we decide to implement anything in this direction, a requirement for a secure random source becomes even more meaningless.

> What does other software in this space do?

No idea. I am pretty sure that any software producing PKCS#12 truststores with a password attempts to use a secure random source. But a password-less PKCS#12 is the new standard for truststores in Java - after they abandoned JKS.

pkg/bundle/internal/truststore/types.go
```diff
@@ -21,6 +21,7 @@ import (
 	"crypto/sha256"
 	"encoding/hex"
 	"fmt"
+	"math/rand"
```
SgtCoDFish, line 24: suggestion: Completely separate from James' concern (which may be a blocker). Sorry for the hastily written comment below!

I think if we're going to do import `math/rand` it should be in a separate package entirely, so it's "quarantined". When we work on this file in the future we don't want to assume that `rand` means `crypto/rand`.

I'd create `internal/insecurerand/insecurerand.go`, have it expose `var Rand io.Reader = rand.New(rand.NewSource(1))`.

That was this file would import `insecurerand.Rand` and `math/rand` would be contained to one file only so it couldn't accidentally be used.

That said, `math/rand.Reader` [says](https://pkg.go.dev/math/rand#NewSource) it's not safe for concurrent use:

> Unlike the default [Source](https://pkg.go.dev/math/rand#Source) used by top-level functions, this source is not safe for concurrent use by multiple goroutines.

Might that be an issue if a user has multiple bundles?

Further to that, is a seeded rand actually what we want? We want a deterministic `io.Reader` - could we create a `bytes.Buffer` with say N bytes all set to zero and use a reader on that instead? It looks like the use of `rand` doesn't actually need that many bytes: https://github.com/SSLMate/go-pkcs12/blob/v0.5.0/pkcs12.go#L705

erikgb: Many good points here! Thanks Ash! I will update this PR - even if it's unlikely it will get merged.

erikgb: I have updated the PR now addressing some of your comments.

SgtCoDFish:
I've asked Venafi's security team if they have any input on this relating to auditing / compliance concerns - will report back if I hear anything!

SgtCoDFish:
Feedback was that it's unlikely an auditor would know any context when they see something like this, and they're likely to just raise an issue that would create a bunch of extra work if they see this. They might _not_ see it depending on how they scan, but in any case it feels like deterministic PKCS#12 might be a non-starter unfortunately 🙃

erikgb:
Thanks for all the input @SgtCoDFish @munnerz! I am going to close this PR now and shift my focus to how we can improve the current `needsUpdate` logic to be less error-prone. Handling data with randomness using SSA is not easy. 😅

Issue #458, opened 2024-10-18
Support for more than one certificate from the same secret in bundle source


juliocamarero:
Hi folks,

Currently, when adding certificates to a bundle from a secret, users are required to specify the exact key for the secret(e.g. `ca.crt`) as shown below:
```yaml
apiVersion: trust.cert-manager.io/v1alpha1
kind: Bundle
metadata:
  name: my-org.com  
spec:
  sources:
  - secret:
      name: "my-certs"
      key: "ca.crt"
```

In some cases, a secret may contain multiple certificates (one per key), and the key names may be unknown to the user or could change dynamically. It would be really helpful to support dynamically including those certificates in the bundle.

One basic approach could be to allow the use of a wildcard `*` in the key field to include all keys, like this:
 ```yaml
spec:
  sources:
  - secret:
      name: "my-certs"
      key: "*"
```

Alternatively, a more flexible solution could be to introduce a new field, such as `matchKey`, which would accept regular expressions. This would allow users to match and include all certificates whose keys follow a specific pattern. For example:
 ```yaml
spec:
  sources:
  - secret:
      name: "my-certs"
      matchKey: "tls-certificate-\\d+"
```

Or to match all keys:
```yaml
spec:
  sources:
  - secret:
      name: "my-certs"
      matchKey: ".*"
```

I would be happy to contribute this feature if there's agreement on the best approach for the API (I've suggested two options here, but I'm open to other ideas as well).

Looking forward to your thoughts! 😊

PR #459, opened 2024-10-21
Support for more than one certificate from the same secret/configMap in bundle source

juliocamarero:
closes: https://github.com/cert-manager/trust-manager/issues/458

This is a first attempt to implement support for including in a bundle not just one PEM file referenced by a key, but all of them which match a regular expression. 

This is how we currently include a bundle a source PEM from a secret referenced by key:
```yaml
kind: Bundle
metadata:
  name: my-org.com  
spec:
  sources:
  - secret:
      name: "my-certs"
      key: "ca.crt"
```

And this is how we could reference more than one using a regular expression with these changes:
```yaml
kind: Bundle
metadata:
  name: my-org.com  
spec:
  sources:
  - secret:
      name: "my-certs"
      matchKey: ".*"
```

Happy to get any feedback since this is my first contribution to the project :)


SgtCoDFish:
Summarising discussion from standup on 2024-10-22:

1) We'll instead add a new key for Secrets and ConfigMaps which allow users to include all keys in the target Secret and ConfigMap
2) We'll add a special case where this new key is disallowed for `tls` type Secrets

Thanks for raising this Julio!

PR #460, opened 2024-10-22
Support for including all certificates from a secret/configMap in bundle source

juliocamarero:
closes: https://github.com/cert-manager/trust-manager/issues/458

### Pull Request Motivation

This add support for including in a bundle not just one certificate referenced by a key in a configMap or secret, but all of the certificates included in the configMap or secret. 

This is how we currently include a bundle a source PEM from a secret referenced by key:
```yaml
kind: Bundle
metadata:
  name: my-org.com  
spec:
  sources:
  - secret:
      name: "my-certs"
      key: "ca.crt"
```

And this is how we could include all the certificates in a secret or configMap after these changes:
```yaml
kind: Bundle
metadata:
  name: my-org.com  
spec:
  sources:
  - secret:
      name: "my-certs"
      includeAllKeys: true
 - configMap:
      name: "more-certs"
      includeAllKeys: true
```

We discussed the usage of regular expression instead in https://github.com/cert-manager/trust-manager/pull/459 but it was discarded for the complexity added to end users.

We also decided to prevent users from using `includeAllKeys` on secrets of type TLS since that would include the private key.

### Kind
 
/kind feature

### Release Note
```
Add support for including in a bundle not just one certificate referenced by a key in a configMap or secret, but all of the certificates included in the configMap or secret. 
```


SgtCoDFish (approved):
/lgtm
/approve

I think this looks like a great improvement, thank you! 🚀 

Really great to have you involved with the project 😁

SgtCoDFish:
@juliocamarero thanks again for this!

Do you fancy updating https://github.com/cert-manager/website/blob/master/content/docs/trust/trust-manager/README.md to add some quick docs around this?

juliocamarero:
> @juliocamarero thanks again for this!
> 
> Do you fancy updating https://github.com/cert-manager/website/blob/master/content/docs/trust/trust-manager/README.md to add some quick docs around this?

Absolutely, I'm on it! :)

juliocamarero:
Here is the PR updating the website docs: https://github.com/cert-manager/website/pull/1594 
Thanks again folks! 🙇

PR #468, opened 2024-11-05
WIP: POC of dynamic authority

erikgb:
This PR was created to show how trust-manager could look like without a requirement for cert-manager to be pre-installed. **It is not supposed to be merged** - at least not in the present state.

Inspired by the bootstrap mechanism cert-manager uses to bootstrap it's webhooks, I created https://github.com/erikgb/dynamic-authority. I have no plans to maintain such a project alone, and most of the code comes from cert-manager, so I hope this eventually could become a new project under the cert-manager org. 😺 

I also had to copy some awesome PKI code from cert-manager into the new project. IMO we should create a dedicated cert-manager PKI project with all the nice PKI helpers battle-tested as part of cert-manager. This could allow reuse across other cert-manager projects without the unfortunate dependency to cert-manager/cert-manager.


erikgb:
@SgtCoDFish I have now reconfigured the RBAC to (almost) least-privilege, as you commented in one of our stand-ups where we discussed this PR. The new control loops still need RBAC to read all validation webhook configurations (cluster-wide) and secrets in the controller namespace (it already has this permission). But the create/patch permissions are now targeted on resources by name. I personally don't see a big benefit of narrowing it even more, as I have a feeling controller-runtime doesn't support watching resources by name out-of-the-box and will require even more complex code in the new module.

Issue #453, opened 2024-10-11
[Chart] Add SecretTemplate for Certificate generated for webhook


guntiskarulis:
Please add an option to include a `spec.SecretTemplate` for the webhook Certificate. Currently, it will create a secret without any labels or annotations, which can be troublesome if we have admission controllers that do not allow mounting secrets without matching labels.

PR #469, opened 2024-11-05
feat(helm): add secretTemplate-ing to certificate

guntiskarulis:
**Pull Request Motivation**
Some environments have admission webhooks that might mandate specific labels or annotations. This change allows chart users to set labels and/or annotations in the Secret created by the Certificate resources.

Close https://github.com/cert-manager/trust-manager/issues/453

### DIFF
--- a/deploy/charts/trust-manager/templates/certificate.yaml
+++ b/deploy/charts/trust-manager/templates/certificate.yaml
@@ -24,6 +24,10 @@ spec:
   dnsNames:
   - "{{ include "trust-manager.name" . }}.{{ include "trust-manager.namespace" . }}.svc"
   secretName: {{ include "trust-manager.name" . }}-tls
+  {{- with .Values.app.webhook.tls.certificate.secretTemplate }}
+  secretTemplate:
+    {{- toYaml .| nindent 4 }}
+  {{- end }}
   revisionHistoryLimit: 1
   issuerRef:
     name: {{ include "trust-manager.name" . }}
--- a/deploy/charts/trust-manager/values.yaml
+++ b/deploy/charts/trust-manager/values.yaml
@@ -251,6 +251,17 @@ app:
 
         # The name of cert-manager's Service Account. Only used if `app.webhook.tls.approverPolicy.enabled` is true.
         certManagerServiceAccount: "cert-manager"
+
+      # Add labels/annotations to secrets created by Certificate resources when using cert-manager provisioned TLS certificate.
+      certificate:
+        secretTemplate: {}
+          # For example:
+          #   annotations:
+          #     my-secret-annotation-1: "foo"
+          #     my-secret-annotation-2: "bar"
+          #   labels:
+          #     my-secret-label: foo
+
     # This value specifies if the app should be started in hostNetwork mode. It is required for use in some managed Kubernetes clusters (such as AWS EKS) with custom CNI.
     hostNetwork: false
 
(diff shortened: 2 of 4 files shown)


deploy/charts/trust-manager/templates/certificate.yaml
```diff
@@ -24,6 +24,8 @@ spec:
   dnsNames:
   - "{{ include "trust-manager.name" . }}.{{ include "trust-manager.namespace" . }}.svc"
   secretName: {{ include "trust-manager.name" . }}-tls
+  secretTemplate:
+    {{- toYaml .Values.certificates.certManager.secretTemplate | nindent 4 }}
```
erikgb, line 28: Maybe add a conditional around this, so we avoid setting `secretTemplate` to an empty dict by default?

guntiskarulis: Thanks! Added condition around secretTemplate and fixed value ref.

deploy/charts/trust-manager/templates/certificate.yaml
```diff
@@ -24,6 +24,10 @@ spec:
   dnsNames:
   - "{{ include "trust-manager.name" . }}.{{ include "trust-manager.namespace" . }}.svc"
   secretName: {{ include "trust-manager.name" . }}-tls
+  {{- if and (hasKey .Values.app.webhook.tls "secretTemplate") .Values.app.webhook.tls.secretTemplate }}
+  secretTemplate:
+    {{- toYaml .Values.app.webhook.tls.secretTemplate | nindent 4 }}
+  {{- end }}
```
erikgb, line 30:
```suggestion
  {{-  with .Values.app.webhook.tls.secretTemplate }}
  secretTemplate:
    {{- toYaml . | nindent 4 }}
  {{- end }}
```

I think something like this should work. Please verify with "helm template" locally. :-)

guntiskarulis: yes, way better.

deploy/charts/trust-manager/values.yaml
```diff
@@ -251,6 +251,16 @@ app:
 
         # The name of cert-manager's Service Account. Only used if `app.webhook.tls.approverPolicy.enabled` is true.
         certManagerServiceAccount: "cert-manager"
+
+      secretTemplate: {}
+        # Add labels/annotations to secrets created by Certificate resources
```
erikgb, line 256: I think the comments should be above the value, and not below. At least the description. Note that we generate the schema and docs.

guntiskarulis: Alright, will move description of the key above and yes, I just noticed about schema and docs.

deploy/charts/trust-manager/templates/certificate.yaml
```diff
@@ -24,6 +24,10 @@ spec:
   dnsNames:
   - "{{ include "trust-manager.name" . }}.{{ include "trust-manager.namespace" . }}.svc"
   secretName: {{ include "trust-manager.name" . }}-tls
+  {{- with .Values.app.webhook.tls.secretTemplate }}
```
erikgb, line 27:
```suggestion
  {{- with .Values.app.webhook.tls.certificate.secretTemplate }}
```

Thinking a bit more about this, I think we should make it clear that this only applies if using cert-manager to provision webhook TLS certificate. We already have a Helm alternative, and I think we might add other options later on.

deploy/charts/trust-manager/values.yaml
erikgb: You might have to try a bit locally and see that the schema and docs are generated correctly. :-)

deploy/charts/trust-manager/templates/certificate.yaml
guntiskarulis: Yep, I agree with you on this. I will update to be more explicit.

erikgb:
Thanks @guntiskarulis! 🚀 

/lgtm

PR #474, opened 2024-11-14
Fix crd validation: Key should not be required anymore in bundle source resources

juliocamarero:
Hi folks, 

in my last contribution (https://github.com/cert-manager/trust-manager/pull/460) I made a mistake.

I didn't realize the "key" when specifying a source secret with "includeAllKeys" was still required in the CRD. 
This causes the webhook to reject bundles which use the new field:
```
validatingwebhookconfiguration.admissionregistration.k8s.io/trust-manager configured
The Bundle "ecp-internal-cas" is invalid: spec.sources[1].secret.key: Required value
```


pkg/apis/trust/v1alpha1/types_bundle.go
```diff
@@ -169,9 +169,9 @@ type SourceObjectKeySelector struct {
 	//+optional
 	Selector *metav1.LabelSelector `json:"selector,omitempty"`
 
-	// KeySelector is the key of the entry in the objects' `data` field to be referenced.
+	// Key is the key of the entry in the object's `data` field to be used.
 	//+optional
-	KeySelector `json:",inline,omitempty"`
+	Key *string `json:"key"`
```
erikgb, line 174:
```suggestion
	Key *string `json:"key,omitempty"`
```

What's your motivation for renaming this Go field? I don't think it's required to do, and I assume there was a reason for the extra abstraction when this field was introduced.

erikgb (commented):
Thanks for coming back to fix this, @juliocamarero! I was expecting a minor fix, so my main question is why you are suggesting to rename a (Go) field.

pkg/apis/trust/v1alpha1/types_bundle.go
juliocamarero: Thanks a lot for having a look so fast @erikgb 🙇 

> What's your motivation for renaming this Go field?

It's not just a renaming. I am using a string pointer directly instead of using a struct which embeds a string.
If you look at the autogenerated changes in the CRD, the only change is that now the `key` field is not required anymore, there isn't any field renamed there.

Let me explain the problem as I see it:
- The struct `SourceObjectKeySelector` doest not have an explicit `Key`field, instead, it has a  `KeySelector` field of type struct which contains a `Key`field, and the `KeySelector` field is annotated with `json:inline` which makes the field `key`to be implicetly a field the `SourceObjectKeySelector` in json. 
- We want the `key` field of the struct `SourceObjectKeySelector` in json to be optional (because it shouldn't be used when `includeAllKeys: true`
- The field `KeySelector` in the `SourceObjectKeySelector` was already marked as optional. However, since `Key` is required inside the KeySelector struct, when the CRD is autogenerated, the field `key` is always marked as required. 

These are the different approaches I tried:
- changing the `KeySelector` to `*KeySelector`in the `SourceObjectKeySelector`, this didn't change anything in the CRD, `key` was still required
- changing the `Key` inside the `KeySelector` struct to be optional. This change made the `key` field optional in the CRD, but not only for the `SourceObjectKeySelector`, but also for several other models for which it shouldn't be optional.

>  I was expecting a minor fix

I was expecting a minor fix too 😅 Happy to try any other suggestion you may have :) (hopefully there is a secret trick to make this field not required in the CRD which I am not aware of 🪄 )

I only could think of another approach, which is to introduce a new struct called `OptionalKeySelector` to be used instead of the existing `KeySelector`in the `SourceObjectKeySelector`. It would just contain the string pointer `Key`. I couldn't think of any benefit of having this additional struct for wrapping a string pointer, so that's why I used the `Key` directly.

Thanks again Erik!!  🙇

juliocamarero: You are absolutely right! I hadn't thought of this 🙈 
This simplifies the logic and reduces the amount of changes need 👌

pkg/apis/trust/v1alpha1/types_bundle.go
```diff
@@ -169,9 +169,9 @@ type SourceObjectKeySelector struct {
 	//+optional
 	Selector *metav1.LabelSelector `json:"selector,omitempty"`
 
-	// KeySelector is the key of the entry in the objects' `data` field to be referenced.
+	// Key is the key of the entry in the objects' `data` field to be referenced.
```
erikgb, line 172:
```suggestion
	// Key of the entry in the object's `data` field to be used.
```

Possible better aligned with the new `IncludeAllKeys` field?

erikgb (commented):
Looks great @juliocamarero! Just a minor suggestion to the field description. Let me know what you think. Can you update the PR title and squash everything into a single commit before we merge?

pkg/apis/trust/v1alpha1/types_bundle.go
juliocamarero: Perfect, done! I've squased all commits into one below: https://github.com/cert-manager/trust-manager/pull/474/commits/18ba82c5d77f26cc3a8675d18ddd4121fee657d6

erikgb (commented):
Thanks for this important fix @juliocamarero! It was required to change more code than anticipated, but I think the changes represented a simplification. 🚀 

/lgtm
/approve

Issue #302, opened 2024-02-22
Add matchExpressions to Bundle's spec.target.namespaceSelector


sathieu:
Currently, bundle has:

```yaml
spec:
  target:
    namespaceSelector:
      matchLabels:
        linkerd.io/inject: "enabled"
```

We need support for:

```yaml
spec:
  target:
    namespaceSelector:
      matchExpressions:
      - key: kubitus-project.gitlab.io/inject-ca-bundle
        operator: NotIn
        values:
        - 'false'

```

Ref: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#resources-that-support-set-based-requirements

erikgb:
I think this was discussed in https://github.com/cert-manager/trust-manager/pull/258 - as this new feature supports all types of selectors. It would make sense to add this IMO, and I also think it could be added non-breaking. PRs welcome! 😸

sathieu:
@ocampeau Could you please take a look at this feature request :pray:?

ocampeau:
I think it could be a nice addition yes. This can definitely be done in a non-breaking fashion. The Kubernetes libraries already provides the types for that.

inteon:
> I think this was discussed in #258 - as this new feature supports all types of selectors. It would make sense to add this IMO, and I also think it could be added non-breaking. PRs welcome! 😸

@erikgb do you have a link to the thread in that PR where this was discussed?

erikgb:
> > I think this was discussed in #258 - as this new feature supports all types of selectors. It would make sense to add this IMO, and I also think it could be added non-breaking. PRs welcome! 😸
> 
> @erikgb do you have a link to the thread in that PR where this was discussed?

@inteon https://github.com/cert-manager/trust-manager/pull/258#discussion_r1444016914

PR #476, opened 2024-11-16
feat: support namespaceSelector matchExpressions

erikgb:
And use upstream code to validate label selectors. I am in doubt if additional tests are required. Somehow it feels wrong to test upstream features here. They are certainly tested upstream. WDYT?

Close https://github.com/cert-manager/trust-manager/issues/302


inteon:
Great expansion of the existing functionality + I like that we use the upstream resources and validation more.
/lgtm
/approve
/hold Could you add [here](https://github.com/cert-manager/trust-manager/issues/302#issuecomment-2490955377) a link to the original discussion around this feature? I'm guessing we did not initially add this just so the scope was a bit more limited?

PR #477, opened 2024-11-17
Improve API specification

erikgb:
The proposed changes are extracted from https://github.com/cert-manager/trust-manager/pull/475. Before introducing CEL in the API specification, which will lift the K8s version requirements, I propose adding more "simple" validations to the API spec. Also adding processing instructions to API to ensure it will work smoothly for users using SSA.


docs/api/api.md
```diff
@@ -365,13 +367,16 @@ func (in *BundleSource) DeepCopyInto(out *BundleSource)
...
+## type [BundleSpec](<https://github.com/cert-manager/trust-manager/blob/main/pkg/apis/trust/v1alpha1/types_bundle.go#L60-L69>)
 
 BundleSpec defines the desired state of a Bundle.
 
 ```go
 type BundleSpec struct {
     // Sources is a set of references to data whose data will sync to the target.
+    // +listType=atomic
+    // +kubebuilder:validation:MinItems=1
+    // +kubebuilder:validation:MaxItems=100
```
erikgb, line 379: MaxItems is added to pass the cost requirements when CEL validations are added to the API.

pkg/apis/trust/v1alpha1/types_bundle.go
```diff
@@ -59,6 +59,9 @@ type BundleList struct {
 // BundleSpec defines the desired state of a Bundle.
 type BundleSpec struct {
 	// Sources is a set of references to data whose data will sync to the target.
+	// +listType=atomic
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=100
```
erikgb, line 64: MaxItems is added to pass the cost requirements when CEL validations are added to the API.

pkg/apis/trust/v1alpha1/types_bundle.go
```diff
@@ -67,6 +70,7 @@ type BundleSpec struct {
 
 // BundleSource is the set of sources whose data will be appended and synced to
 // the BundleTarget in all Namespaces.
+// +structType=atomic
```
inteon, line 73: 👍

pkg/apis/trust/v1alpha1/types_bundle.go
```diff
@@ -96,14 +100,17 @@ type BundleSource struct {
 
 // BundleTarget is the target resource that the Bundle will sync all source
 // data to.
+// +structType=atomic
```
inteon, line 103: 👍

inteon, line 103: @erikgb I don't think this has to be atomic, right?

pkg/apis/trust/v1alpha1/types_bundle.go
```diff
@@ -117,16 +124,21 @@ type BundleTarget struct {
 }
 
 // AdditionalFormats specifies any additional formats to write to the target
+// +structType=atomic
```
inteon, line 127: @erikgb same here, does not have to be atomic, right?

pkg/apis/trust/v1alpha1/types_bundle.go
erikgb: Agree, will fix.

pkg/apis/trust/v1alpha1/types_bundle.go
```diff
@@ -171,6 +187,7 @@ type SourceObjectKeySelector struct {
 
 	// Key of the entry in the object's `data` field to be used.
 	//+optional
+	// +kubebuilder:validation:MinLength=1
```
inteon, line 190: @erikgb what does the combination `optional` and `MinLength=1` mean?
Do you have examples of yaml that works and does not work with this configuration + how it is different from eg. just `optional`?

erikgb: Just `optional` will allow the empty string. Which doesn't make sense here.

pkg/apis/trust/v1alpha1/types_bundle.go
erikgb: Fixed

pkg/apis/trust/v1alpha1/types_bundle.go
erikgb: Fixed

inteon:
Great work, this is something that could be useful in our other APIs too.

/approve
/lgtm

erikgb:
> Great work, this is something that could be useful in our other APIs too.

Yeah, I am going on a "round around the block". But I need static CRDs in Git first. 😉

PR #478, opened 2024-11-17
WIP: Remove incomplete validation from webhook

erikgb:
While working on https://github.com/cert-manager/trust-manager/pull/475, I noticed this validation that seems wrong to me, at least incomplete. Why shouldn't a user be allowed to use the same key in source and target? This will only become a potential problem if the "trust namespace" is included as a target namespace. And why would you do that? And after https://github.com/cert-manager/trust-manager/pull/460, this check is definitely incomplete - which means we have to find a better way to exclude sources as targets in the controller.

### DIFF
--- a/pkg/webhook/validation.go
+++ b/pkg/webhook/validation.go
@@ -164,24 +164,6 @@ func (v *validator) validate(obj runtime.Object) (admission.Warnings, error) {
 		))
 	}
 
-	if target := bundle.Spec.Target.ConfigMap; target != nil {
-		path := path.Child("sources")
-		for i, source := range bundle.Spec.Sources {
-			if source.ConfigMap != nil && source.ConfigMap.Name == bundle.Name && source.ConfigMap.Key == target.Key {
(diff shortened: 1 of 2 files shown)


pkg/webhook/validation.go
```diff
@@ -164,24 +164,6 @@ func (v *validator) validate(obj runtime.Object) (admission.Warnings, error) {
 		))
 	}
 
-	if target := bundle.Spec.Target.ConfigMap; target != nil {
-		path := path.Child("sources")
-		for i, source := range bundle.Spec.Sources {
-			if source.ConfigMap != nil && source.ConfigMap.Name == bundle.Name && source.ConfigMap.Key == target.Key {
```
inteon, line 170: We don't just check that the keys match, we also check that the bundle name matches the source config map name.

erikgb: Thanks for pointing this out, @inteon. I will make an effort to fix the currently broken validation smarter.

PR #480, opened 2024-11-19
Fix issue with updating of JKS/PKCS targets when password changes #449 (review)

inteon:
Reviewed https://github.com/cert-manager/trust-manager/pull/449 and added 2 commits (see their description for details):
- https://github.com/cert-manager/trust-manager/commit/8c74647dfb25181793093455f8e12c3373e76106
- https://github.com/cert-manager/trust-manager/commit/9e9918cba880421ac267b0527e230ed1b4a71222

Can be added to the original PR or can be merged directly, made sure to keep [@arsenalzp](https://github.com/arsenalzp) as the original PR on the commit.

fixes https://github.com/cert-manager/trust-manager/issues/433
supersedes https://github.com/cert-manager/trust-manager/pull/449


pkg/bundle/internal/target/target.go
```diff
@@ -396,3 +397,21 @@ func (b *Data) Populate(pool *util.CertPool, formats *trustapi.AdditionalFormats
 	}
 	return nil
 }
+
+func TruststoreHash(data []byte, bundle *trustapi.AdditionalFormats) string {
```
erikgb, line 401:
```suggestion
func TrustBundleHash(data []byte, bundle *trustapi.AdditionalFormats) string {
```

I would suggest a more generic function name, as the hash is also used when no additional formats used. IMO a truststore is a binary format. The term trust bundle covers both PEM and additional formats (truststores).

erikgb, line 401: Now that you suggest using this function in tests, I think we need a unit test of the function. I believe it is now possible to change the function behavior without breaking the tests.

erikgb (commented):
Thanks for adding a test for the fixed issue @inteon. But I am not sure if I like the use of the function currently named `TruststoreHash` for test assertions.

pkg/bundle/internal/target/target.go
inteon: I updated the name.

inteon: I added a unit test.

erikgb (commented):
One (hopefully final) suggestion suggestion for renaming vars. Feel free to unhold!

/lgtm
/approve
/hold

Maybe also update the PR description - so that merging this PR will auto-close the issue and original PR by @arsenalzp?

PR #482, opened 2024-11-21
Check-in testdata such that test can be run in IDE/ using go without using Makefile

inteon:
Alternative solution for https://github.com/cert-manager/makefile-modules/pull/209.
Makes it possible to run Go tests without Makefile (eg. when running using IDE).

Checks-in the testdata instead of generating on demand as part of the Makefile target.

cc @erikgb

### DIFF
--- a/make/test-integration.mk
+++ b/make/test-integration.mk
@@ -12,17 +12,19 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-trust_manager_crds := $(bin_dir)/scratch/trust.cert-manager.io_bundles.yaml
-$(trust_manager_crds): $(helm_chart_archive) | $(NEEDS_HELM) $(NEEDS_YQ)
+trust_manager_crds := test/testdata/trust.cert-manager.io_bundles.yaml
+
+generate-testdata: $(helm_chart_archive) | $(NEEDS_HELM) $(NEEDS_YQ)
 	$(HELM) template test "$(helm_chart_archive)" | \
-		$(YQ) e '. | select(.kind == "CustomResourceDefinition")' \
-		> $@
+		$(YQ) e '. | select(.kind == "CustomResourceDefinition") | del(.metadata.labels)' \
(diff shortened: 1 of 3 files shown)


erikgb:
The proposed changes will solve my immediate issues, but I am bit worried about the make target dependencies. If I modify the API in Go source code or modify the "testdata" CRD and then run `make generate-crds`, the "testdata" CRD is not updated. This could be not very clear for users. Do you see a way to fix this with your proposed solution?

We should at least ensure that the "testdata" CRD is up-to-date in CI, but that's maybe OK? Not sure how the verfy-generated works.

inteon:
> The proposed changes will solve my immediate issues, but I am bit worried about the make target dependencies. If I modify the API in Go source code or modify the "testdata" CRD and then run `make generate-crds`, the "testdata" CRD is not updated. This could be not very clear for users. Do you see a way to fix this with your proposed solution?
> 
> We should at least ensure that the "testdata" CRD is up-to-date in CI, but that is maybe OK? Not sure how the verfy-generated mechanism actually works.

When I update eg. `types_bundle.go` and run `make generate`, I get the following result:
```
$ git status
On branch simplify_go_test
Your branch is up to date with 'inteon/simplify_go_test'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   deploy/charts/trust-manager/templates/crd-trust.cert-manager.io_bundles.yaml
        modified:   pkg/apis/trust/v1alpha1/types_bundle.go
        modified:   test/testdata/trust.cert-manager.io_bundles.yaml

no changes added to commit (use "git add" and/or "git commit -a")
```

Both the `deploy/charts/trust-manager/templates/crd-trust.cert-manager.io_bundles.yaml` and `test/testdata/trust.cert-manager.io_bundles.yaml` file got updated + CI should check that they are kept up-to-date.

Issue #63, opened 2022-10-27
nit: Rename "Bundle" to "ClusterBundle"


evankanderson:
Since Bundle is cluster-scoped, ClusterBundle seems like it would be the proper name (which seems correct, given that `spec.target.namespaceSelector` suggests that the Bundle should be at a Cluster scope or need an "AcceptBundle" object in each namespace to indicate that the namespace wants to receive the certs).

SgtCoDFish:
Agree with `ClusterBundle` being a better match for current behaviour, although IIRC from past conversations amongst the maintainer team I've made the case for us to change Bundle to default to only write the target in a single namespace, with the option to write to all.

In any case, I think the behaviour of writing to all namespaces shouldn't be the default and I feel like `Bundle` would fit pretty well in that case. What do you think?

(Thank you for your contribution!)

ProbstDJakob:
I think `ClusterBundle` is still a better name since it does not belong to a single namespace. Furthermore this would allow adding a `Bundle` CRD which is bound to a namespace later. Much like `ClusterIssure` and `Issuer`. (See also https://github.com/cert-manager/trust-manager/issues/131#issuecomment-1656545012)

SgtCoDFish:
(Added to milestone, annoying that there's another group for it!)

FWIW I agree about Bundle / ClusterBundle now. I see the use case and I think it's worth renaming the resource we currently have to ClusterBundle. I also see the value in adding Bundle as a namespace-scoped resource, but I'm currently inclined to think we don't need to do both things at once.

I suspect we'll make the rename from Bundle -> ClusterBundle in v1alpha2 or v1beta1, which is on the roadmap after us hearing so much demand for trust-manager v1 at KubeCon NA in Chicago!

sftim:
Consider if you want to emit [warnings](https://kubernetes.io/blog/2020/09/03/warnings/) about people using the legacy Bundle name (assuming you can tell - I think you can).

I'd use a ValidatingAdmissionPolicy to actually emit the warnings.

Issue #242, opened 2023-11-23
New version of Bundle API


erikgb:
We have already discussed some breaking changes we are considering making to the Bundle API. This issue is proposed as an umbrella issue to agree on what to do and how to proceed. Browsing through our currently open issues, I can find the following candidates:

- [ ] #4 
- [ ] #39 
- [ ] #63

Our first decision should be if we want to do #63 - as it suggests renaming the kind from `Bundle` to `ClusterBundle`. And AFAIK it is not possible to have any conversion webhook to convert between kinds. A conversion webhook can only convert between versions of the same kind. Doing #63 will create space to eventually fix #131 in the future. I am not saying we should do #131, but without #63 it will not be possible to do #131 with good naming of resources.

Over to how I think we should introduce a version (or kind). Either way, I suggest starting by introducing a new version (in a dedicated PR) as an exact copy of the existing API version, and make the new version unserved. As long as the new version is unserved we can now proceed changing the new version (and/or kind) in multiple PRs.

kdubb:
@erikgb We created a `CertificateBundler` CRD and operator that is nearly identical to trust manager's basic `Bundle` with a couple major differences that help support rotation.

The operator for the bundler reads sources, builds PEM & JKS bundles and then creates `ConfigMap`s in target namespaces. We create Certificate style `Secret`s only (`tls.crt` with PEM) for ease and it fits our use cases.

### 1. Append Certificates
The status of `CertificateBundler` caches the current set of certs (as a PEM bundle). When reconciling the CRD (e.g. due to change in a source `Secret`) we scan the sources and append any new certificates, leaving old ones in place.

### 2. Filter Expired Certificates
During reconciliation, when building the certificate bundle, any expired certificates are excluded. Additionally, we schedule an appropriately timed reconciliation if any certificates expire sooner than the normal 10hour drift reconcile.

### 3. Deduplication
Duplicate certificates are removed by comparing `Issuer + Subject + Serial Number`; although `Issuer + Serial Number` seems to actually be the right unique key.

### Implementation

Our actual implementation goes like this:

1) Scan sources for certificates
2) Append current certificates (from cached)
3) Filter expired certificates
4) Deduplicate certificates
5) Cache bundle in status
6) Generate or update config maps in target namespaces
7) Schedule reconcile for earliest expiring certificate (if needed)

[...]

PR #485, opened 2024-11-24
Migration Design: Rename Bundle to ClusterBundle

erikgb:
Proposal on how to implement https://github.com/cert-manager/trust-manager/issues/63.

The proposed solution is based on [input from "the greater community"](https://kubernetes.slack.com/archives/C0EG7JC6T/p1732370214737539). 🥇

### DIFF
--- a/design/20241124-rename-bunde-to-clusterbundle.md
+++ b/design/20241124-rename-bunde-to-clusterbundle.md
@@ -28,7 +28,7 @@ TODO: add more items to checklist
 
 We want to rename `Bundle` to `ClusterBundle` for reasons described in "Motivation" below.
 
-Conversion webhooks in Kubernetes can only be used to convert between versions of and API,
+Conversion webhooks in Kubernetes can only be used to convert between versions of an API,
 and not to convert between resources. This makes renaming harder to do, as it will represent
 deleting the old API and creating a new API.
 
--- a/design/20241124-rename-bunde-to-clusterbundle.md
+++ b/design/20241124-rename-bunde-to-clusterbundle.md
@@ -104,7 +104,10 @@ When the functional controller is migrated to target `ClusterBundle`, we want to
 controller targeting the now deprecated `Bundle` resource.
 
 This controller should just create or update the `ClusterBundle` with the same name as the reconciled `Bundle`.
-Since the APIs should have identical `spec`, this should be a simple controller.
+Since the APIs should be quite similar, this should be a simple controller.
+
+- `Bundle` `.spec` should be copied (and eventually slightly converted) into `ClusterBundle` `.spec`
+- `Bundle` `.status` should be updated from `ClusterBundle` `.status`
 
 To allow a user to migrate their resources from `Bundle` to `ClusterBundle`, the controller should not act
 on `Bundle` delete events - nor should it add an owner reference to `ClusterBundle` to avoid the Kubernetes garbage collector
--- a/design/20241124-rename-bunde-to-clusterbundle.md
+++ b/design/20241124-rename-bunde-to-clusterbundle.md
@@ -51,11 +51,14 @@ The name would also be better aligned with upstream [ClusterTrustBundle](https:/
 
 - `Bundle` resource is renamed to `ClusterBundle`
 - Provide a smooth migration for users of trust-manager
+- Use the opportunity window to improve the API
+  - [Issue summarizing proposed API changes (#242)](https://github.com/cert-manager/trust-manager/issues/242)
+  - [Draft PR for proposed API changes done when migrating (#486)](https://github.com/cert-manager/trust-manager/pull/486)
 
 
 ### Non-Goals
 
-- Change/fix/improve `Bundle`/`ClusterBundle` API while renaming
+- Behavior changes in `Bundle`/`ClusterBundle` API while migrating (only simpler API improvements)
 - Introduce a new API version
 - Introduce a namespace-scoped bundle API (at least not now)
 
@@ -66,6 +69,7 @@ We propose to do this change over multiple releases:
 1. Version X: Only `Bundle` API is served and has the functional controller. (present state)
 2. Version X+1: We introduce `ClusterBundle` with API otherwise identical to `Bundle`:
    - `Bundle` is marked as deprecated.
+   - `ClusterBundle` introduced with agreed API improvements.
    - Functional controller is migrated to `ClusterBundle`
    - Validating webhook for both `Bundle` and `ClusterBundle` with same validations
    - New technical controller added for `Bundle`, converting bundles to cluster bundles
--- a/design/20241124-rename-bunde-to-clusterbundle.md
+++ b/design/20241124-rename-bunde-to-clusterbundle.md
@@ -140,8 +140,13 @@ TODO
 
 ### Upgrade / Downgrade Strategy
 
-Downgrading from a version where the migration has started will probably not be possible and should
-be clearly documented and stated in the release notes.
+Upgrading should represent less of challenge if we ensure enough release between version X+1 and Y.
+We should document that upgrading directly from version X (or earlier) to Y is explicitly **NOT** supported.
+
+Downgrading to version X (or earlier) from a version where the migration has started could be problematic.
+We can probably mitigate this risk by keeping owner references (non-controller, as Kubernetes doesn't allow
+multiple controller references) from `Bundle` to target resources, in addition to the controller references from
+`ClusterBundle`, during the migration period.
 
 ## Alternatives
 
--- a/design/20241124-rename-bunde-to-clusterbundle.md
+++ b/design/20241124-rename-bunde-to-clusterbundle.md
@@ -156,6 +156,27 @@ multiple controller references) from `Bundle` to target resources, in addition t
 
 ## Alternatives
 
+### Introduce `ClusterBundle` as a new and Independent API
+
+Since a `Bundle` and a `ClusterBundle` could then address the same target resources (by name), we would have to
+take this into account in controllers. In addition, we would have a **lot** more code to maintain and test.
+This could be partly mitigated by creating reusable abstractions that work for both resources.
+But if we want to allow `ClusterBundle` to provide new features, this would become complex.
+
+The migration path from `Bundle` to `ClusterBundle` would also be unclear and complex.
+Use case: As a user, I would like to use `ClusterBundle` instead of `Bundle` **without interrupting the target resources**.
+
+```
+Bundle -> ClusterBundle -> target resources
+```
+
+seems a lot simpler to manage than
+
+```
+Bundle -> target resources AND
+ClusterBundle -> target resources
+```
+
 ### Just rename resource between releases
 
 Since the `Bundle` API version is `v1alpha1`, we could justify just doing the simplest thing and rename.
--- a/design/20241124-rename-bunde-to-clusterbundle.md
+++ b/design/20241124-rename-bunde-to-clusterbundle.md
@@ -118,6 +118,61 @@ It should be possible to read the field manager name used (by user) to create/up
 If we use the same field manager name in controller (per resource) to create/update `ClusterBundle`, we think it's less
 likely that users will have to force conflicts when migrating from `Bundle` to `ClusterBundle`.
 
+### Proposed API changes when migrating
+
+Improving the trust-manager API while migrating is not something we have to do, but
+appears like nice opportunity to fix some of the long-standing proposed API changes.
+
+There is an [open draft PR for proposed (and breaking) API changes](https://github.com/cert-manager/trust-manager/pull/486),
+but in order to make this a non-braking change for users, only the `ClusterBundle` API will include these changes.
+
+While this is subject to change, an example on how a user would migrate from `Bundle` to `ClusterBundle` without
+changing anything for target resources is shown below:
+
+```yaml
+---
+apiVersion: trust.cert-manager.io/v1alpha1
+kind: Bundle
+metadata:
+  name: my-bundle
+spec:
+  sources:
+    - configMap:
+        name: my-ca
+        key: ca.crt
+  target:
+    configMap:
+      key: ca.crt
+    additionalFormats:
+      jks:
+        key: bundle.jks
+      pkcs12:
+        key: bundle.p12
+```
+
+Which would be migrated to:
+
+```yaml
+---
+apiVersion: trust-manager.io/v1alpha1
+kind: ClusterBundle
+metadata:
+  name: my-bundle
+spec:
+  sources:
+    - configMap:
+        name: my-ca
+        key: ca.crt
+  target:
+    configMap:
+      - key: ca.crt
+      - key: bundle.jks
+        format: JKS
+      - key: bundle.p12
+        format: PKCS12
+    namespaceSelector: {} # Selecting all namespaces here, but now required in API
+```
+
 ### Risks and Mitigations
 
 #### Target configmaps/secrets are accidentally deleted
--- a/design/20241124-rename-bunde-to-clusterbundle.md
+++ b/design/20241124-rename-bunde-to-clusterbundle.md
@@ -47,9 +47,15 @@ for trust bundle management inside namespaces in multi-tenant clusters.
 In addition, `ClusterBundle` would be a better name looking at cert-manager having `Issuer` and `ClusterIssuer`.
 The name would also be better aligned with upstream [ClusterTrustBundle](https://kubernetes.io/docs/reference/access-authn-authz/certificate-signing-requests/#cluster-trust-bundles).
 
+By pushing users to migrate smoothly to a new API, we can perform requested API changes at the same time.
+Consider it similar to a one-way conversion webhook.
+
+Since trust-manager is maturing and getting more attention, we suggest establishing its own API group: `trust-manager.io`.
+
 ### Goals
 
 - `Bundle` resource is renamed to `ClusterBundle`
+- `ClusterBundle` is created in the new `trust-manager.io` API group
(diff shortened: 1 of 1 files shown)


erikgb:
/hold

We want to discuss this proposal in our next bi-weekly meeting and allow some time to add missing details.

damsien:
The design looks good to me 👍
I wanted to open a discussion about the migration path because trust-manager uses helm chart to manage installation & upgrades.

Suppose that an user is currently using the version X. We introduce the version X+1 and then Y. After a few weeks, the user do an `helm repo udpate` and sees that the version Y is available. If he directly upgrade to the version Y, he will skip the patch and all of its Bundles will not be served and unusable.

erikgb:
Thanks for reviewing @damsien! ❤️ 

> Suppose that an user is currently using the version X. We introduce the version X+1 and then Y. After a few weeks, the user do an `helm repo udpate` and sees that the version Y is available. If he directly upgrade to the version Y, he will skip the patch and all of its Bundles will not be served and unusable.

That's why I named version Y, and not version X+2. 😉 You are addressing a real risk, and we should allow some time (maybe many releases) for users to migrate from `Bundle` to `ClusterBundle`. And clearly document the required upgrade path.

design/20241124-rename-bunde-to-clusterbundle.md
```diff
@@ -0,0 +1,158 @@
...
+- [ ] Test plan has been agreed upon and the tests implemented
+- [ ] User-facing documentation has been PR-ed against the release branch in [cert-manager/website]
+
+TODO: add more items to checklist
+
+## Summary
+
+We want to rename `Bundle` to `ClusterBundle` for reasons described in "Motivation" below.
+
+Conversion webhooks in Kubernetes can only be used to convert between versions of and API,
```
evankanderson, line 31:
```suggestion
Conversion webhooks in Kubernetes can only be used to convert between versions of an API,
```

design/20241124-rename-bunde-to-clusterbundle.md
```diff
@@ -0,0 +1,158 @@
...
+
+### Goals
+
+- `Bundle` resource is renamed to `ClusterBundle`
+- Provide a smooth migration for users of trust-manager
+
+
+### Non-Goals
+
+- Change/fix/improve `Bundle`/`ClusterBundle` API while renaming
```
evankanderson, line 58: If there are already known improvements, there's no particular reason to block them during this rename -- you can add them to the new ClusterBundle resource and adapt the Bundle shape as part of the migration.

design/20241124-rename-bunde-to-clusterbundle.md
```diff
@@ -0,0 +1,158 @@
...
+
+## Design Details
+
+### Test Plan
+
+TODO
+
+### Upgrade / Downgrade Strategy
+
+Downgrading from a version where the migration has started will probably not be possible and should
```
evankanderson, line 139: The risk on downgrade is that the ownerReference will be pointed to the ClusterBundle resources that have just been deleted, right? 

What if you point the ownerReference to *both* the Bundle and the ClusterBundle objects?

evankanderson (commented):
Since I suggested a bunch of the ideas here, I've added a few thoughts/comments, but feel free to ignore any input that's not helpful -- it's your project, not mine.  😁

design/20241124-rename-bunde-to-clusterbundle.md
erikgb: True! One example might be to make the namespaceSelector required. But any change will add complexity to the migration controller.

evankanderson: I think the complexity will be similar to a storage conversion controller, but it does add some complexity.

erikgb: Thanks a lot for pushing me on this @evankanderson! 👍 It seems to actually represent a huge opportunity for trust-manager. I did a simple draft of proposed API changes in this PR: https://github.com/cert-manager/trust-manager/pull/486. Will use it when discussing the proposal in our next bi-weekly meeting.

design/20241124-rename-bunde-to-clusterbundle.md
erikgb: Thanks! I need to dig deeper into potential upgrade/downgrade risks and eventual mitigations.

design/20241124-rename-bunde-to-clusterbundle.md
```diff
@@ -0,0 +1,158 @@
...
+
+When the functional controller is migrated to target `ClusterBundle`, we want to introduce a temporary technical
+controller targeting the now deprecated `Bundle` resource.
+
+This controller should just create or update the `ClusterBundle` with the same name as the reconciled `Bundle`.
+Since the APIs should have identical `spec`, this should be a simple controller.
+
+To allow a user to migrate their resources from `Bundle` to `ClusterBundle`, the controller should not act
+on `Bundle` delete events. Nor add an owner reference to `ClusterBundle` to avoid the Kubernetes garbage collector
+to act on deletes.
```
SgtCoDFish, line 105: nitpick (wording): (it really doesn't matter though, it's perfectly understandable as is!)

```suggestion
on `Bundle` delete events - nor should it add an owner reference to `ClusterBundle` to avoid the Kubernetes garbage collector
acting on deletes.
```

design/20241124-rename-bunde-to-clusterbundle.md
erikgb: Ok, I believe we can gain something if improving the API when migrating. I've now made adjustments to the proposal.

design/20241124-rename-bunde-to-clusterbundle.md
```diff
@@ -0,0 +1,162 @@
...
+We propose to do this change over multiple releases:
+
+1. Version X: Only `Bundle` API is served and has the functional controller. (present state)
+2. Version X+1: We introduce `ClusterBundle` with API otherwise identical to `Bundle`:
+   - `Bundle` is marked as deprecated.
+   - `ClusterBundle` introduced with agreed API improvements.
+   - Functional controller is migrated to `ClusterBundle`
+   - Validating webhook for both `Bundle` and `ClusterBundle` with same validations
+   - New technical controller added for `Bundle`, converting bundles to cluster bundles
+3. Version Y: Remove `Bundle` and only serve `ClusterBundle` (desired state)
```
SgtCoDFish, line 76: question: Is there any reason we couldn't have the following:

1. Version X: Status quo
2. Version X+1: Introduce `ClusterBundle` resource
    - Totally separate controller to `Bundle`
    - `Bundle` completely unaffected
    - Same `BundleSpec` shared between both resources
    - Warn in logs whenever `Bundle` is created (saying users should be creating `ClusterBundle`s instead)
3. Version Y: Start migrating `Bundle` resources to `ClusterBundle`s
4. Version Z: Remove `Bundle`

We can start the migration at the same time as we add the new CRD, but it seems like this way we can make a bit of progress (i.e. add ClusterBundle) without impacting anything else?

design/20241124-rename-bunde-to-clusterbundle.md
SgtCoDFish: I'd be super interested to see how multiple owner references plays out - that does seem like a way to enable downgrades at least on the face of it? I guess if someone downgraded then the ownerref for the clusterbundle will be invalid (since clusterbundle won't exist any more) but I don't know what problems that would cause in practice?

design/20241124-rename-bunde-to-clusterbundle.md
```diff
@@ -0,0 +1,162 @@
...
+### Test Plan
+
+TODO
+
+### Upgrade / Downgrade Strategy
+
+Downgrading from a version where the migration has started will probably not be possible and should
+be clearly documented and stated in the release notes.
+
+## Alternatives
```
SgtCoDFish, line 151: question: Is it worth listing the alternative of "don't try to migrate anyone"?

Like, if we marked `Bundle` as deprecated but continued to support a controller for it, it wouldn't cause any problems beyond a bit more code to maintain, right?

(I'm not necc. saying that we _should_ do that, but it seems like a scenario we should address in this doc right?)

(I'm assuming that we change the group to `trust-manager.io` for `ClusterBundle`, which would mean that keeping `trust.cert-manager.io Bundle` wouldn't conflict with a future namespace scoped `trust-manager.io Bundle`)

design/20241124-rename-bunde-to-clusterbundle.md
erikgb: Ok, I have tried to improve the upgrade/downgrade strategy according to your feedback. Thanks again!

SgtCoDFish (commented):
I think this is brilliantly written and it does fill me with confidence to see this written down! The link to the chat on k8s slack was enlightening too, thanks for that!

General comment: if we implemented this design as written, the only reason for someone to migrate to `ClusterBundle` is because we (the maintainers) are saying they _have_ to migrate.

There's no new features or anything to draw people to the new name, it's just a thing we changed. That doesn't mean we can't make this change (or that this is a bad change generally), but I think it's better if there's a reason for people to want to migrate, not just a reason they have migrate.

I wrote in one of my review comments about us adding `ClusterBundle` and making no effort to migrate (yet). If we did that - and started only adding new features to `ClusterBundle` - eventually, there'd be a reason that people would want to migrate. I think that's worth considering at least? But I wouldn't block progress over this, and this design is great either way!

What do you think? (sorry that's a lot of words)

design/20241124-rename-bunde-to-clusterbundle.md
erikgb: Since a `Bundle` and `ClusterBundle` could then address the same target resources (by name), we would have to this into account in controllers. In addition, we would have a LOT more code to maintain and test. This is something I don't recommend, but I will add this to the list of alternatives. 👍

design/20241124-rename-bunde-to-clusterbundle.md
erikgb: I have tried to explain above why I **don't** think we should support both resources independently - at the same time. But I will add this option to the alternatives for sure.

Renaming the API group is excellent input! I will add this too.

erikgb:
> There's no new features or anything to draw people to the new name, it's just a thing we changed. That doesn't mean we can't make this change (or that this is a bad change generally), but I think it's better if there's a reason for people to want to migrate, not just a reason they have migrate.

This is addressed by the motivation IMO. Doing this rename will allow us more flexibility in adding new (and already requested) features. In particular a new namespace-scoped `Bundle` resource. We should also improve the API while migrating, which will, at least IMO, reduce complexity significantly - both in code and testing. And simplicity is always a good platform for adding new features.

design/20241124-rename-bunde-to-clusterbundle.md
erikgb: I have added your suggestion as an alternative now.

design/20241124-rename-bunde-to-clusterbundle.md
erikgb: New `trust-manager.io` API group is now added to the list of goals. 👍 

And added a new independent controller as one of the alternatives.

SgtCoDFish (approved):
/lgtm

I don't see any reason we couldn't proceed with this personally, thanks for the effort on this Erik!

erikgb:
> Could you add an example YAML of how the new spec will look?

It makes to add an example of how we could improve the API when migrating to ClusterBundle. At the same time, this design is mainly a proposal on **how** to migrate from `Bundle` to `ClusterBundle`. The API improvements are not mandatory, but just a very attractive optional thing to do at the same time. 😉

design/20241124-rename-bunde-to-clusterbundle.md
```diff
@@ -0,0 +1,252 @@
...
+      - key: bundle.jks
+        format: JKS
+      - key: bundle.p12
+        format: PKCS12
+    namespaceSelector: {} # Selecting all namespaces here, but now required in API
+```
+
+### Risks and Mitigations
+
+#### Target configmaps/secrets are accidentally deleted
```
inteon, line 178: There are two conflicting goals here:
- we do not want to break the behavior when deleting the `Bundle` resource
- we want to support users who remove the `Bundle` resource and start using the `ClusterBundle` instead.

inteon: An alternative approach here would be to add owner references and document how to remove those owner references when you want to start using the `ClusterBundle` as the new "source of truth".

inteon: If "owner references" are too complex, we can also implement our own garbage collector (eg. using labels).

design/20241124-rename-bunde-to-clusterbundle.md
```diff
@@ -0,0 +1,252 @@
...
+### Temporary Technical Controller
+
+When the functional controller is migrated to target `ClusterBundle`, we want to introduce a temporary technical
+controller targeting the now deprecated `Bundle` resource.
+
+This controller should just create or update the `ClusterBundle` with the same name as the reconciled `Bundle`.
+Since the APIs should be quite similar, this should be a simple controller.
+
+- `Bundle` `.spec` should be copied (and eventually slightly converted) into `ClusterBundle` `.spec`
+- `Bundle` `.status` should be updated from `ClusterBundle` `.status`
```
inteon, line 110: I think we should also add a clear message in the `Bundle` `.status` stating that using the `Bundle` is deprecated.
Also, the validation webhook should throw a warning explaining that the `Bundle` resources are deprecated.

design/20241124-rename-bunde-to-clusterbundle.md
erikgb: Why do you consider this as "conflicting goals"? I did not mention "do not want to break the behavior when deleting the Bundle resource" as a goal specifically, did I? 🤔 Considering the alpha state of Bundle API, I don't see a big harm in potentially orphaning some resources that can easily be deleted by the user as a workaround. The alternative could be a lot more destructive.

As a user, I **don't** want my `ClusterBundle` to be deleted when the corresponding `Bundle` is deleted. Maybe I am missing your point. Please clarify!

design/20241124-rename-bunde-to-clusterbundle.md
erikgb: CRDs supports deprecating APIs/versions, and we can tailor the warning message. Isn't this enough?

ThatsMrTalbot: Im pretty sure the CRD deprecation warning tells you every time you run kubectl against it, we can test this though to make sure thats true with all the commands (describe, get, apply). If that is the case then I think a CRD deprecation warning is enough

design/20241124-rename-bunde-to-clusterbundle.md
```diff
@@ -0,0 +1,252 @@
...
+apiVersion: trust-manager.io/v1alpha1
+kind: ClusterBundle
+metadata:
+  name: my-bundle
+spec:
+  sources:
+    - configMap:
+        name: my-ca
+        key: ca.crt
+  target:
```
ThatsMrTalbot, line 166: I've often thought that instead of having `spec.target.configMap` and `spec.target.secret` we should just have `spec.targets` which is an array of object references including apiVersion and kind. 

This makes supporting future sources much easier, say for example if we wanted to add a plugin system like cert-manager has.

erikgb: Interesting! Could you draw an example? 😊

ThatsMrTalbot: 🤦 I was thinking of sources not targets, for example:

```yaml
apiVersion: trust-manager.io/v1alpha1
kind: ClusterBundle
metadata:
  name: my-bundle
spec:
  sources:
  # Using apiVersion/kind means we can utilise this same array for 
  # any future "extensions"/"plugins".
  #
  # Loading from other sources could be a valuable feature and this API change
  # would enable that option in the future.
  - apiVersion: v1
    kind: ConfigMap
    name: my-ca
    key: ca.crt
  ...
```

erikgb: Good input, thanks! I suggest moving the detailed API changes discussion to https://github.com/cert-manager/trust-manager/issues/242. I have also opened https://github.com/cert-manager/trust-manager/pull/486, which will be modified to change the initially unserved `ClusterBundle` - after it's introduced.

For this design, I suggest we agree that we **should** consider changing the API. And not exactly what we want to change. Are you ok with this approach?

design/20241124-rename-bunde-to-clusterbundle.md
```diff
@@ -0,0 +1,252 @@
...
+
+By pushing users to migrate smoothly to a new API, we can perform requested API changes at the same time.
+Consider it similar to a one-way conversion webhook.
+
+Since trust-manager is maturing and getting more attention, we suggest establishing its own API group: `trust-manager.io`.
+
+### Goals
+
+- `Bundle` resource is renamed to `ClusterBundle`
+- `ClusterBundle` is created in the new `trust-manager.io` API group
```
hawksight, line 58: to trust-manager.io from cert-manager.io?

erikgb: Yes, from `trust.cert-manager.io` to `trust-manager.io` - as a grown-up. 😉

SgtCoDFish (approved):
/lgtm
/approve

Erik raised this a while back and I don't see any blockers to merging this. Especially for a design document, I think we should be agile here and unblock progress on implementation. We can always go back and improve the design doc as we learn more during implementation!

Issue #243, opened 2023-11-23
More flexible and better organized target specification in API


erikgb:
I will start with a "max" target specification using the existing API (also including #235 - which I think will be merged before this issue is eventually addressed):

````yaml
target:
    configMap:
      key: root-certs.pem
    secret:
      key: root-certs.pem
    additionalFormats:
      jks:
        key: bundle.jks
        password: jks-password
      pkcs12:
        key: bundle.p12
        password: pkcs12-password
````

I can identify several issues with this specification structure:

- It forces me to include a PEM encoded in target configmap/secret. There is no way to opt-out and just use JKS/PKCS12.
- The term `additional` appears a bit "discriminating", and does not motivate users to suggest (or contribute) new formats (e.g. DER).
- The PEM encoded bundle allows different `key` in configmap and secret, while for "additional formats" this is not supported.
- Encodings that require a password do not support specifying a different password for configmap and secret.
- The format does not allow me to output the same encoding in two different keys of target. I am not sure if we should support this, but it could theoretically make sense in a migration scenario. But the simple workaround is just to create a new bundle for migration.

To solve most of these issues, I suggest a more flexible (and less "discriminating") structure. It might appear more complex, but remember that most users will probably not use a "max" specification, and only specify relevant fields for their setup.

````yaml
target:
    configMap:
      pem:
        key: root-certs.pem
      jks:
        key: bundle.jks
        password: jks-password
      pkcs12:
        key: bundle.p12
        password: pkcs12-password
    secret:
      pem:
        key: root-certs.pem
      jks:
        key: bundle.jks
        password: jks-password
      pkcs12:
        key: bundle.p12
        password: pkcs12-password
````

**Comment by hazmat345** [comment]
Maybe this could also be a chance to bring the target side more in line with the sources side? What would you think about something more like this instead:

```
targets:
  - configMap:
      format: pem
      key: root-certs.pem
  - configMap:
      format: jks
      key: bundle.jks
      password: jks-password
  - configMap:
      format: pkcs12
      key: bundle.p12
      password: pkcs12-password
  - secret:
      format: pem
      key: root-certs.pem
  - secret:
      format: jks
      key: bundle.jks
      password: jks-password
  - secret:
      format: pkcs12
      key: bundle.p12
      password: pkcs12-password
```

erikgb:
[...]

Issue #39, opened 2022-07-21
Don't sync targets to all namespaces by default


SgtCoDFish:
PR #37 introduced the ability for users to specify a subset of namespaces to which a target should be synced. We propose that the `namespaceSelector` becomes a required field, replacing the current behaviour of a target being synced everywhere by default.

This would be a breaking change. We agreed in the 2022-07-21 standup that breaking changes are something we're willing to accept for `trust` given it's pre-v1.0.

## Justification

Imagine a multi tenanted cluster.

In namespace `foo`, FooCorp has their root certificate with subject `CN=FooCorp`. This is added to `FooCorp`'s bundle.

In namespace `bar`, BarCorp is a different company and does totally different things.

By default, FooCorp's bundle will by synced to BarCorp's namespace. BarCorp employees will see a ConfigMap with FooCorp's root certificate in it. They won't be able to _use_ the root, but they will be able to see that they're tenanted on the same cluster as FooCorp.

Likewise, if someone at BarCorp erroneously used the FooCorp bundle it would enable anyone with access to the FooCorp root to MitM any affected BarCorp service.

Syncing to all namespaces by default introduces an _absolutely critical security requirement_ for some users; they _must_ ensure they always set a namespaceSelector. Effectively they need to manage a denylist from a default position of "allow all".

Making namespaceSelector required would change this to an allowlist. It's a little more work for the average user but it will present a significant security win.

erikgb:
@SgtCoDFish I think it could make sense to add this to the v1 milestone.

erikgb:
> Yeah I guess that'd make sense! This is something I assume we'd change in v1alpha2 or some other new API version. I'll add to the milestone!

I have a couple of other suggestions for a new API version, so if you are ready for it, I can start on the job. I just finished introducing a new API version in https://github.com/cybozu-go/accurate, so my mind is full of experiences.

hawksight:
The use case certainly swayed me here, but I would like to think about the usability element. Thinking of cert-manager `Certificate` spec, there are "default" values that often people want to change but cannot (easily). It might well be easier in the long run to have a configurable default that allows for easier roll out in many use cases.

So in the example above, what if you are not a multi tenant cluster? 
What if you just want the current default behaviour to sync everywhere?
Currently that would involve:

```yaml
spec:
  target:
    namespaceSelector:
      matchLabels:
        key: value
```

Plus labelling all the applicable namespaces. (Which could be many) `k label ns test key=value`.

[...]

PR #486, opened 2024-11-24
WIP: Proposed API changes for ClusterBundle migration

erikgb:
After getting feedback on the migration proposal, I believe the proposed (one-way) migration from Bundle to ClusterBundle could represent an opportunity to improve the API of trust-manager. So I did a little experiment and drafted some of the changes I think make sense. Relates to https://github.com/cert-manager/trust-manager/pull/485.

The proposed changes to the API will address https://github.com/cert-manager/trust-manager/issues/39 (simple).

Changes will also make the target(s) specification a lot more consistent IMO, ref. https://github.com/cert-manager/trust-manager/issues/243. The proposed spec will remove the need for complex and error-prone webhook validations, make the tests simpler, and also provide a new feature: the ability to get multiple keys of the same trust-bundle format in a target.

Since the proposed migration represents a one-way "conversion", I believe the conversion logic will be quite simple.


erikgb:
This PR should be ready for review now, but as it contains a massive change to the API, I will put a hold on it to avoid a premature merge.

/hold

hawksight:
@erikgb I pulled down your branch and ran through tests. I was poking around the cluster but everything appears to be the same. Am I right to assume that this PR does not expose a `ClusterBundle` resource? I still only see the `bundles` CRD / api-resource.

It might be worth adding a note on the intended outcome of this PR from a user perspective?

pkg/apis/trustmanager/v1alpha2/types_cluster_bundle.go
```diff
@@ -212,13 +192,47 @@ type SourceObjectKeySelector struct {
...
+
+	// JKS specifies configs for target JKS files.
+	// May only be used when format is JKS.
+	// +optional
+	JKS *JKS `json:"jks,omitempty"`
+
+	// PKCS12 specifies configs for PKCS#12 files.
+	// May only be used when format is PKCS12.
+	// +optional
+	PKCS12 *PKCS12 `json:"pkcs12,omitempty"`
```
SgtCoDFish, line 221: comment: Does this imply that if I want a single ConfigMap target with JKS, PKCS#12 and PEM then I need three entries? That seems a little cumbersome to write, no?

erikgb:
> @erikgb I pulled down your branch and ran through tests. I was poking around the cluster but everything appears to be the same. Am I right to assume that this PR does not expose a `ClusterBundle` resource? I still only see the `bundles` CRD / api-resource.
> 
> It might be worth adding a note on the intended outcome of this PR from a user perspective?

@hawksight It is intentionally a no-op for users, as controller-gen is instructed to skip this new API group/version. Forming the new `ClusterBundle` API will probably take some time, and we don't want to block releases of trust-manager while still working trunk-based.

pkg/apis/trust/v1alpha1/conversion.go
```diff
@@ -34,6 +38,71 @@ func (src *Bundle) ConvertTo(dstRaw conversion.Hub) error {
...
+func Convert_v1alpha1_JKS_To_v1alpha2_JKS(in *JKS, out *trustv1alpha2.JKS, _ apimachineryconversion.Scope) error {
+	out.Password = in.Password
+	return nil
+}
+
+func Convert_v1alpha1_PKCS12_To_v1alpha2_PKCS12(in *PKCS12, out *trustv1alpha2.PKCS12, _ apimachineryconversion.Scope) error {
+	out.Password = in.Password
+	out.Profile = trustv1alpha2.PKCS12Profile(in.Profile)
+	if out.Profile == "" {
+		// Default profile changed from LegacyRC2 to LegacyDES in v1alpha2
```
SgtCoDFish, line 100: suggestion: It would be really cool to create an ongoing list of changes we're making so we can more easily document it later. It's great that we have this comment here - maybe we could alos add all these kinds of comments to, say, `docs/alphav2-changes.md` and keep a list there?

SgtCoDFish (commented):
Took an initial look! Thanks for driving this!

pkg/apis/trust/v1alpha1/conversion.go
erikgb: Yes, excellent suggestion. I will add this right away, but keep it simplistic for now, as I don't want to update it for each and every review suggestion. 😅 But listing the changes in behavior is excellent!

pkg/apis/trustmanager/v1alpha2/types_cluster_bundle.go
erikgb: A simple configmap with your specification could look like this:

````yaml
target:
  configMap:
    - key: root-certs.pem
    - key: bundle.jks
      format: jks
    - key: bundle.p12
      format: pkcs12
````

That isn't a lot more than today, or?

File: pkg/apis/trustmanager/v1alpha2/types_cluster_bundle.go [hunk 3]
```diff
@@ -212,13 +192,47 @@ type SourceObjectKeySelector struct {
...
-type KeySelector struct {
-	// Key is the key of the entry in the object's `data` field to be used.
+// KeyValueTarget is the specification of key value target resources as ConfigMaps and Secrets.
+// +listType=map
+// +listMapKey=key
+// +kubebuilder:validation:MinItems=1
+type KeyValueTarget []TargetKeyValue
+
+// TargetKeyValue is the specification of a key with value in a key-value target resource.
+// +structType=atomic
```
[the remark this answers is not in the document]
SgtCoDFish: Sorry, my comment was really unclear. My fault! I don't think there's anything wrong here; I was talking about something else but on second thoughts I don't think it matters. Ignore this one!

pkg/apis/trustmanager/v1alpha2/types_cluster_bundle.go
SgtCoDFish: My bad, I think I misread this a bit. Yeah that looks good!

pkg/apis/trust/v1alpha1/conversion.go
erikgb: I have added  `docs/alphav2-changes.md`. It appears useful and I hope it addresses your suggestion. PTAL!

erikgb:
If the deprecation of explicit JKS support is accepted, ref. https://github.com/cert-manager/trust-manager/pull/603, the API proposed in this PR can be simplified.

docs/alphav2-changes.md
```diff
@@ -0,0 +1,28 @@
+# Version v1alpha2 API changes
+
+In v1alpha2, `ClusterBundle` was introduced as a replacement for v1alpha1 `Bundle`.
+This was mainly done to make room for a future namespace-scoped `Bundle`.
```
wallrj, line 4: ### Why do we want a namespace-scoped `Bundle`?

The [ClusterBundle design](https://github.com/cert-manager/trust-manager/blob/main/design/20241124-rename-bunde-to-clusterbundle.md#motivation) says:
> it turns out that the current name somehow blocks introducing an equivalent namespace-scoped resource - for trust bundle management inside namespaces in multi-tenant clusters.

Why wouldn't the application-team just put the CA certificates in a `ConfigMap` and mount that into their Pod, alongside the ConfigMap containing up-to-date public CAs, that has been published to their tenant namespace by the platform-team, using trust-manager?

### Why chose the name `ClusterBundle`?

The [ClusterBundle design](https://github.com/cert-manager/trust-manager/blob/main/design/20241124-rename-bunde-to-clusterbundle.md#motivation) says:
> In addition, ClusterBundle would be a better name looking at cert-manager having Issuer and ClusterIssuer. The name would also be better aligned with upstream [ClusterTrustBundle](https://kubernetes.io/docs/reference/access-authn-authz/certificate-signing-requests/#cluster-trust-bundles).

I agree that the name `ClusterBundle` is more consistent with the name `ClusterIssuer` in cert-manager.

I do not agree that the name is better aligned with ClusterTrustBundle....I expect it will cause confusion.
`ClusterTrustBundle` seems to be a cluster-scoped container for CA certificates,  which can be mounted into Pods.
`ClusterBundle` (AFAIU) is a configuration for trust-manager, telling it where to pull CA certificates from, and where to push them to. Ultimately: ConfigMap, Secret, ClusterTrustBundle.
So we will be writing documentation asking users to:
> Create a trust-manager `Bundle` or `ClusterBundle` to copy CA certificates into a Kubernetes `ClusterTrustBundle` and then mount that Kubernetes  `ClusterTrustBundle` into your Pod.

Do you agree that this is confusing? 
If we're going to change the name, shouldn't we choose something completely different, like `TrustManagerClusterMapping` and `TrustManagerMapping`, to differentiate it from the Kubernetes resource.

### Why did Kubernetes choose a cluster-scoped `Bundle`?

I am confused about why Kubernetes chose cluster-scoped ClusterTrustBundle and why they seem to explicitly avoid introducing a namespaced resource. In my mind, this conflicts your aim of making it easier to manage CA bundles in a multi-tenant cluster.
Do you know if they ever talked about how ClusterTrustBundle would work for multiple application teams who share a cluster?

For example: https://github.com/kubernetes/enhancements/tree/master/keps/sig-auth/3257-cluster-trust-bundles#access-control
> All ClusterTrustBundle objects are assumed to be publicly-readable within the cluster, due to the fact that the kubelet trustAnchors volume type will allow any pod to mount any ClusterTrustBundle. This will be backed up by a built-in grant of get, list, and watch for ClusterTrustBundle objects to the system:serviceaccounts and system:nodes groups.
> For ClusterTrustBundle objects without .spec.signerName specified, no nonstandard access control is applied. Due to the clusterTrustBundle projected volume, any workload in the cluster will be able to load the contents of any ClusterTrustBundle.

Another example: in https://github.com/kubernetes/enhancements/tree/master/keps/sig-auth/3257-cluster-trust-bundles#use-a-configmap-rather-than-defining-a-new-type which says:
> Use a ConfigMap rather than defining a new type
ConfigMaps have two problems in this application:
> They are namespace-scoped, whereas signers are cluster-scoped.
> They have poor human factors. They are not often considered to contain security-critical data, and so may have wide update permissions assigned to them.

wallrj (commented):
@erikgb You asked for more feedback during the cert-manager standup on 13 May 2025, so here are some of my questions and thoughts.

SgtCoDFish (approved):
/lgtm
/approve
/hold

I'm a little out of it today, but I can't really see a reason that this shouldn't be merged so I'm adding a quick approval in case that unblocks Erik / allows us to iterate on this. Thank you for this!

SgtCoDFish:
This was discussed in standup today, and it was agreed we should unhold so this can move forwards. Thanks for your patience Erik!

/unhold

PR #493, opened 2024-11-30
Simplify integration test setup

erikgb:
After https://github.com/cert-manager/makefile-modules/pull/214 is available in this project, we can simplify the integration test setup to use the generated CRD manifests.


inteon:
see https://github.com/cert-manager/approver-policy/pull/551 for a similar PR.

PR #495, opened 2024-12-01
Introduce ClusterBundle API as a copy of Bundle

erikgb:
Initial PR to implement https://github.com/cert-manager/trust-manager/pull/485 and eventually address https://github.com/cert-manager/trust-manager/issues/63.

In this PR, I propose to introduce `ClusterBundle` as a "copy" of the current `Bundle` API, but I have configured controller-gen to skip this new API for now. That's why there is no new CRD yet. This will also allow us to release trust-manger while we are working on this.

Some important details worth mentioning:

- `ClusterBundle` is the first member of a new API group: `trust-manager.io`  🥳 
- The initial `ClusterBundle` API version is `v1alpha2`, to make it clear that it's the way forward from the current Bundle in `v1alpha1`. This is done intentionally, even if `v1alpha1` could have been the initial version in the new API group.
- I have added `conversion-gen` to the CI to ensure we can convert from `Bundle` to `ClusterBundle` and vice versa. This should catch future changes to `Bundle` and/or `ClusterBundle`. To test that the conversion is round-trippable, I have added a fuzzer test. Thanks to the maintainers of https://github.com/kubernetes-sigs/cluster-api for showing how it could be done. 🏅 

The next natural step would be to update https://github.com/cert-manager/trust-manager/pull/486 to address the new `ClusterBundle` API instead. 

/hold


pkg/api/v1alpha2/types_cluster_bundle.go
```diff
@@ -0,0 +1,294 @@
...
+}
+
+// BundleSource is the set of sources whose data will be appended and synced to
+// the BundleTarget in all Namespaces.
+// +structType=atomic
+type BundleSource struct {
+	// ConfigMap is a reference (by name) to a ConfigMap's `data` key(s), or to a
+	// list of ConfigMap's `data` key(s) using label selector, in the trust Namespace.
+	// +optional
+	ConfigMap *SourceObjectKeySelector `json:"configMap,omitempty"`
```
ThatsMrTalbot, line 78: As opposed to having dedicated fields for ConfigMap/Secret could we instead have fields for Group and Kind? That seems like a very common pattern in modern K8s APIs and means we can add new support without API changes.

The Group could default to "" for convenience so users just specify Kind:

```yaml
sources:
- kind: Secret
  name: example
  key: example
- kind: ConfigMap
  selector:
    matchLabels:
      foo: bar
  key: example
```

I think this looks cleaner and is more in line with current patterns, happy to hear other opinions though.

ThatsMrTalbot: Also means we could use the same pattern to add support for the new `ClusterTrustBundles` K8s now has

erikgb: Thanks for reviewing @ThatsMrTalbot! I would be happy to discuss the new API, but the goal of this PR is just to introduce `ClusterBundle` as an initial copy of `Bundle`. We could then move over to adjusting the API. My gut feeling tells me that doing everything in the same PR would delay the process a lot more. But open for other opinions here!

For your concrete API suggestions, maybe you could add them to https://github.com/cert-manager/trust-manager/issues/242?

ThatsMrTalbot: Makes sense, added a comment to the other issue

SgtCoDFish (approved):
/lgtm
/approve

As discussed in standup, this is a no-op and is a base for collaboration. I think this can be merged.

Ideally, this would have a marker to tell people not to import it, but it's OK as-is!

Issue #183, opened 2023-09-20
Create trust bundle based on Debian bookworm


SgtCoDFish:
We currently provide [cert-manager-package-debian](https://quay.io/repository/jetstack/cert-manager-package-debian) which contains the trust store from Debian bullseye. That trust store hasn't been updated for a while - since it hasn't needed to be - but it won't be updated to add new CA certificates or to prune CA certificates which have expired.

Since there's a new stable version of Debian - Debian bookworm - we could provide a new trust bundle which uses the ca-certificates package from that new version. That would be more "up to date" for users who want that.

We can create a new bundle for bookworm and update to use it by default in trust-manager. This would involve copying the "debian" bundle [here](https://github.com/cert-manager/trust-manager/tree/main/trust-packages) into a new folder ("debian-bookworm") and then updating the new bundle to use bookworm rather than bullseye.

There's some GCB CI/CD for our trust bundles, which is manually configured in the cert-manager GCP org, but I can set that up. The GCB file [here](https://github.com/cert-manager/trust-manager/blob/main/gcb/ci-update-debian-trust-package.yaml) would need to be copied and the new version would need to be updated to bookworm.

ℹ️ This would be a great issue for an open source community member to tackle!

arsenalzp:
Hello,
I can pick it up, but I would like to know - is there any terms for this issue?

SgtCoDFish:
>  I can pick it up, but I would like to know - is there any terms for this issue?

Sorry, I didn't see this until now!

Could you explain what you mean by "terms" there?

arsenalzp:
> > I can pick it up, but I would like to know - is there any terms for this issue?
> 
> Sorry, I didn't see this until now!
> 
> Could you explain what you mean by "terms" there?

I meant "deadlines".

SgtCoDFish:
Ah, no deadlines! We can be flexible, no rush 😁

arsenalzp:
Could you please assign it to me?

arsenalzp:
Hello,
I'm back and ready to work.
Could you be so kind to explain me why we need to copy old trust package stuff (in `debian` folder) instead just change base image directive to use the latest Debian docker package?

SgtCoDFish:
>  Could you be so kind to explain me why we need to copy old trust package stuff (in debian folder) instead just change base image directive to use the latest Debian docker package?

Sure, I'll try to explain my thinking!

[...]

PR #511, opened 2024-12-13
Bump to Go 1.23 (and allow certificates with negative serial numbers)

erikgb:
Preparations for https://github.com/cert-manager/trust-manager/pull/510 as K8s 1.32 dependencies require Go 1.23.

It seems like Go 1.23 rejects negative serial numbers by default in [x509.ParseCertificate](https://pkg.go.dev/crypto/x509#ParseCertificate), but the old behavior can be restored - as least as a short-term fix.

Long-term fix is probably to ensure our default trust bundles don't contain offending certificates. Maybe https://github.com/cert-manager/trust-manager/issues/183?

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,10 @@
 module github.com/cert-manager/trust-manager
 
-go 1.22.0
+go 1.23.0
+
+// Added on bump to Go 1.23 as it seems like our current trust bundle contains certs with negative serial numbers.
+// See also https://pkg.go.dev/crypto/x509#ParseCertificate
+godebug x509negativeserial=1
 
 require (
 	github.com/go-logr/logr v1.4.2


inteon:
@erikgb do we know which certificate has a negative serial number?

erikgb:
> @erikgb do we know which certificate has a negative serial number? Let's investigate that first.

@inteon See https://kubernetes.slack.com/archives/CDEQJ0Q8M/p1734108493381519?thread_ts=1734105432.142589&cid=CDEQJ0Q8M

erikgb:
We have decided to take another approach to fix this issue. See https://github.com/cert-manager/trust-manager/pull/515

inteon:
This is a good first fix, we have a long-term plan to remove this godebug option in the future (see https://github.com/cert-manager/trust-manager/pull/515).

PR #512, opened 2024-12-13
add metric tracking negative serial numbers on trust bundle certs

SgtCoDFish:
Related to #511  and #510 

This lets us track where we _would_ have failed if we hadn't set the godebug


SgtCoDFish:
I don't think we'll actually need this any time soon, this was more of a PoC. Closing!

PR #514, opened 2024-12-16
Pull preloaded images based on digest, not tag

inteon:
WIP: see https://github.com/cert-manager/makefile-modules/pull/227


inteon:
See https://github.com/cert-manager/makefile-modules/pull/227 instead.

PR #516, opened 2024-12-17
Manual makefile modules update

SgtCoDFish:
This should fix issues with the smoke test, to re-enable testing, fix the auto upgrade process and unblock other upgrades.


SgtCoDFish:
This has hit a different makefile-modules issue: https://github.com/cert-manager/makefile-modules/pull/228#issuecomment-2549188602

SgtCoDFish:
This will be unblocked by https://github.com/cert-manager/makefile-modules/pull/229 and then this PR will need to be updated

PR #540, opened 2025-01-30
Add build for trust package based on Debian bookworm

SgtCoDFish:
Also switches to use the new trust package by default in tests, but not yet for trust-manager properly (since there's no image built yet - that'll come in a follow-up PR).

This is a brute force approach of pretty much copying everything from the existing trust package.

### DIFF
--- a/make/00_mod.mk
+++ b/make/00_mod.mk
@@ -15,13 +15,14 @@
 oci_platforms := linux/amd64,linux/arm/v7,linux/arm64,linux/ppc64le,linux/s390x
 
 include make/00_debian_version.mk
+include make/00_debian_bookworm_version.mk
 
 repo_name := github.com/cert-manager/trust-manager
 
 kind_cluster_name := trust-manager
 kind_cluster_config := $(bin_dir)/scratch/kind_cluster.yaml
 
-build_names := manager package_debian
+build_names := manager package_debian package_debian_bookworm
 
 go_manager_main_dir := ./cmd/trust-manager
 go_manager_mod_dir := .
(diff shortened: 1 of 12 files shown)


erikgb:
Thanks, Ashley! I believe merging this PR would close https://github.com/cert-manager/trust-manager/issues/183. I will try to take a look at the proposed changes, but that will not happen tonight. 😪

erikgb (commented):
I must admit that I am not that familiar with Makefiles, but the proposed changes looks like it would do what's specified in the PR description.

One overall suggestion: I find the Debian "codenames" a bit confusing, and they use version numbers in their images. What if we avoid the codenames bullseye and bookworm and instead use the version numbers: 11 and 12?

/approve

make/00_mod.mk
```diff
@@ -55,7 +67,7 @@ define helm_values_mutation_function
 $(YQ) \
 	'( .image.repository = "$(oci_manager_image_name)" ) | \
 	( .image.tag = "$(oci_manager_image_tag)" ) | \
-	( .defaultPackageImage.repository = "$(oci_package_debian_image_name)" ) | \
-	( .defaultPackageImage.tag = "$(oci_package_debian_image_tag)" )' \
+	( .defaultPackageImage.repository = "$(oci_package_debian_bookworm_image_name)" ) | \
+	( .defaultPackageImage.tag = "$(oci_package_debian_bookworm_image_tag)" )' \
```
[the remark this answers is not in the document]
SgtCoDFish: This doesn't _need_ to change now. We can choose!

This changes the default image for testing only, so it's safe because we build the image for the tests. I'd be fine keeping the old image here and changing this + the helm templates all in one, but I think it's harmless to have this here since I think it makes sense to switch to bookworm by default for the next trust-manager release anyway, so as soon as this merges + the image is build I'd raise a second PR for switching the helm templates!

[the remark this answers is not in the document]
erikgb: Ah, it's just changing the image for testing! Sorry! Then we should do it now.

make/00_mod.mk
```diff
@@ -33,14 +34,25 @@ oci_manager_image_name_development := cert-manager.local/trust-manager
...
 oci_package_debian_image_tag := $(DEBIAN_BUNDLE_VERSION)
 oci_package_debian_image_name_development := cert-manager.local/cert-manager-package-debian
 debian_package_layer := $(bin_dir)/scratch/debian-trust-package
 oci_package_debian_additional_layers += $(debian_package_layer)
 
+go_package_debian_bookworm_main_dir := .
+go_package_debian_bookworm_mod_dir := ./trust-packages/debian
+go_package_debian_bookworm_ldflags :=
+oci_package_debian_bookworm_base_image_flavor := static
+oci_package_debian_bookworm_image_name := quay.io/jetstack/cert-manager-package-debian-bookworm
```
SgtCoDFish, line 49: @erikgb what do you think about this package name? The current one is `quay.io/jetstack/cert-manager-package-debian`.

I worry when we migrate to a cert-manager branded repo it'll be repetitive:

`quay.io/cert-manager/cert-manager-package-debian-bookworm`

Would `quay.io/jetstack/trust-package-debian-bookworm` work better?

SgtCoDFish: `quay.io/jetstack/trust-pkg-debian-bookworm`, to be shorter? 🤔

erikgb: Yes, I think this is a good rename. Also taken trust-manager maturing into consideration.

make/debian-trust-package-bookworm.mk
```diff
@@ -0,0 +1,45 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+package_name_bookworm := cert-manager-debian-bookworm
+
+debian_package_bookworm_layer := $(bin_dir)/scratch/debian-trust-package-bookworm
+debian_package_bookworm_json := $(debian_package_bookworm_layer)/debian-package/cert-manager-package-debian.json
```
erikgb, line 18: Is this correct? Looks like it might be a copy-paste that hasn't been modified.

SgtCoDFish:
Followed up by #543 and #544

PR #543, opened 2025-02-04
Add fetch-depth to checkout action

SgtCoDFish:
This should hopefully fix the error in https://github.com/cert-manager/trust-manager/actions/runs/13136435669/job/36652783894

### DIFF
--- a/.github/workflows/debian-trust-package-release-bookworm.yaml
+++ b/.github/workflows/debian-trust-package-release-bookworm.yaml
@@ -16,6 +16,10 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
+      # Adding `fetch-tags: true` is required since the makefiles use tags for versioning
+      # See https://github.com/actions/checkout/issues/701 for extra info
+      with:
+        fetch-tags: true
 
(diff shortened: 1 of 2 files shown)


.github/workflows/debian-trust-package-release-bookworm.yaml
```diff
@@ -16,6 +16,10 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
+      # Adding `fetch-depth: 0` makes sure tags are also fetched. We need
+      # the tags so `git describe` returns a valid version.
+      # see https://github.com/actions/checkout/issues/701 for extra info about this option
+      with: { fetch-depth: 0 }
```
erikgb, line 22:
```suggestion
      with:
        fetch-tags: true
```

SgtCoDFish: I was blindly copying the other job but this definitely seems worth testing, thanks :D

I guess the comment is out of date, I'll update that and repush

erikgb (commented):
/lgtm
/approve

PR #550, opened 2025-02-07
Use Context to pass Logger

erikgb:
As noted in https://github.com/cert-manager/trust-manager/pull/442#discussion_r1946705109, it seems like we prefer to use the Context to pass around the logger. I've also been annoyed by this when attempting to refactor code. 🦄 

This PR removes all use of Logger as fields in structs - except for the CertPool, which will probably be refactored out of this repository to a more reusable module.

This change will most certainly change the looks of trust-manager logs. Instead of creating a many sub-loggers, we now rely on controller-runtime to initialize the logger supplied with Context. To my experience it does a pretty good job enriching the structured logger with relevant key/value pairs.

### DIFF
--- a/pkg/bundle/bundle.go
+++ b/pkg/bundle/bundle.go
@@ -22,7 +22,6 @@ import (
 	"fmt"
 	"strings"
 
-	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -34,6 +33,7 @@ import (
 	"k8s.io/utils/clock"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
 
 	trustapi "github.com/cert-manager/trust-manager/pkg/apis/trust/v1alpha1"
 	"github.com/cert-manager/trust-manager/pkg/bundle/internal/ssa_client"
@@ -43,9 +43,6 @@ import (
 
 // Options hold options for the Bundle controller.
 type Options struct {
-	// Log is the Bundle controller logger.
-	Log logr.Logger
-
 	// Namespace is the trust Namespace that source data can be referenced.
 	Namespace string
 
@@ -106,7 +103,8 @@ func (b *bundle) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, 
 }
 
 func (b *bundle) reconcileBundle(ctx context.Context, req ctrl.Request) (result ctrl.Result, statusPatch *trustapi.BundleStatus, returnedErr error) {
-	log := b.Log.WithValues("bundle", req.NamespacedName.Name)
+	log := logf.FromContext(ctx).WithValues("bundle", req.NamespacedName.Name)
+	ctx = logf.IntoContext(ctx, log)
 	log.V(2).Info("syncing bundle")
 
 	var bundle trustapi.Bundle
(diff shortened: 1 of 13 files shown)


pkg/webhook/webhook.go
```diff
@@ -26,15 +26,10 @@ import (
...
-	Log logr.Logger
-}
-
 // Register the webhook endpoints against the Manager.
-func Register(mgr manager.Manager, opts Options) error {
-	opts.Log.Info("registering webhook endpoints")
-	validator := &validator{log: opts.Log.WithName("validation")}
+func Register(mgr manager.Manager, log logr.Logger) error {
+	log.Info("registering webhook endpoints")
+	validator := &validator{}
```
SgtCoDFish, line 32: suggestion: I don't think we need to pass the logger here for this message, it doesn't seem worth it to me. I'd vote to move `log.Info("registering webhook endpoints")` to where the `Register` function is called and drop the logger here entirely

pkg/bundle/internal/target/target.go
```diff
@@ -151,7 +149,7 @@ func (r *Reconciler) syncConfigMap(
 		return false, fmt.Errorf("failed to patch %s %s: %w", target.Kind, target.NamespacedName, err)
 	}
 
-	log.V(2).Info(fmt.Sprintf("synced bundle to namespace for target %s", target.Kind))
+	logf.FromContext(ctx).V(2).Info("synced bundle to namespace")
```
SgtCoDFish, line 152: note: We're not setting the target type in `pkg/bundle/bundle.go`

suggestion: is it worth adding the namespace name to this log line (and to  the equivalent for `syncSecret`)? Otherwise this might be quite a spammy log line with little context.

pkg/bundle/bundle.go
```diff
@@ -106,7 +103,7 @@ func (b *bundle) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result,
 }
 
 func (b *bundle) reconcileBundle(ctx context.Context, req ctrl.Request) (result ctrl.Result, statusPatch *trustapi.BundleStatus, returnedErr error) {
-	log := b.Log.WithValues("bundle", req.NamespacedName.Name)
+	log := logf.FromContext(ctx)
```
SgtCoDFish, line 106: suggestion: Previously we were setting the bundle name here (`b.Log.WithValues("bundle", req.NamespacedName.Name)`). Is it intentional to remove that? I think it's useful context

SgtCoDFish (changes requested):
Got a few suggestions - I think this will be a big improvement and I'm looking forward to it!

pkg/webhook/webhook.go
erikgb: Agreed! Will fix.

pkg/bundle/bundle.go
erikgb: Yeah, it was probably a mistake to remove it. My idea was that it is covered by the context set here:

https://github.com/erikgb/trust-manager/blob/8779ee435fe0303e265abadb9417ed520ee8cd72/pkg/bundle/bundle.go#L282-L283

But the following log-line looses context on this.

pkg/bundle/internal/target/target.go
erikgb: The log-context should be covered by:

https://github.com/erikgb/trust-manager/blob/8779ee435fe0303e265abadb9417ed520ee8cd72/pkg/bundle/bundle.go#L282-L283

pkg/webhook/webhook.go
erikgb: Fixed now.

pkg/bundle/bundle.go
erikgb: Retained the old behavior now.

SgtCoDFish (approved):
/lgtm
/approve

Thanks for this, really great improvement!

PR #573, opened 2025-03-24
Fix Ginkgo commands

erikgb:
... failing with newer versions of Ginkgo, ref. CI failing in https://github.com/cert-manager/trust-manager/pull/572.

Ginkgo CLI is now validating commands more strictly than it used to, and I think our current commands broke with https://github.com/onsi/ginkgo/releases/tag/v2.23.1.


SgtCoDFish (approved):
/lgtm
/approve

Thank you!

PR #578, opened 2025-04-02
feat: add toplevel enable value to helmchart

### LINKED ISSUE #574: Add a top level variable called `enabled` to the Helm chart schema
cert-manager already does this, and it's quite useful when you're using it as a subchart because you don't need to create another variable in your own `values.yaml`. trust-manager has defined in its schema that it doesn't accept variables not defined in the schema, so it needs to be added in the original Helm chart.

https://github.com/cert-manager/cert-manager/blob/9aa47a01ded1a47e772909f03a816f1a73680769/deploy/charts/cert-manager/values.yaml#L1458-L1465

FloSchick:
Fixes #574. Adds Field that can be used as a condition when cert-manager is a dependency.

### DIFF
--- a/deploy/charts/trust-manager/values.linter.exceptions
+++ b/deploy/charts/trust-manager/values.linter.exceptions
@@ -1,0 +1,1 @@
+value missing from templates: enabled
--- a/deploy/charts/trust-manager/values.schema.json
+++ b/deploy/charts/trust-manager/values.schema.json
@@ -592,6 +592,11 @@
       "description": "Override the image tag of the default package image. If no value is set, the chart's appVersion is used.",
       "type": "string"
     },
+    "helm-values.enabled": {
+      "default": true,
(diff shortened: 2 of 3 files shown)


SBejga:
Working together with @FloSchick at Contribfest cert-manager on KubeCon 2025 :-)

erikgb:
Thanks! Please run `make generate-helm-schema` and commit the result to fix CI!

hawksight:
So I think that you need to include "value missing from templates: enabled" in [this file](https://github.com/FloSchick/trust-manager/blob/feat/toplevel-chart-enable/deploy/charts/trust-manager/values.linter.exceptions).

This is because the value is not used in any actual helm template. Our linter calls this out. So we need that string to have it exempt from the linter. I worked that out by checking the cert-manager [exemptions here](https://github.com/cert-manager/cert-manager/blob/master/deploy/charts/cert-manager/values.linter.exceptions).

FloSchick:
@hawksight thanks for looking into this, i was not able to resolve this on my own. Added the ingore value.

hawksight:
@FloSchick - thank you for your contribution. I created another PR in #680 so that tests passed and commits were signed. This now merged and hopefully will go out in the next release. Thank you for taking a good first stab at the issue!

PR #583, opened 2025-04-08
Fixup PDB namespace population logic

alloveras:
The current implementation does not account for namespace overrides via Helm Values for the PDB resource.

Instead, it uses `Release.Namespace` or, if undefined, it falls back to the 'default' namespace, which is inconsistent with how the namespace field is populated for all the other resources.

This PR makes the population of the `namespace` field consistent with all the other resources in the Helm Chart.


erikgb (commented):
Nice catch! 💟 Thanks!

/lgtm
/approve

PR #589, opened 2025-04-11
Remove outdated and non-generated apply configurations

erikgb:
Somehow `applyconfiguration-gen` got lost in makefile-module improvements. This means that the applyconfigurations are no longer kept up-to-date with the trust-manager APIs. I would personally prefer to fix the generate issue, but I seem to be the only maintainer who likes the ACs. And since they are unused, at least in this project, let's delete them.

Reverts most of https://github.com/cert-manager/trust-manager/pull/217, but keeps the `genclient` markers in case users are generating something.

Hopefully, there will be more appetite for ACs when https://github.com/kubernetes-sigs/controller-tools/pull/818 is released.


erikgb:
I just noticed someone is actually using apply configurations in this project, ref. https://github.com/cert-manager/cert-manager/issues/5783#issuecomment-2303062857. Maybe I should fix the generator instead? WDYT @inteon 

/hold

inteon:
> I just noticed someone is actually using apply configurations in this project, ref. [cert-manager/cert-manager#5783 (comment)](https://github.com/cert-manager/cert-manager/issues/5783#issuecomment-2303062857). Maybe we should fix the generator instead? WDYT @inteon?
> 
> /hold

Yes, let's maybe do that, `applyconfiguration-gen` should be available through makefile-modules.
I think there is a valid usecase for these resources, just by this repo necessarily (since the trust-manager controller only updates the status of the Bundles). As long as we don't have a separate API package,  we should keep these up-to-date here.

erikgb:
> > I just noticed someone is actually using apply configurations in this project, ref. [cert-manager/cert-manager#5783 (comment)](https://github.com/cert-manager/cert-manager/issues/5783#issuecomment-2303062857). Maybe we should fix the generator instead? WDYT @inteon?
> > /hold
> 
> Yes, let's maybe do that, `applyconfiguration-gen` should be available through makefile-modules. I think there is a valid usecase for these resources, just by this repo necessarily (since the trust-manager controller only updates the status of the Bundles). As long as we don't have a separate API package, we should keep these up-to-date here.

Agreed! Here is an alternative PR: https://github.com/cert-manager/trust-manager/pull/598. PTAL!

PR #594, opened 2025-04-12
Fix inconsistent TestCiphersSuite fails on webhook server not starting

### LINKED ISSUE #593: TestCiphersSuite fails when running unit tests locally
````
=== Failed
=== FAIL: cmd/trust-manager/app TestCiphersSuite/TLS_1.2_client_connect_to_TLS_1.2_server,_ciphers_suite_supported (0.00s)
    ciphers_test.go:76: error to connect to webhook server, Get "https://localhost:6443": dial tcp [::1]:6443: connect: connection refused
````

I also saw this happening in CI yesterday. And the same issue reported by @terricain in https://github.com/cert-manager/trust-manager/pull/582#issuecomment-2799049424.

terricain:
The Webhook server sometimes starts slower than it takes for the `TestCiphersSuite` tests to start running, this PR adds in a short loop to wait for the server to start accepting connections.

Fixes #593

### DIFF
--- a/cmd/trust-manager/app/ciphers_test.go
+++ b/cmd/trust-manager/app/ciphers_test.go
@@ -175,6 +175,22 @@ func setupWebHookServer(tmpDir string) error {
 		cmd.PrintErrf("error: %v\n", err)
 		return err
 	}
+
+	// Wait for the webhook server to be ready
+	for attempt := 0; attempt < 10; attempt++ {
+		var conn net.Conn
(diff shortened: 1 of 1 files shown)


erikgb:
I'm inclined to make a more dramatic change to this, as I don't think an integration test is required here. A simpler unit test should be sufficient IMO, as we don't need to test that TLS options work in Go. Let's keep this open for now while I try to look for a fix in this direction.

inteon:
Thank you for your PR.
@erikgb solved the problem in a different manner in https://github.com/cert-manager/trust-manager/pull/595

Thank you both for looking into this issue and fixing it.

PR #596, opened 2025-04-13
Use upstream metav1.Condition instead of our own BundleCondition

erikgb:
This PR replaces our own `BundleCondition` with the upstream `metav1.Condition`. I also tried to use the upstream `meta.SetStatusCondition` helper, but that will require changes to our test strategy. (Or support for `clock.Clock` upstream.)


inteon (commented):
/approve
/lgtm

PR #602, opened 2025-04-16
Fix slightly misleading PKCS#12 profile API docs

erikgb:
While working on https://github.com/cert-manager/trust-manager/pull/486, I noticed that the API doc (probably copied from cert-manager) is slightly misleading for trust-manager use. This PR attempts to fix this.

/cc @terricain @SgtCoDFish


SgtCoDFish (approved):
/lgtm
/approve

This is definitely an improvement! Thank you!

PR #603, opened 2025-04-16
Use PKCS#12 encoder to encode JKS

erikgb:
While working on https://github.com/cert-manager/trust-manager/pull/486, I think the new API could be greatly simplified by removing explicit support for JKS. And after digging a bit deeper into code, issues and PRs in https://github.com/SSLMate/go-pkcs12, I became convinced that just using this module for trust stores in trust-manager is the way forward. Password-less JKS truststores, which is the new Java default (starting Java 18) is also not supported by our currently used JKS module. Support for it was rejected in https://github.com/pavlo-v-chernykh/keystore-go/issues/37.

In this PR I propose to:

- Deprecate the JKS "additional format" by providing a migration path to PKCS#12 with JKS/Java support.
- Remove all use of https://github.com/pavlo-v-chernykh/keystore-go and instead use https://github.com/SSLMate/go-pkcs12

I have added a (skipped) test that can be used to verify that the Java keytool can read the new "JKS" truststores, but it requires paths to installed Java and cannot be enabled in CI.

We should probably also verify that the trust stores are usable by a Java application. I don't know how simple this would be to test, but we could ask the community for help in testing after releasing an alpha release of trust-manager? Maybe we could specifically ask historical contributors of JKS features?


pkg/bundle/internal/truststore/types_test.go
```diff
@@ -146,3 +143,60 @@ func Test_certAlias(t *testing.T) {
 		t.Fatalf("expected alias to be %q but got %q", expectedAlias, alias)
 	}
 }
+
+func Test_manual_read_JKS(t *testing.T) {
+	t.Skip("these tests require installation of Java to run")
```
SgtCoDFish, line 148: suggestion: maybe we could look in $PATH for keytool, and only skip this test if keytool isn't found?

pkg/bundle/internal/truststore/types_test.go
```diff
@@ -146,3 +143,60 @@ func Test_certAlias(t *testing.T) {
...
+		"Java 17": {
+			keytool: "/usr/lib/jvm/java-17-openjdk-amd64/bin/keytool",
+		},
+		"Java 11": {
+			keytool: "/usr/lib/jvm/java-11-openjdk-amd64/bin/keytool",
+		},
+		"Java 8": {
+			keytool: "/usr/lib/jvm/java-8-openjdk-amd64/bin/keytool",
+		},
+	}
```
SgtCoDFish, line 165: suggestion: I don't know if the differing versions of keytool would meaningfully change this - it's not the same as running different versions of Java AFAIK. Plus, these paths assume amd64 (and so wouldn't work for me)

I'd vote to remove the different versions here and, like in my other suggestion, just run this test if keytool can be found on $PATH.

pkg/bundle/internal/truststore/types_test.go
```diff
@@ -146,3 +143,60 @@ func Test_certAlias(t *testing.T) {
...
+	encoder := NewPKCS12Encoder(v1alpha1.DefaultJKSPassword, v1alpha1.LegacyRC2PKCS12Profile)
+	store, err := encoder.Encode(certPool)
+	if err != nil {
+		t.Fatalf("didn't expect an error but got: %s", err)
+	}
+
+	f, err := os.Create(path.Join(t.TempDir(), "cacerts"))
+	if err != nil {
+		t.Fatalf("didn't expect an error but got: %s", err)
+	}
```
SgtCoDFish, line 183: suggestion: I think we should remove the TempDir we create after the test finishes - especially if we change this to run if any version of keytool is found.

pkg/bundle/internal/truststore/types_test.go
```diff
@@ -146,3 +143,60 @@ func Test_certAlias(t *testing.T) {
...
+
+	for name, test := range tests {
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+
+			cmd := exec.Command(test.keytool, "-list", "-keystore", f.Name(), "-storepass", v1alpha1.DefaultJKSPassword) // #nosec G204
+			out, err := cmd.CombinedOutput()
+			if err != nil {
+				t.Fatalf("cmd.Run() failed with %s\n", err)
```
SgtCoDFish, line 197: nitpick: Don't need the newline in Fatalf!

```suggestion
				t.Fatalf("cmd.Run() failed with %s", err)
```

pkg/bundle/internal/truststore/types_test.go
```diff
@@ -146,3 +143,60 @@ func Test_certAlias(t *testing.T) {
...
+			}
+			t.Logf("combined out:\n%s\n", string(out))
```
SgtCoDFish, line 199: nitpick: As above!

```suggestion
			t.Logf("combined out:\n%s", string(out))
```

SgtCoDFish (approved):
I think this is great, and I'd merge it happily! I've got a bunch of suggestions about the keytool test because I think that'd be a great addition even if it's skipped often - so I'll add a hold in case you want to change anything there

/lgtm
/approve
/hold

pkg/bundle/internal/truststore/types_test.go
erikgb: The Godoc states that is done automatically for you: https://pkg.go.dev/testing#T.TempDir.

pkg/apis/trust/v1alpha1/types_bundle.go
```diff
@@ -128,9 +128,13 @@ type AdditionalFormats struct {
 	// The bundle has "changeit" as the default password.
 	// For more information refer to this link https://cert-manager.io/docs/faq/#keystore-passwords
 	// +optional
+	// Deprecated: Explicit JKS support is subject for removal. Please migrate to PKCS12.
+	// PKCS#12 trust stores created by trust-manager are compatible with Java.
```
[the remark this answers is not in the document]
erikgb: Yes, it's confusing, but I struggled to find a good wording here. Please suggest something! 🙏

pkg/bundle/internal/truststore/types_test.go
erikgb: > it's not the same as running different versions of Java AFAIK.

Important note! Hmmm, then I must find a better way to test this. 🤔

pkg/apis/trust/v1alpha1/types_bundle.go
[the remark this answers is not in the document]
SgtCoDFish: I think maybe instead of "explicit JKS support" if we wrote "Writing JKS files"? It's a little clearer, but the rest of the comment is good.

Really doesn't matter much though!

pkg/bundle/internal/truststore/types_test.go
SgtCoDFish: Oh wow, TIL that even existed! It'd completely passed me by! Awesome 😁

pkg/apis/trust/v1alpha1/types_bundle.go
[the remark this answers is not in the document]
erikgb: Updated! Feel free to suggest something even better.

pkg/bundle/internal/truststore/types_test.go
erikgb: @SgtCoDFish, I hacked together a test that actually runs Java against a Go HTTP server using a self-signed certificate. And it seems to work on my machine for all the LTS Java versions: 8 (EOL), 11, 17 and 21. Do you have Java installed? It would be nice if you could also test! 😅

test/java_client/TestHTTPConnection.java
SgtCoDFish, line 1: Haven't been able to test this yet but I'd like to! Thank you for looking into this

test/java_client/TestHTTPConnection.java
```diff
@@ -0,0 +1,51 @@
+import java.net.*;
+
+/*
+ * This is a simple program to test SSL connections from a Java client.
+ * Credits to https://github.com/ChristopherSchultz/ssltest.
```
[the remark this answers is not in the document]
erikgb: Good catch! This comment originates from the early editions of the code. I am no longer using the same approach, so I just removed the comment.

test/java_client/TestHTTPConnection.java
erikgb: Please do! I am 99% sure this should work fine, and all tests passes on my machine with Java 8, 11, 17 and 21.

wallrj: It works on my Windows + WSL2  Ubuntu Linux laptop:
```
$ sudo apt install openjdk-21-jdk-headless
...

$ go test -v ./pkg/bundle/internal/truststore/...
=== RUN   Test_Java_keytool_can_read_JKS
    java_test.go:73: combined out:
        Keystore type: PKCS12
        Keystore provider: SUN

        Your keystore contains 3 entries

        3c95e845|cn=cmct-test-root,o=cert-manager, Apr 24, 2025, trustedCertEntry,
        Certificate fingerprint (SHA-256): 3C:95:E8:45:AC:75:2D:FC:13:A9:8B:71:67:41:7E:89:2D:83:02:BC:0F:FB:48:7F:10:A9:DF:AA:C0:D1:26:73
        548b988f|cn=cmct-test-root,o=cert-manager, Apr 24, 2025, trustedCertEntry,
        Certificate fingerprint (SHA-256): 54:8B:98:8F:4B:AD:7B:DD:0D:3B:75:23:DE:37:15:4E:E4:7F:28:5E:EE:36:D3:B1:F5:3F:AA:27:20:FC:A3:07
        96bcec06|cn=isrg root x1,o=internet security research group,c=us, Apr 24, 2025, trustedCertEntry,
        Certificate fingerprint (SHA-256): 96:BC:EC:06:26:49:76:F3:74:60:77:9A:CF:28:C5:A7:CF:E8:A3:C0:AA:E1:1A:8F:FC:EE:05:C0:BD:DF:08:C6
--- PASS: Test_Java_keytool_can_read_JKS (0.22s)
=== RUN   Test_Java_can_use_JKS
    java_test.go:136: combined out:
        Note: TestHTTPConnection.java uses or overrides a deprecated API.
        Note: Recompile with -Xlint:deprecation for details.
    java_test.go:168: combined out:
        Successfully connected to https://127.0.0.1:42583
```

wallrj (approved):
Thanks @erikgb 

@ali-hamza-noor and I have tested the new Java integration tests on Ubuntu Linux and Mac.

Ash has reviewed the code, so I'll approve this, as discussed at the cert-manager standup 2025-04-24.

/approve
/lgtm

PR #615, opened 2025-04-29
Adding common annotations for all the resources

ali-hamza-noor:
Adding a way for the users to add annotations for all the trust-manager resources. This PR is being raised as part of this [issue](https://github.com/cert-manager/trust-manager/issues/597) which indicates a use case for using argocd sync-waves.

Fixes #597 

Close https://github.com/cert-manager/trust-manager/pull/214

PR #619, opened 2025-05-02
Update LICENSES file

erikgb:
It seems like the licenses generation is non-idempotent. Example: https://github.com/cert-manager/trust-manager/pull/615/files

To avoid unreleased changes on a functional PR, I suggest updating the file in a separate PR.


erikgb:
Seems to be a bug upstream making the result inconsistent. @inteon will look into it.

Issue #276, opened 2024-01-18
Improve filtered certs error reporting


Hoega:
When creating a bundle from a source (configmap, secret, or inline) with exclusively expired certificates, the bundle will fail and indicate that the source contains no PEM data. There is no indication that certs are expired and it can be tricky to troubleshoot.

This is because[ 'ValidateAndSanitizePEMBundleWithOptions' parses sources individually](https://github.com/cert-manager/trust-manager/blob/83f1a23e11537da8bee7a4ce4d99802a19035b85/pkg/bundle/sync.go#L119C38-L119C38), therefore if a source has only expired cert(s), the cert(s) will be skipped and[ the sanitize function will find an empty output thus returning an error](https://github.com/cert-manager/trust-manager/blob/83f1a23e11537da8bee7a4ce4d99802a19035b85/pkg/util/pem.go#L73C1-L74C1).

For example, if I have: 
- 1 valid cert in configmap source
- 1 valid cert in secret source
- 1 expired cert in inline source

The bundle will fail with this error message:
`  Warning  SourceBuildError  1s (x12 over 12s)  bundles  Failed to build bundle sources: invalid PEM data in source: bundle contains no PEM certificates`

Also, the bundle is pending: 
```
NAME         TARGET   SYNCED   REASON   AGE
my-org.com                              5m15s
```
Whereas If I add 1 valid cert to the inline source the bundle would be OK. This is because the sanitize function requires at least one valid cert by source.

I propose to parse and sanitize sources globally and/or at least to improve the error reporting to highlight a cert is expired.

arsenalzp:
Can we accept this as the feature I can work for?

erikgb:
> Can we accept this as the new feature? I think I can work on it once the decision is made.

Please work on it! I think we all agree that the error handling could be improved here. Could you start by describing the proposed solution in more detail? I agree this would become a lesser problem if a bundle is sanitized after reading all sources. But what if the resulting bundle would contain exclusively expired certificates?

arsenalzp:
> > Can we accept this as the new feature? I think I can work on it once the decision is made.
> 
> Please work on it! I think we all agree that the error handling could be improved here. Could you start by describing the proposed solution in more detail? I agree this would become a lesser problem if a bundle is sanitized after reading all sources. But what if the resulting bundle would contain exclusively expired certificates?

Hello,
Thank you for letting me work again!
[...]

Issue #434, opened 2024-09-10
Update bundle status with error when encountering problem with sources files


Shawcs:
Hi !

I noticed that the status do not reflect problem with bad/empty sources.

---
### How to reproduce:

create a bundle that target a ConfigMap with a valid cert entry
```yaml
apiVersion: trust.cert-manager.io/v1alpha1
kind: Bundle
metadata:
  name: my-bundle
spec:
  sources:
    - configMap:
        key: ca.crt
        selector:
          matchLabels:
            trust-bundle.my-bundle: include
  target:
    additionalFormats:
      pkcs12:
        key: ca.p12
        password: changeit
    namespaceSelector:
      matchLabels:
        kubernetes.io/metadata.name: infra
    secret:
      key: ca.crt
```
```yaml
kind: ConfigMap
apiVersion: v1
metadata:
  name: my-cert.crt
  namespace: infra-cert-manager
  labels:
    trust-bundle.my-bundle: include
data:
  ca.crt: |
    -----BEGIN CERTIFICATE-----
    MIIDojCCAYqgAwIBAgIQV5ocp05c1d2ULNLEDrdCpTANBgkqhkiG9w0BAQsFADBH
    MQswCQYDVQQGEwJDSDEnMCUGA1UEChMeQmFucXVlIExvbWJhcmQgT2RpZXIgZXQg
    Q2llIFNBMQ8wDQYDVQQDEwZMTyBEUEkwHhcNMjQwMzA3MDAwMDAwWhcNMjUwMzA3
    MjM1OTU5WjAVMRMwEQYDVQQDEwpnaXRodWIuY29tMFkwEwYHKoZIzj0CAQYIKoZI
    zj0DAQcDQgAEQrPuGOisrWzPTzsVzujNAMvKeM1GRDs18c2N5R6LemewOMjO0Ep1
    yESxF/xn4Zj7tlsTeMT5zz4Li1DQN/K1zKOBhjCBgzAdBgNVHQ4EFgQUO2g/NDr1
    RzTK76ZOPZq9Xm56zJ8wDgYDVR0PAQH/BAQDAgeAMAwGA1UdEwEB/wQCMAAwHQYD
    VR0lBBYwFAYIKwYBBQUHAwEGCCsGAQUFBwMCMCUGA1UdEQQeMByCCmdpdGh1Yi5j
[...]

PR #624, opened 2025-05-04
CertPool should not error input contains no certificates

erikgb:
While working on https://github.com/cert-manager/trust-manager/pull/623, I discovered that `CertPool` skips invalid input data. This results in an error (`no non-expired certificates found in input bundle`) only if the input adds no certificates to the pool. This does not seem right to me, and I would at least prefer consistent behavior. As a first step, I suggest removing the check requiring the input to add at least one certificate to the pool. This will also fix https://github.com/cert-manager/trust-manager/issues/276, which I wrongly closed as fixed.

As a second step, to address https://github.com/cert-manager/trust-manager/issues/434, I would prefer if `CertPool` didn't just ignore invalid PEM data. But there might be a reason for doing it this way? The used function to decode, [pem.Decode](https://pkg.go.dev/encoding/pem#Decode), seems pretty ignorant.

Fixes https://github.com/cert-manager/trust-manager/issues/276


pkg/util/cert_pool.go
```diff
@@ -90,13 +90,11 @@ func NewCertPool(options ...Option) *CertPool {
 // https://www.rfc-editor.org/rfc/rfc7468
 //
 // Additionally, if the input PEM bundle contains no non-expired certificates, an error is returned.
```
farazkhawaja, line 92: does this line also needs to go away? @erikgb

PR #642, opened 2025-06-20
chore: Bump go to 1.24.4 in both applications

hawksight:


hawksight:
Did not realise I was bumping the minimum go version ([ref](https://go.dev/doc/modules/gomod-ref#go-syntax)).
Will close this as it appears latest go is [in the makefiles](https://github.com/cert-manager/trust-manager/blob/357ea506222454e305a86e0666fd49305304d45d/make/_shared/tools/00_mod.mk#L175).

PR #647, opened 2025-07-03
WIP: ClusterBundle source API rework

erikgb:


erikgb:
The more reviews, the better. So to avoid a premature merge of this PR
/hold

pkg/apis/trustmanager/v1alpha2/types_cluster_bundle.go
```diff
@@ -64,6 +64,21 @@ type BundleSpec struct {
...
+	// If default CAs were not configured at start-up, any request to use the default
+	// CAs will fail.
+	// The version of the default CA package which is used for a Bundle is stored in the
+	// defaultCAPackageVersion field of the Bundle's status field.
+	// +optional
+	IncludeDefaultCAs *bool `json:"includeDefaultCAs,omitempty"`
+
+	// InLine is a simple string to append as the source data.
+	// +optional
+	InLineCAs *string `json:"InLineCAs,omitempty"`
```
SgtCoDFish, line 80: note: As we discussed this morning, there's the possibility this will be removed in a future version and marked as deprecated here. We could add a deprecation notice here, or do it elsewhere - doesn't really matter!

(Just wanted to capture that discussion in text form)

pkg/apis/trustmanager/v1alpha2/types_cluster_bundle.go
```diff
@@ -72,30 +87,17 @@ type BundleSpec struct {
...
-	// If default CAs were not configured at start-up, any request to use the default
-	// CAs will fail.
-	// The version of the default CA package which is used for a Bundle is stored in the
-	// defaultCAPackageVersion field of the Bundle's status field.
-	// +optional
-	UseDefaultCAs *bool `json:"useDefaultCAs,omitempty"`
+	// IncludeAllKeys is a flag to include all keys in the object's `data` field to be used. False by default.
+	// This field must not be true when `Key` is set.
+	//+optional
+	IncludeAllKeys bool `json:"includeAllKeys,omitempty"`
```
SgtCoDFish: suggestion: I'm trying to think forwards and the obvious future source we'd add is ClusterTrustBundle (CTB), right?

Neither of these fields would be appropriate for CTB - there's only one key and it's `trustBundle` (https://kubernetes.io/docs/reference/access-authn-authz/certificate-signing-requests/#ctb-signer-unlinked)

How do you envision CTB fitting into new source abstraction down the road?

SgtCoDFish: suggestion: I found it a little difficult from reading the description of the IncludeAllKeys field here to work out what it's for.

```suggestion
	// IncludeAllKeys, if true, will include all keys in the source's `data` field when extracting certificates.
	// Defaults to "false". This field must not be true when `Key` is set.
	//+optional
	IncludeAllKeys bool `json:"includeAllKeys,omitempty"`
```

erikgb: Both `Key` and `IncludeAllKeys` are optional fields, so this new structure should fit well for CTB, where a user would only set the fields of `SourceReference`. And the idea is to use CEL to validate the use of optional fields. We could create another (inline) struct for the fields only relevant for configmaps/secrets to make this even clearer. WDYT? As always, a good name would be the biggest challenge. Maybe `KeySelector`, `MappedKeyRef` or `DataKeyRef`.

erikgb:
> Another possible source would be like `trustManagerPodFS: {path: "/etc/foo.pem"}` to load the certs at `/etc/foo.pem` from the trust-manager pod's filesystem - and that doesn't fit into the new source type either.

This is a good question. Are you thinking about including certificates through trust-manager pod volumes, like https://kubernetes.io/docs/concepts/storage/volumes/#image? I will try to see how this can be adopted.

SgtCoDFish (approved):
/lgtm
/approve

(Change was a simple rebase)

PR #654, opened 2025-07-16
Add design for trust source plugins

SgtCoDFish:
Follows discussion in standup on 2025-07-16


design/20250716-source-plugins.md
```diff
@@ -0,0 +1,80 @@
...
+```yaml
+apiVersion: example.io/v1alpha1
+kind: ImageTrustSource
+metadata:
+  name: example
+spec:
+  image: docker.io/library/debian:bookworm
+  path: /etc/ssl/certs/ca-certificates.crt
+status:
+  bundleData: |
```
ThatsMrTalbot, line 66: To avoid object size limits I would have the status expose an array of secrets that contain the actual bundle, most implementations would probably only ever need a single secret, but cant hurt to future proof it in the design.

Something like:
```yaml
status:
  bundleSecrets:
  - name: example-bundle-01
  - name: example-bundle-02
```

For namespaced sources the secrets would exist in the same namespace as the resource, for cluster sources the secrets would exist in the trust-manager namespace

design/20250716-source-plugins.md
```diff
@@ -0,0 +1,80 @@
...
+    MIID...
+    -----END CERTIFICATE-----
+```
+
+The bundle data from the source would then be fed into the trust-manager target, in a similar way to how the existing sources are used.
+
+## Open Questions
+
+- If the source plugins need a separate controller, why should that controller not simply write its data into a ConfigMap or Secret, which trust-manager can then consume natively?
+  - Concretely: What do we gain from having CRDs here?
```
ThatsMrTalbot, line 77: 1. If a source produces a bundle larger than 1mb then it will need to shard it across multiple secrets, you would need to _know_ that to include all the shards in your trust bundle. By referencing the CRD you get a single object you can reference
2. It gives us the option of having cluster scope sources, if you just reference secrets you can never have a namespaced TrustBundle reference anything outside its own namespace. I think cluster scope sources would be a great way for us to ship our "default" CAs with a `ClusterImageTrustSource`
3. I think its better UX. If you used secrets then a `ImageTrustSource` would ether need to produce a deterministic secret name or have a configurable secret name, and IMO the secret is just an implementation detail.

design/20250716-source-plugins.md
```diff
@@ -0,0 +1,80 @@
...
+  - Does this make setup difficult? Plugin CRDs should be able to be installed before trust-manager.
```
ThatsMrTalbot, line 78: The controller-runtime uses a dynamic discovery implementation, so CRDs do not need to be installed first.

design/20250716-source-plugins.md
```diff
@@ -0,0 +1,80 @@
...
+- Does this require every plugin to bind a role with CRD read permissions to the trust-manager service account?
```
ThatsMrTalbot, line 79: This is a good point, probably yes

design/20250716-source-plugins.md
```diff
@@ -0,0 +1,80 @@
...
+- Does the key from the status field need to be configurable? Most likely it does - how do we configure it?
```
ThatsMrTalbot, line 80: I would not have it configurable, I would have a "contract" we have with plugins that says you must have this field to function. That is how ClusterAPI works for example, the providers have their own CRDs that work with ClusterAPI if they fulfil the API contract.

design/20250716-source-plugins.md
erikgb: If any source exceeds the size limit, the targets will also exceed the size limit. I don't think we need to account for this additional complexity.

ThatsMrTalbot: IIRC we have had issues raised in the past about peoples bundles exceeding the size, I would argue that is where a CSI driver could help 🙂 

I also wonder if post-quantum will increase CA cert size enough to make hitting the 1mb limit more likely

PR #658, opened 2025-07-18
Final minor adjustments to new ClusterBundle API

erikgb:
After scrutinizing the new (unpublished) `ClusterBundle` API, I found a few potential improvements. I have created separate commits for each change, so this PR is better reviewed commit by commit. A summary of the proposed API changes:

1. Make `.spec.target` optional. This was done for the current `Bundle` in https://github.com/cert-manager/trust-manager/pull/661.
2. Drop the source `includeAllKeys` field in favor of simple wildcard (`*` only) support in the source `key`. This change allows us to make the source `key` mandatory, thus simplifying overall validation. We can always make it optional later on, when adding support for key-less sources. Since I am planning to use https://pkg.go.dev/path#Match for matching keys, I also added regex pattern validation to the source `key` field.
3. Added target `key` field regex pattern validation, inspired by the changes in source `key` (but no wildcard support here).

In addition, I fixed a minor bug (in<->out) in the conversion logic.

And finally I added the new Go 1.24 `omitzero` marker to bundle status. Since the status field is not a struct-pointer, we can ensure safer marshalling by adding this marker. This change was also backported to the current Bundle API.

### DIFF
--- a/pkg/apis/trustmanager/v1alpha2/types_cluster_bundle.go
+++ b/pkg/apis/trustmanager/v1alpha2/types_cluster_bundle.go
@@ -90,15 +90,12 @@ type BundleSpec struct {
 type BundleSource struct {
 	SourceReference `json:",inline"`
 
-	// Key of the entry in the object's `data` field to be used.
-	//+optional
+	// Key(s) of the entry in the object's `data` field to be used.
+	// Wildcards "*" in Key matches any sequence characters.
+	// A Key containing only "*" will match all data fields.
 	// +kubebuilder:validation:MinLength=1
-	Key string `json:"key,omitempty"`
-
-	// IncludeAllKeys, if true, will include all keys in the source's `data` field when extracting certificates.
-	// Defaults to "false". This field must not be true when `Key` is set.
-	//+optional
-	IncludeAllKeys bool `json:"includeAllKeys,omitempty"`
+	// +kubebuilder:validation:Pattern=`^[0-9A-Za-z_.\-*]+$`
+	Key string `json:"key"`
 }
 
 // BundleTarget is the target resource that the Bundle will sync all source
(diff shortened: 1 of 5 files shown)


pkg/apis/trustmanager/v1alpha2/types_cluster_bundle.go
```diff
@@ -80,7 +80,8 @@ type BundleSpec struct {
 	InLineCAs *string `json:"InLineCAs,omitempty"`
 
 	// Target is the target location in all namespaces to sync source data to.
-	Target BundleTarget `json:"target"`
+	// +optional
+	Target *BundleTarget `json:"target,omitempty"`
```
erikgb: In https://github.com/cert-manager/trust-manager/pull/661, it seemed simpler to add the `omitzero` marker instead of making the field a pointer. Any thoughts here? Pointers makes things more complicated, but still considered best practice for optional struct fields AFAIK.

erikgb:
I suggest that we wait to merge this until https://github.com/cert-manager/trust-manager/pull/661 has landed.
/hold

pkg/apis/trustmanager/v1alpha2/types_cluster_bundle.go
ThatsMrTalbot: I've started using non pointers and omitzero, I think pointers offers a nice `foo != nil` for checking if the struct is set, but the alternative `foo != StructThing{}` is not bad enough that I would keep using pointers.

It also reduces the chances of nil pointer explosions.

erikgb: I prefer to avoid pointers when it makes sense, and have now updated this PR with the same `omitzero` as in https://github.com/cert-manager/trust-manager/pull/661. It makes the conversion simpler and avoids potential nil pointers.

PR #661, opened 2025-07-20
Make bundle target optional

erikgb:
This PR makes the bundle target spec optional, and also makes it possible to remove a target configmap/secret configuration. Somehow, we were validating this in our update validating webhook, but I consider it a bug.

Relates to https://github.com/cert-manager/trust-manager/pull/658. This backport of (non-breaking) allowing no targets will simplify the migration to `ClusterBundle`.

Also, removing the validation of the target key from the validating webhook, as the Bundle schema fully validates this.


SgtCoDFish (approved):
/lgtm
/approve

Cheers!

PR #662, opened 2025-07-21
Generate CRD for new ClusterBundle API

erikgb:
This adds generated CRDs for the new ClusterBundle API, ref. https://github.com/cert-manager/trust-manager/pull/485. Since controller-gen does not allow me to generate a CRD without a served version, I had to add a dummy v1alpha served version. This is not optimal, but generating the CRD will allow me to continue adding API validations.


SgtCoDFish (approved):
/lgtm
/approve

Erik explained in standup this morning that this step would help with development of the new `ClusterBundle`. Looks great to me!

PR #664, opened 2025-07-22
Add ClusterBundle API validations

erikgb:
This PR adds all API validations for the new `CLusterBundle` API that I was able to implement without adding a validating webhook. I have added a new "integration" test, based on envtest (simplified control plane), to test these validations. I suggested introducing CEL validations in https://github.com/cert-manager/trust-manager/pull/475, but the `Bundle` API wasn't really ready for it, as the expressions became way too complex. But in the new API, I think the CEL expressions are simple enough.....

Label selectors cannot currently be validated using CEL expressions (or OpenAPI schema) and require a webhook for validation. This is probably true for labels/annotations, even if Kubernetes is now providing the [format CEL library](https://kubernetes.io/docs/reference/using-api/cel/#kubernetes-format-library) that might be used for this. I am personally inclined to NOT add a webhook just for this, as it will hopefully be available in CEL later on. And no webhooks implies that trust-manager will not have cert-manager as a prerequisite.

The added tests detected a few minor bugs/improvements in the API:
- PKCS12 password cannot have an API default to be validated correctly. We document the default value and will use a controller default to implement it.
- We have to allow empty `sources`, as the `includeDefaultCAs` has moved out of the `sources` array field. (And also the `inLineCAs` field, which we are planning to deprecate.)
- The `inLineCAs` field had the wrong init-case in the JSON serialization marker, which is now fixed.
- We should set max items for all array fields and max length for string fields to ensure CEL cost estimation doesn't blow up.

### DIFF
--- a/deploy/crds/trust-manager.io_clusterbundles.yaml
+++ b/deploy/crds/trust-manager.io_clusterbundles.yaml
@@ -59,7 +59,7 @@ spec:
           spec:
             description: Desired state of the Bundle resource.
             properties:
-              InLineCAs:
+              inLineCAs:
                 description: InLine is a simple string to append as the source data.
                 type: string
               includeDefaultCAs:
@@ -99,6 +99,7 @@ spec:
                       description: |-
                         Name is the name of the source object in the trust Namespace.
                         This field must be left empty when `selector` is set
+                      maxLength: 253
                       minLength: 1
                       type: string
                     selector:
@@ -154,8 +155,12 @@ spec:
                   - kind
                   type: object
                   x-kubernetes-map-type: atomic
+                  x-kubernetes-validations:
+                  - message: 'exactly one of the following fields must be provided:
+                      [name, selector]'
+                    rule: '[has(self.name), has(self.selector)].exists_one(x,x)'
                 maxItems: 100
-                minItems: 1
+                minItems: 0
                 type: array
                 x-kubernetes-list-type: atomic
               target:
@@ -183,12 +188,11 @@ spec:
                               type: string
                             key:
                               description: Key is the key of the entry in the object's
-                                `data`field to be used.
+                                `data` field to be used.
                               minLength: 1
                               pattern: ^[0-9A-Za-z_.\-]+$
                               type: string
                             password:
-                              default: ""
                               description: |-
                                 Password for PKCS12 trust store.
                                 By default, no password is used (password-less PKCS#12).
@@ -214,6 +218,18 @@ spec:
                           - key
                           type: object
                           x-kubernetes-map-type: atomic
+                          x-kubernetes-validations:
+                          - fieldPath: .password
+                            message: may only be set when format is 'PKCS12'
+                            reason: FieldValueForbidden
+                            rule: '!has(self.password) || (has(self.format) && self.format
+                              == ''PKCS12'')'
+                          - fieldPath: .profile
+                            message: may only be set when format is 'PKCS12'
+                            reason: FieldValueForbidden
+                            rule: '!has(self.profile) || (has(self.format) && self.format
+                              == ''PKCS12'')'
+                        maxItems: 10
                         minItems: 1
                         type: array
                         x-kubernetes-list-map-keys:
@@ -229,12 +245,22 @@ spec:
                             description: Annotations is a key value map to be copied
                               to the target.
                             type: object
+                            x-kubernetes-validations:
+                            - message: must not use forbidden domains as prefixes
+                                (e.g., trust-manager.io)
+                              reason: FieldValueForbidden
+                              rule: self.all(k, !k.startsWith('trust-manager.io/'))
                           labels:
                             additionalProperties:
                               type: string
                             description: Labels is a key value map to be copied to
                               the target.
                             type: object
+                            x-kubernetes-validations:
+                            - message: must not use forbidden domains as prefixes
+                                (e.g., trust-manager.io)
+                              reason: FieldValueForbidden
+                              rule: self.all(k, !k.startsWith('trust-manager.io/'))
                         type: object
                     required:
                     - data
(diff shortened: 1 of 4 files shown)


deploy/crds/trust-manager.io_clusterbundles.yaml
```diff
@@ -183,12 +188,11 @@ spec:
                               type: string
                             key:
                               description: Key is the key of the entry in the object's
-                                `data`field to be used.
+                                `data' field to be used.
```
hawksight:
```suggestion
                                'data' field to be used.
```

pkg/apis/trustmanager/v1alpha2/types_cluster_bundle.go
```diff
@@ -206,8 +210,10 @@ func (t *KeyValueTarget) GetLabels() map[string]string {
 
 // TargetKeyValue is the specification of a key with value in a key-value target resource.
 // +structType=atomic
+// +kubebuilder:validation:XValidation:rule="!has(self.password) || (has(self.format) && self.format == 'PKCS12')", reason=FieldValueForbidden, fieldPath=".password", message="may only be set when format is 'PKCS12'"
+// +kubebuilder:validation:XValidation:rule="!has(self.profile) || (has(self.format) && self.format == 'PKCS12')", reason=FieldValueForbidden, fieldPath=".profile", message="may only be set when format is 'PKCS12'"
 type TargetKeyValue struct {
-	// Key is the key of the entry in the object's `data`field to be used.
+	// Key is the key of the entry in the object's `data' field to be used.
```
hawksight:
```suggestion
	// Key is the key of the entry in the object's 'data' field to be used.
```

erikgb: Nice catch, thanks! We should at least use the same quote type on both ends. 😆

test/integration/bundle/cluster_bundle_validation_test.go
```diff
@@ -0,0 +1,275 @@
...
+					if matchErr != "" {
+						Expect(cl.Create(ctx, bundle)).Should(MatchError(ContainSubstring(matchErr, field)))
+					} else {
+						Expect(cl.Create(ctx, bundle)).To(Succeed())
+					}
+				},
+				Entry("when unset", trustmanagerapi.BundleFormat(""), ""),
+				Entry("when PEM", trustmanagerapi.BundleFormatPEM, ""),
+				Entry("when PKCS12", trustmanagerapi.BundleFormatPKCS12, ""),
+				Entry("when unknown", trustmanagerapi.BundleFormat("JKS"), "spec.target.%s.data[0].format: Unsupported value: \"JKS\": supported values: \"PEM\", \"PKCS12\""),
```
SgtCoDFish: question:

1) I presume the validation would reject "pem" or "PeM"? If so, is there any way to allow users to specify a value in a case-insensitive way? 

2) Either way, can we add a test for a bundle format with a different case just to capture what the behaviour is?

test/integration/bundle/cluster_bundle_validation_test.go
```diff
@@ -0,0 +1,275 @@
...
+					Expect(cl.Create(ctx, bundle)).Should(MatchError(ContainSubstring(matchErr, field, pkcs12Field)))
+
+					targetKeyValue.Format = trustmanagerapi.BundleFormatPKCS12
+					selectorAccessor(&trustmanagerapi.KeyValueTarget{
+						Data: []trustmanagerapi.TargetKeyValue{targetKeyValue},
+					})
+					Expect(cl.Create(ctx, bundle)).To(Succeed())
+				},
+				Entry("when password set", trustmanagerapi.PKCS12{Password: ptr.To("my-password")}),
+				Entry("when profile set", trustmanagerapi.PKCS12{Profile: trustmanagerapi.Modern2023PKCS12Profile}),
```
SgtCoDFish: question: are we missing tests for when profile != PKCS12 but "password" is provided? I might be missing something

pkg/apis/trustmanager/v1alpha2/types_cluster_bundle.go
```diff
@@ -238,10 +244,12 @@ const (
 type TargetMetadata struct {
 	// Annotations is a key value map to be copied to the target.
 	// +optional
+	// +kubebuilder:validation:XValidation:rule="self.all(k, !k.startsWith('trust-manager.io/'))", reason=FieldValueForbidden, message="must not use forbidden domains as prefixes (e.g., trust-manager.io)"
```
SgtCoDFish, line 247: question: Does this mean that if we add trust-manager specific annotations or labels, we'd need to list them all in these comments here?

SgtCoDFish (commented):
Few questions - I don't see any reason this can't merge but maybe there are some extra tests which would be helpful?

test/integration/bundle/cluster_bundle_validation_test.go
erikgb: No, AFAIK there is no way to make enums case-insensitive. I will add a test to verify the behavior.

pkg/apis/trustmanager/v1alpha2/types_cluster_bundle.go
erikgb: Yes, if we want to allow certain annotations/labels with the `trust-manager.io/`-prefix **to trust-manager CRD(s)**, we need to add exclude rules here. FYI: This validation is currently enforced by our validating webhook, so really no change in behavior.

test/integration/bundle/cluster_bundle_validation_test.go
erikgb: Good catch! It was there, but a bit hidden. I have now refactored the test fixture to make it clearer.

test/integration/bundle/cluster_bundle_validation_test.go
erikgb: I have added a test for that now.

pkg/apis/trustmanager/v1alpha2/types_cluster_bundle.go
SgtCoDFish: Yeah fair enough - it's definitely not a problem, I was just curious :)

PR #668, opened 2025-08-01
Add ClusterBundle validating webhook

erikgb:
In this PR, I create a new validating webhook for `ClusterBundle` to cover the validations that still require a validating webhook. Hopefully, we can get rid of the validating webhooks completely when more validations are supported by Kubernetes CEL. I haven't added unit tests for the new validating webhook, but we should have very good coverage in the integration tests. This is mainly done because I hope we can get rid of the webhooks at some point.

Note: I have also made the "official" (Helm) `ValidatingWebhookConfiguration` more precise to avoid it from diverging from the simplified configuration used in the integration tests. This change should probably go in a separate PR, but I was a little in doubt. Please let me know what you think!

### DIFF
--- a/pkg/webhook/cluster_bundle.go
+++ b/pkg/webhook/cluster_bundle.go
@@ -1,0 +1,113 @@
+/*
+Copyright 2021 The cert-manager Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package webhook
+
+import (
+	"context"
+	"fmt"
+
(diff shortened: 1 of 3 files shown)


pkg/webhook/cluster_bundle.go
```diff
@@ -0,0 +1,113 @@
...
+
+func (webhook *ClusterBundle) ValidateDelete(_ context.Context, _ runtime.Object) (admission.Warnings, error) {
+	// always allow deletes
+	return nil, nil
+}
+
+func (webhook *ClusterBundle) validate(ctx context.Context, obj runtime.Object) (admission.Warnings, error) {
+	bundle, ok := obj.(*trustmanagerapi.ClusterBundle)
+	if !ok {
+		return nil, fmt.Errorf("expected a Bundle, but got a %T", obj)
```
SgtCoDFish: suggestion:

```suggestion
		return nil, fmt.Errorf("expected a ClusterBundle, but got a %T", obj)
```

SgtCoDFish (changes requested):
One small thing, just needs updating - otherwise looks great 😁

pkg/webhook/cluster_bundle.go
erikgb: Great catch. Those evil copy-modify practices. 🙊  PTAL!

SgtCoDFish (approved):
/lgtm
/approve

Great stuff, cheers!

PR #669, opened 2025-08-03
Refactor scheme setup to support multi-group APIs

erikgb:
While working on https://github.com/cert-manager/trust-manager/pull/668, I faced issues with the GlobalScheme in the `trust.cert-manager.io` API group package. This refactoring will allow us to better support multiple groups while migrating from Bundle to ClusterBundle.


cmd/trust-manager/app/app.go
```diff
@@ -51,6 +54,13 @@ into an output bundle 'target', which can then be made available to all
 services across a cluster.`
 )
 
+var scheme = runtime.NewScheme()
+
+func init() {
+	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
+	utilruntime.Must(trustapi.AddToScheme(scheme))
+}
+
```
SgtCoDFish: suggestion: If we're going to refactor the GlobalScheme stuff (which I think is good!) I think we should avoid the init function here if we can. init functions are just so confusing and dangerous.

Could we create and initialise `scheme` inside `NewCommand`, and avoid the init function?

erikgb: We can, but are you OK to use the `utilruntime.Must` ? 🤔

erikgb: I have moved it inside now. PTAL!

SgtCoDFish (approved):
/lgtm
/approve

Can't see a reason to not merge, thank you :)

PR #670, opened 2025-08-03
Make Bundle webhook configuration precise

erikgb:
While working on https://github.com/cert-manager/trust-manager/pull/668, I noticed that the validating webhook configuration for bundles is imprecise. Since I want to add a sibling webhook for ClusterBundle, I think this should be corrected.


SgtCoDFish (approved):
/lgtm
/approve

Thank you! 😁

PR #671, opened 2025-08-03
Add missing liveness/health probe

erikgb:
This PR adds a missing liveness/health probe to trust-manager. It's already configured on the c/r manager (same port as the readiness probe), but the liveness probe on the Deployment was missing. And also adding the webhook server health check. I am pretty sure c/r already contains the logic to wait for caches to be synced, so I remove that check from the readiness check.

Also, a minor refactoring to make room for multiple webhooks in future PRs (`ClusterBundle`).

Another thing I discovered while working on https://github.com/cert-manager/trust-manager/pull/668.


SgtCoDFish:
We agreed in standup today to split this into two PRs, because livenessprobe specifically might need more discussion

erikgb:
> We agreed in standup today to split this into two PRs, because livenessprobe specifically might need more discussion

@SgtCoDFish I have now updated the PR and removed adding the liveness probe to trust-manager Deployment (even if I think it should be added). PTAL!

SgtCoDFish (approved):
/lgtm
/approve

Discussed in standup, we agreed that we probably want both a readiness and liveness probe in the longer term and this is a good step forwards towards that!

PR #673, opened 2025-08-03
Remove use of deprecated c/r Requeue

erikgb:
Just returning the error and letting the default error handling kick in seems like a viable option.


pkg/bundle/bundle.go
SgtCoDFish, line 1: comment: This is hard to review because controller-runtime's godoc is rough, IMO. Looking at https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.21.0/pkg/reconcile#Reconciler I'm not told _at all_ what the Result and error mean when returned by a Reconciler... so I have to do more digging to work out the actual practical effect of this is :(

pkg/bundle/bundle.go
```diff
@@ -288,7 +288,7 @@ func (b *bundle) reconcileBundle(ctx context.Context, req ctrl.Request) (result
 				},
 			)
 
-			return ctrl.Result{Requeue: true}, statusPatch, nil
```
SgtCoDFish, line 291: comment: This is the biggest change here, but I'm happy from looking at a controller-runtime [example](https://github.com/kubernetes-sigs/controller-runtime/blob/e8c5c5445a20af34453aab1194fc7d64464d2150/example_test.go#L217) that this behavior makes sense and will trigger a requeue

PR #676, opened 2025-08-06
feat(helm): Support revisionhistorylimit

DrFaust92:


deploy/charts/trust-manager/values.yaml
```diff
@@ -29,6 +29,13 @@ crds:
 # +docs:type=number,string,null
 replicaCount: 1
 
+# The number of old ReplicaSets to retain to allow rollback.
+# This is used to control the number of old ReplicaSets that are retained to allow rollback.
+# If set to 0, no old ReplicaSets are retained.
+# If set to null, the default value of 10 is used.
```
```suggestion
```

Is this true? Is the field nullable in K8s?

DrFaust92: default value is indeed 10, i dont know if setting null is 10 or 0 (0 is not to retain replicasets) but i assume 10 as its the value of not setting it

erikgb:
@DrFaust92 Could you please run `make generate` and squash everything to a single commit?

I'll probably cut a trust-manager release tomorrow, so if you want this new feature soon please act. 😄

DrFaust92:
> @DrFaust92 Could you please run `make generate` and squash everything to a single commit?
> 
> I'll probably cut a trust-manager release tomorrow, so if you want this new feature soon please act. 😄

thanks @erikgb , done

PR #680, opened 2025-08-08
feat: Add a global value of enabled for wrapping trust-manager chart

### LINKED ISSUE #574: Add a top level variable called `enabled` to the Helm chart schema

hawksight:
Fixes #574.

Closes #578 as this PR should replace that one.

I have hidden this option in documentation just like the [cert-manager example here](https://github.com/cert-manager/cert-manager/blob/master/deploy/charts/cert-manager/values.yaml#L1481-L1482).

/cc @erikgb

### DIFF
--- a/deploy/charts/trust-manager/values.yaml
+++ b/deploy/charts/trust-manager/values.yaml
@@ -206,7 +206,7 @@ filterExpiredCertificates:
 app:
   # Minimum TLS version supported. If omitted, the default Go minimum version will be used.
   minTLSVersion: ""
-  
+
   # Comma-separated list of cipher suites for the server. If omitted, the default Go cipher suites will be used.
   cipherSuites: ""
 


deploy/charts/trust-manager/values.yaml
```diff
@@ -377,3 +376,10 @@ commonAnnotations: {}
 #        - toEntities:
 #            - kube-apiserver
 extraObjects: []
+
+# Field to optionally disable installation of the chart when wrapping it as a
+# dependency in another chart.
+# This matched the helm best practices:
+# https://helm.sh/docs/chart_best_practices/dependencies/#conditions-and-tags
+# +docs:hidden
+enabled: true
```
```suggestion
enabled: true

```

Nit

deploy/charts/trust-manager/values.linter.exceptions
```diff
@@ -0,0 +1 @@
+value missing from templates: enabled
```
erikgb: Is this exception really needed? Maybe you can explain why in the description if it is?

hawksight:
/hold

Doing some testing in a wrapper chart first.

deploy/charts/trust-manager/values.linter.exceptions
hawksight: This is to satisfy the helm schema validation.
Because the value doesn't get used inside an actual chart template, it flags up this warning, hence why it is added here to skip :)

erikgb:
```suggestion
value missing from templates: enabled

```

Ok, then another nit comment from me.

hawksight: I think this is resolved now?

hawksight:
@erikgb locally, those files appear to have newlines:

Screenshot 2025-08-11 at 10 28 33

Screenshot 2025-08-11 at 10 28 38

They don't appear to on here, or am I just looking at the wrong place?

hawksight:
@erikgb I checked with @SgtCoDFish on standup and those should indeed be resolved. I was having a slow morning.

Also as for testing, I have confirmed that it does indeed work as expected. Here is a Chart.yaml I was using:

```yaml
apiVersion: v2
# note os for open source. No closed source things here.
name: cert-manager-os-wrap
description: Install core open source cert-manager components together
type: application
version: 0.1.0
# Based on the core cert-manager version
appVersion: "1.14.4"
# Potentially many deps
dependencies:
- name: cert-manager
  version: "v1.16.3"
  repository: "https://charts.jetstack.io"
- name: cert-manager-approver-policy
  version: "v0.18.0"
  repository: "https://charts.jetstack.io"
- name: cert-manager-csi-driver
  version: "v0.10.1"
  repository: "https://charts.jetstack.io"
- name: trust-manager
  version: "v0.0.0"
  repository: "file:///Users/peter.fiddes/ghq/github.com/cert-manager/trust-manager/deploy/charts/trust-manager"
  condition: trust-manager.enabled
```

Worked perfectly when not enabled, and also when it was. I'll remove the hold on this.

erikgb (commented):
Thanks @hawksight! 🚀 

/lgtm
/approve

PR #681, opened 2025-08-10
WIP: Add new Bundle (migration) controller

erikgb:
We need a new Bundle controller, as the current Bundle controller will be changed into the new ClusterBundle controller. The migration Bundle controller will:

- If the ClusterBundle isn't managed by a user, the controller will create/update the ClusterBundle from the Bundle, and set an owner-reference back to the Bundle.
- If the ClusterBundle is determined as user-managed, the controller will apply an empty patch to give up ownership over all previously owned fields. Including the owner-reference. From this point, the user is fully managing the ClusterBundle and the obsolete Bundle can be deleted.


erikgb:
> We should document this also when documenting migration steps.

One of the major remaining tasks is to write migration docs and update the trust-manager docs. Hoping to get some help from @hawksight here! 🙏 

I am going to let this merge now, so I can continue on the next PR. Merging this does not yet trigger any user-facing change.

/approve

PR #702, opened 2025-08-25
User-facing migration to ClusterBundle

erikgb:
This PR introduces the new _ClusterBundle_ CRD and enables the migration controller according to the accepted [design](https://github.com/cert-manager/trust-manager/blob/main/design/20241124-rename-bunde-to-clusterbundle.md).

The main controller is changed to reconcile _ClusterBundle_ instead of _Bundle_.

I've tried to keep this PR as small as possible to make it easier to review. We will likely need to add additional tests for the new features in the _ClusterBundle_ API (e.g., multiple keys) and consolidate the existing tests due to the more generic API. However, I suggest handling this in follow-up PRs. Let me know what you think!

PR #703, opened 2025-08-25
Move additional formats handling from source to target

erikgb:
This PR is a preparation for https://github.com/cert-manager/trust-manager/pull/702. As the new API is considerably more flexible, it makes sense to move the handling of "additional formats". In the `ClusterBundle` API, the format and password (in case of a truststore format) can be set per key in the target, so it doesn't make sense to prepare this in sources.

This PR is required before I can continue to migrate the current Bundle controller.

The test coverage is probably slightly reduced in this PR, as I am not testing with a custom password in the target test, just the default passwords. But we have other tests of the binary encoders, and I would prefer not spending extra time on the binary formats.


erikgb:
@inteon @ThatsMrTalbot Could I get a review of this, please? It's in the migration to `ClusterBundle` line.

PR #716, opened 2025-08-30
WIP: fix(deps): updraging k8s deps

hjoshi123:


erikgb:
@inteon, if you have some time, I think we need your help on this one!

hjoshi123:
> @inteon, if you have some time, I think we need your help on this one!

yes I agree.. we could hack the tests and add the required fields to make it happy (for ex the status fields) but we would love to see if the tests need to be rewritten or changed so that this doesnt happen in future.

Issue #726, opened 2025-09-03
Feature: Limit trust-manager to watch only a list of namespaces


asmaoune:
**Summary** 

Add an opt-in mode where trust-manager only watches a specific list of Kubernetes namespaces provided at startup or via Helm values. 

When the list is empty, behavior remains unchanged (watch all namespaces). 

**Proposed change** 

- Introduce a new flag: --watch-namespaces, which accepts a comma-separated list of namespaces that trust-manager should watch. 
- Restrict the controller’s cache to only these namespaces when the list is non-empty. 
- Also restrict the target cache (used for writing/reading targets) to the same allowlist when the feature is enabled, keeping behavior unchanged when the list is empty. 
- Keep cluster-scoped resources (e.g., Bundle) unchanged. 

**RBAC behavior (when the feature is enabled)** 

- Keep ClusterRole permissions only for cluster-scoped resources (e.g., Bundles, Namespaces etc.). 
- Replace cluster-wide write permissions on ConfigMaps with per-namespace Role/RoleBinding in each allowed namespace (granting only the necessary verbs there). 
- If secretTargets.enabled=true, keep the same behavior for secrets, but move write permissions to per-namespace Role/RoleBinding in the allowed namespaces (no cluster-wide writes for secrets). 
- In addition to the per-namespace RBAC for the allowlisted namespaces, a dedicated Role and RoleBinding will be deployed in the trust namespace to grant trust-manager the required permissions (read sources). 
- The helm chart will manage these per-namespace Role/RoleBinding objects based on the provided allowlist and will avoid creating cluster-wide write permissions when this feature is enabled.  The Creation of these per-namespace RBAC objects will be a toggled option. 

**Notes** 

This issue is watch-only. It does not change reconciliation semantics (e.g., target expansion / namespaceSelector). Writes outside the allowlist may simply fail if RBAC is tightened separately (out of scope here). 

By default, when the list is empty, behavior is exactly the same as current releases. 

**Examples** 

watchNamespaces empty → trust-manager watches cluster-wide (current behavior). 

watchNamespaces = ["namespace-1","namespace-2"] → trust-manager will have permissions on namespaced objects (configmaps and secrets if secretsTargets is activated) in namespace-1 and namespace-2 (and continues handling cluster-scoped resources). The target cache is also configured with the same namespace allowlist.

erikgb:
[...]

PR #744, opened 2025-09-10
feat: add ability to limit the target namespaces managed

asmaoune:
Closes: #726

## Summary

This PR adds a new restricted mode to scope trust-manager’s  and target caches to a specific set of Kubernetes namespaces provided at startup. When this feature is not used, behavior remains unchanged (cluster-wide watch).

## What’s in this PR

1. New flag: target-namespaces=ns-1,ns-2,...
2. Helm values:
```yaml
app:
  # Limit namespaces managed by trust-manager
  targetNamespaces: []       # e.g., ["ns-1", "ns-2"]; empty = cluster-wide
```

3. Controller-runtime caches:

- Manager cache uses DefaultNamespaces when app.restricted.targetNamespaces is non-empty
- Manger cache will be configured with namespaces defined in targetNamespaces + the trust-namespce
- Target cache is restricted to targetNamespaces

Reconciler behavior: 

- Pre-filters target namespaces; no attempts to write outside the targetNamespaces (avoids RBAC noise)

## RBAC / Helm behavior

- Cluster-scoped permissions (e.g., Bundles, Namespaces) remain cluster-scoped
- Namespaced writes (ConfigMaps; optional Secrets if enabled) are limited to the targetNamespaces
- The chart creates per-namespace Role/RoleBinding in each target namespace, trust-namespace and trust-manager namespace

## Backward compatibility

- With targetNamespaces: [] (or no flag), behavior is identical to current releases.
- namespaceSelector semantics are unchanged; writes now occur to the intersection of the selector and the targetNamespaces list

## Tests

Unit: add cases for restricted mode

- No namespaceSelector ⇒ writes only to target namespaces; success message updated
- With namespaceSelector including outside namespaces ⇒ writes only to the selected namespaces that are in the targetNamespaces

## How to validate

```bash
# Helm (restricted mode)
helm upgrade --install trust-manager ./deploy/charts/trust-manager \
  -n cert-manager --wait \
  --set app.trust.namespace=cert-manager \
  --set-json app.targetNamespaces='["ns-1","ns-2"]' \
```

### DIFF
--- a/cmd/trust-manager/app/app.go
+++ b/cmd/trust-manager/app/app.go
@@ -89,7 +89,7 @@ func NewCommand() *cobra.Command {
 				Metrics: server.Options{
 					BindAddress: fmt.Sprintf("0.0.0.0:%d", opts.MetricsPort),
 				},
-				Cache: bundle.CacheOpts(opts.Bundle),
+				Cache: bundle.CacheOpts(opts.Bundle, opts.TargetNamespaces),
 			})
 			if err != nil {
 				return fmt.Errorf("failed to create manager: %w", err)
@@ -106,7 +106,7 @@ func NewCommand() *cobra.Command {
 			logf.IntoContext(ctx, log)
 
 			// Add Bundle controller to manager.
-			if err := bundle.SetupWithManager(ctx, mgr, opts.Bundle); err != nil {
+			if err := bundle.SetupWithManager(ctx, mgr, opts.Bundle, opts.TargetNamespaces); err != nil {
 				return fmt.Errorf("failed to register Bundle controller: %w", err)
 			}
 
--- a/cmd/trust-manager/app/app.go
+++ b/cmd/trust-manager/app/app.go
@@ -89,7 +89,7 @@ func NewCommand() *cobra.Command {
 				Metrics: server.Options{
 					BindAddress: fmt.Sprintf("0.0.0.0:%d", opts.MetricsPort),
 				},
-				Cache: bundle.CacheOpts(opts.Bundle, opts.TargetNamespaces),
+				Cache: bundle.CacheOpts(opts.Bundle),
 			})
 			if err != nil {
 				return fmt.Errorf("failed to create manager: %w", err)
@@ -106,7 +106,7 @@ func NewCommand() *cobra.Command {
 			logf.IntoContext(ctx, log)
 
 			// Add Bundle controller to manager.
-			if err := bundle.SetupWithManager(ctx, mgr, opts.Bundle, opts.TargetNamespaces); err != nil {
+			if err := bundle.SetupWithManager(ctx, mgr, opts.Bundle); err != nil {
 				return fmt.Errorf("failed to register Bundle controller: %w", err)
 			}
 
--- a/cmd/trust-manager/app/options/options.go
+++ b/cmd/trust-manager/app/options/options.go
@@ -83,9 +83,6 @@ type Options struct {
 	LeaderElectionConfig LeaderElectionConfig
 
 	TLSConfig TLSConfig
-
-	// Limit both the manager and the target caches to the provided list of namespaces
-	TargetNamespaces []string
 }
 
 type TLSConfig struct {
@@ -237,10 +234,6 @@ func (o *Options) addAppFlags(fs *pflag.FlagSet) {
 	fs.IntVar(&o.MetricsPort,
 		"metrics-port", 9402,
 		"Port to expose Prometheus metrics on 0.0.0.0 on path '/metrics'.")
-
-	fs.StringSliceVar(&o.TargetNamespaces,
-		"target-namespaces", nil,
-		"Comma-separated list of namespaces to limit both the manager and target caches")
 }
 
 func (o *Options) addBundleFlags(fs *pflag.FlagSet) {
@@ -259,6 +252,10 @@ func (o *Options) addBundleFlags(fs *pflag.FlagSet) {
 	fs.BoolVar(&o.Bundle.FilterExpiredCerts,
 		"filter-expired-certificates", false,
 		"Filter expired certificates from the bundle.")
+
+	fs.StringSliceVar(&o.Bundle.TargetNamespaces,
+		"target-namespaces", nil,
+		"Comma-separated list of namespaces to limit both the manager and target caches.")
 }
 
 func (o *Options) addLoggingFlags(fs *pflag.FlagSet) {
--- a/cmd/trust-manager/app/options/options.go
+++ b/cmd/trust-manager/app/options/options.go
@@ -240,7 +240,7 @@ func (o *Options) addAppFlags(fs *pflag.FlagSet) {
 
 	fs.StringSliceVar(&o.TargetNamespaces,
 		"target-namespaces", nil,
-		"Comma-seperated list of namespaces to limit both the manager and target caches")
+		"Comma-separated list of namespaces to limit both the manager and target caches")
 }
 
 func (o *Options) addBundleFlags(fs *pflag.FlagSet) {
--- a/cmd/trust-manager/app/options/options.go
+++ b/cmd/trust-manager/app/options/options.go
@@ -83,6 +83,9 @@ type Options struct {
 	LeaderElectionConfig LeaderElectionConfig
 
 	TLSConfig TLSConfig
+
+	// Limit both the manager and the target caches to the provided list of namespaces
+	TargetNamespaces []string
 }
 
 type TLSConfig struct {
@@ -234,6 +237,10 @@ func (o *Options) addAppFlags(fs *pflag.FlagSet) {
 	fs.IntVar(&o.MetricsPort,
 		"metrics-port", 9402,
 		"Port to expose Prometheus metrics on 0.0.0.0 on path '/metrics'.")
+
+	fs.StringSliceVar(&o.TargetNamespaces,
+		"target-namespaces", nil,
+		"Comma-seperated list of namespaces to limit both the manager and target caches")
 }
 
 func (o *Options) addBundleFlags(fs *pflag.FlagSet) {
--- a/deploy/charts/trust-manager/templates/deployment.yaml
+++ b/deploy/charts/trust-manager/templates/deployment.yaml
@@ -112,6 +112,10 @@ spec:
           {{- if .Values.filterExpiredCertificates.enabled }}
           - "--filter-expired-certificates=true"
           {{- end }}
+            # target-namespaces
+          {{- if .Values.app.restricted.targetNamespaces }}
+          - '--target-namespaces={{ join "," .Values.app.restricted.targetNamespaces }}'
+          {{- end }}
         volumeMounts:
         - mountPath: /tls
           name: tls
--- a/deploy/charts/trust-manager/templates/deployment.yaml
+++ b/deploy/charts/trust-manager/templates/deployment.yaml
@@ -113,8 +113,8 @@ spec:
           - "--filter-expired-certificates=true"
           {{- end }}
             # target-namespaces
-          {{- if .Values.app.restricted.targetNamespaces }}
-          - '--target-namespaces={{ join "," .Values.app.restricted.targetNamespaces }}'
+          {{- if .Values.app.targetNamespaces }}
+          - '--target-namespaces={{ join "," .Values.app.targetNamespaces }}'
           {{- end }}
         volumeMounts:
         - mountPath: /tls
--- a/deploy/charts/trust-manager/templates/deployment.yaml
+++ b/deploy/charts/trust-manager/templates/deployment.yaml
@@ -112,7 +112,6 @@ spec:
           {{- if .Values.filterExpiredCertificates.enabled }}
           - "--filter-expired-certificates=true"
           {{- end }}
-            # target-namespaces
           {{- if .Values.app.targetNamespaces }}
           - '--target-namespaces={{ join "," .Values.app.targetNamespaces }}'
           {{- end }}
--- a/deploy/charts/trust-manager/templates/rbac-per-namespace.yaml
+++ b/deploy/charts/trust-manager/templates/rbac-per-namespace.yaml
@@ -1,5 +1,5 @@
-{{- if and .Values.app.restricted.targetNamespaces .Values.app.restricted.rbacPerNamespace.create }}
-{{- $targetNamespaces := .Values.app.restricted.targetNamespaces -}}
+{{- if .Values.app.targetNamespaces }}
+{{- $targetNamespaces := .Values.app.targetNamespaces -}}
 {{- $allNamespaces := append (append $targetNamespaces .Values.app.trust.namespace) (include "trust-manager.namespace" . | trim) -}}
 {{- range $ns := uniq $allNamespaces }}
 ---
--- a/deploy/charts/trust-manager/templates/rbac-per-namespace.yaml
+++ b/deploy/charts/trust-manager/templates/rbac-per-namespace.yaml
@@ -1,7 +1,6 @@
 {{- if .Values.app.targetNamespaces }}
-{{- $targetNamespaces := .Values.app.targetNamespaces -}}
-{{- $allNamespaces := append (append $targetNamespaces .Values.app.trust.namespace) (include "trust-manager.namespace" . | trim) -}}
-{{- range $ns := uniq $allNamespaces }}
+{{- $namespaces := append (append .Values.app.targetNamespaces .Values.app.trust.namespace) (include "trust-manager.namespace" . ) -}}
+{{- range $ns := uniq $namespaces }}
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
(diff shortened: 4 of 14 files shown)


erikgb:
@asmaoune, can you please fix the DCO on your commits, so I can enable CI and start my review?

/cc

cmd/trust-manager/app/options/options.go
```diff
@@ -83,6 +83,9 @@ type Options struct {
 	LeaderElectionConfig LeaderElectionConfig
 
 	TLSConfig TLSConfig
+
+	// Limit both the manager and the target caches to the provided list of namespaces
+	TargetNamespaces []string
```
erikgb: This new field should be moved to the controller.Options. That will give it more/better context.

https://github.com/cert-manager/trust-manager/blob/5d94b1348f4666bfc933aec7c5d3ac7b893ed38e/pkg/bundle/controller/options.go#L20-L35

cmd/trust-manager/app/app.go
```diff
@@ -106,7 +106,7 @@ func NewCommand() *cobra.Command {
 			logf.IntoContext(ctx, log)
 
 			// Add Bundle controller to manager.
-			if err := bundle.SetupWithManager(ctx, mgr, opts.Bundle); err != nil {
+			if err := bundle.SetupWithManager(ctx, mgr, opts.Bundle, opts.TargetNamespaces); err != nil {
```
erikgb:
```suggestion
			if err := bundle.SetupWithManager(ctx, mgr, opts.Bundle); err != nil {
```

No need for the extra param here after moving the new option.

deploy/charts/trust-manager/values.yaml
```diff
@@ -331,6 +331,15 @@ app:
...
         labels: {}
 
+  # +docs:section=Restricted mode
+  restricted:
+    # list of namespaces that trust-manager can write to.
+    # Empty list = cluster-wide (default behavior).
+    targetNamespaces: []
+    rbacPerNamespace:
+      # If true, the chart creates a Role and RoleBinding in each target namespace.
+      create: false
```
erikgb: Not sure if we should add this option now. Do you need it in your setup? It can easily be added later on. If dropping it now, we have one less Helm value/structure to discuss. 😆 And we don't yet have an option to disable cluster-wide RBAC.

deploy/charts/trust-manager/values.yaml
```diff
@@ -331,6 +331,15 @@ app:
         # Additional labels to add to the ServiceMonitor.
         labels: {}
 
+  # +docs:section=Restricted mode
+  restricted:
```
erikgb: Not sure if I like the `restricted` term. Does it provide any additional value? What's wrong with just `app.targetNamespaces`?

pkg/bundle/bundle.go
```diff
@@ -177,6 +181,7 @@ func (b *bundle) reconcileBundle(ctx context.Context, req ctrl.Request) (statusP
 		return nil, fmt.Errorf("failed to build NamespaceSelector: %w", err)
 	}
 
+	writtenNamespaces := make(map[string]struct{})
```
erikgb: Can't we use a simple slice here? There should be no duplicate namespaces when listing them.

pkg/bundle/bundle.go
```diff
@@ -301,9 +314,23 @@ func (b *bundle) reconcileBundle(ctx context.Context, req ctrl.Request) (statusP
...
+	var message string
+	if len(b.targetNamespaces) > 0 {
+		message = "Successfully synced Bundle to all allowed namespaces"
+		if !namespaceSelector.Empty() {
+			namespaces := make([]string, 0, len(writtenNamespaces))
+			for ns := range writtenNamespaces {
+				namespaces = append(namespaces, ns)
+			}
+			sort.Strings(namespaces)
+			message = fmt.Sprintf("Successfully synced Bundle to namespaces: %s", strings.Join(namespaces, ","))
```
erikgb: Is it useful to list namespaces only in this situation? It's quite inconsistent with the messages used in other setups. That will also remove the need for the `writtenNamespaces ` variable.

```suggestion
			message = fmt.Sprintf("Successfully synced Bundle to allowed namespaces that match this label selector: %s", namespaceSelector)
```

pkg/bundle/bundle.go
```diff
@@ -60,6 +61,9 @@ type bundle struct {
 	bundleBuilder *source.BundleBuilder
 
 	targetReconciler *target.Reconciler
+
+	// target namespaces where to write bundles
+	targetNamespaces map[string]struct{}
```
erikgb: No need to add this new field. We already have a field with options in this struct.

pkg/bundle/controller.go
```diff
@@ -67,25 +67,47 @@ func CacheOpts(opts controller.Options) cache.Options {
...
 	}
+	if len(targetNamespaces) > 0 {
+		nsMap := make(map[string]cache.Config)
+		for _, ns := range targetNamespaces {
+			nsMap[ns] = cache.Config{}
+		}
+		nsMap[opts.Namespace] = cache.Config{}
+		ctrCacheOptions.DefaultNamespaces = nsMap
+	}
+	return ctrCacheOptions
```
erikgb: I think this function will be easier to read if you start by declaring (but not making) a map variable for `DefaultNamespaces`  before the cache options are created.

pkg/bundle/controller.go
```diff
@@ -67,25 +67,47 @@ func CacheOpts(opts controller.Options) cache.Options {
 			},
 		},
 	}
+	if len(targetNamespaces) > 0 {
+		nsMap := make(map[string]cache.Config)
+		for _, ns := range targetNamespaces {
+			nsMap[ns] = cache.Config{}
+		}
+		nsMap[opts.Namespace] = cache.Config{}
```
erikgb: This is the most important namespace, and should be added when the map is initialized above.

pkg/bundle/controller.go
```diff
@@ -67,25 +67,47 @@ func CacheOpts(opts controller.Options) cache.Options {
...
+
+	if len(targetNamespaces) > 0 {
+		nsMap := make(map[string]cache.Config)
+		for _, ns := range targetNamespaces {
+			nsMap[ns] = cache.Config{}
+		}
+		targetCacheOpts.DefaultNamespaces = nsMap
+		logf.FromContext(ctx).Info("restricting target cache to namespaces",
+			"namespaces", targetNamespaces)
+	}
```
erikgb: To avoid duplicating the code from the setup of the main cache, I would suggest a helper function to create the map. If the helper can return the nil map when all namespaces are targets, the code will become a lot more compact and more readable.

pkg/bundle/controller.go
```diff
@@ -95,7 +117,7 @@ func SetupWithManager(
 	}
 
 	// Add Bundle controller to manager.
-	if err := addBundleController(ctx, mgr, opts, targetCache); err != nil {
+	if err := addBundleController(ctx, mgr, opts, targetCache, targetNamespaces); err != nil {
```
erikgb:
```suggestion
	if err := addBundleController(ctx, mgr, opts, targetCache); err != nil {
```

There is already a parameter for the options.

deploy/charts/trust-manager/values.yaml
asmaoune: @erikgb In our setup, the RBAC will be deployed by another operator, that's why we need a toggled option. Also, we will need to add an a toggled option for the cluster-wide RBAC. In our setup we can't deploy RBAC for other namespaces (we are in multi-tenant environment, for security reason we need a toggled option)

deploy/charts/trust-manager/values.yaml
asmaoune: @erikgb Ok, I will use app.targetNamespaces

deploy/charts/trust-manager/values.yaml
erikgb: Ok, then I suggest introducing a general RBAC toggle in a follow-up PR. No need to mix it into this PR IMO.

deploy/charts/trust-manager/templates/deployment.yaml
```diff
@@ -111,6 +111,10 @@ spec:
           {{- end }}
           {{- if .Values.filterExpiredCertificates.enabled }}
           - "--filter-expired-certificates=true"
+          {{- end }}
+            # target-namespaces
```
erikgb:
```suggestion
```

Nit: Not sure if this comment adds anything useful.

pkg/bundle/bundle_test.go
```diff
@@ -1523,6 +1627,16 @@ func Test_Reconcile(t *testing.T) {
...
+
+			if len(tt.targetNamespaces) > 0 {
+				b.Options.TargetNamespaces = make([]string, len(tt.targetNamespaces))
+				for _, ns := range tt.targetNamespaces {
+					if ns == "" {
+						continue
+					}
+					b.Options.TargetNamespaces = append(b.Options.TargetNamespaces, ns)
+				}
+			}
```
erikgb:
```suggestion
			b.Options.TargetNamespaces = tt.targetNamespaces
```

I think this should work. As this code is in a test, we should expect the test to be configured without blanks.

pkg/bundle/controller.go
```diff
@@ -296,3 +312,18 @@ func labelsMatchSelector(objLabels map[string]string, labelSelector *metav1.Labe
 	}
 	return selector.Matches(labels.Set(objLabels))
 }
+
+// setupCacheNamespaces configure cache namespaces
+func setupCacheNamespaces(targetNamespaces []string, trustNamespace string) map[string]cache.Config {
```
erikgb:
```suggestion
func setupCacheNamespaces(namespaces ...string) map[string]cache.Config {
```

I think the code will be simpler if using a variadic param here.

pkg/bundle/bundle_test.go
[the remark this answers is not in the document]
erikgb: I would suggest accepting the suggestion in https://github.com/cert-manager/trust-manager/pull/744#discussion_r2345359733 to resolve this.

pkg/bundle/controller.go
```diff
@@ -46,8 +46,11 @@ import (
 )
 
 func CacheOpts(opts controller.Options) cache.Options {
+	ctrCacheNamespaces := setupCacheNamespaces(append([]string{opts.Namespace}, opts.TargetNamespaces...)...)
```
erikgb:
```suggestion
	ctrCacheNamespaces := setupCacheNamespaces(opts.Namespace, opts.TargetNamespaces...)
```

I think this should work and is simpler to read.

erikgb (commented):
Almost there, thanks!

/label tide/merge-method-squash

pkg/bundle/controller.go
asmaoune: @erikgb at first I used it but it didn't work

erikgb: Hmm, strange. I will take a look.

erikgb: Ok, TIL! Go has some sharp corners. Oh well....

erikgb:
```suggestion
	ctrCacheNamespaces := setupCacheNamespaces(append(opts.TargetNamespaces, opts.Namespace)...)
```

But this works, and is a little easier to read.

deploy/charts/trust-manager/templates/rbac-per-namespace.yaml
```diff
@@ -0,0 +1,51 @@
+{{- if .Values.app.targetNamespaces }}
+{{- $targetNamespaces := .Values.app.targetNamespaces -}}
+{{- $allNamespaces := append (append $targetNamespaces .Values.app.trust.namespace) (include "trust-manager.namespace" . | trim) -}}
+{{- range $ns := uniq $allNamespaces }}
```
erikgb:
```suggestion
{{- $namespaces := append (append .Values.app.targetNamespaces .Values.app.trust.namespace) (include "trust-manager.namespace" . ) -}}
{{- range $ns := uniq $namespaces }}
```

deploy/charts/trust-manager/templates/rbac-per-namespace.yaml
```diff
@@ -0,0 +1,51 @@
...
+    resources: ["secrets"]
+    verbs: ["get","list","watch"]
+  - apiGroups: [""]
+    resources: ["secrets"]
+    verbs: ["create","patch","delete"]
+    resourceNames:
+{{ toYaml $.Values.secretTargets.authorizedSecrets | nindent 6 }}
+    {{- end }}
+  {{- end }}
+
```
erikgb: Would it be an idea to deduplicate these RBAC rules with a Helm helper? If not, I fear that these will be forgotten when modifying the `ClusterRole` rules.

pkg/bundle/controller.go
```diff
@@ -79,12 +83,19 @@ func SetupWithManager(
 		return fmt.Errorf("failed to create target label requirement: %w", err)
 	}
 
+	targetCacheNamespaces := setupCacheNamespaces(opts.TargetNamespaces...)
+	if targetCacheNamespaces != nil {
+		logf.FromContext(ctx).Info("restricting target cache to namespaces",
+			"namespaces", opts.TargetNamespaces)
+	}
```
erikgb:
```suggestion
	if len(targetCacheNamespaces) > 0 {
		logf.FromContext(ctx).Info("restricting target cache to namespaces",
			"namespaces", maps.Keys(targetCacheNamespaces))
	}
```

asmaoune: @erikgb using maps.keys will display the memory adress

erikgb: Yes, you are correct. My bad. For someone viewing this, `maps.Keys` doesn't work the way I thought it would in Go, as referenced. https://go.dev/play/p/S2kn96mQ9qE

deploy/charts/trust-manager/templates/rbac-per-namespace.yaml
```diff
@@ -0,0 +1,28 @@
...
+{{- $namespaces := append (append .Values.app.targetNamespaces .Values.app.trust.namespace) (include "trust-manager.namespace" . ) -}}
+{{- range $ns := uniq $namespaces }}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: trust-manager-role
+  namespace: {{ $ns }}
+rules:
+{{ include "trust-manager.rbac.namespacedResourcesRules" $ | nindent 2 }}
```
erikgb:
```suggestion
{{ include "trust-manager.rbac.namespacedResourcesRules" $ }}
```

I am not an expert on Helm, but why is the include indented here, and not in the `ClusterRole`?

deploy/charts/trust-manager/values.yaml
```diff
@@ -331,6 +331,10 @@ app:
         # Additional labels to add to the ServiceMonitor.
         labels: {}
 
+  # List of target namespaces that trust-manager can write to.
+  # Empty list = cluster-wide (default behavior).
+  targetNamespaces: []
```
```suggestion
    # List of target namespaces that trust-manager can write to. By default, trust-manager can write targets in any namespace.
  # +docs:property
  # targetNamespaces: ["ns-1", "ns-2"]
```

This section is misplaced in the metrics section. I tried to push a fix to your branch, but it seems like you haven't allowed maintainers to guide you. 😒

> ERROR: Permission to orange-cloudfoundry/trust-manager.git denied to erikgb.

I suggest moving it right after the `app.trust.namespace` value.

pkg/bundle/controller.go
```diff
@@ -79,12 +83,19 @@ func SetupWithManager(
 		return fmt.Errorf("failed to create target label requirement: %w", err)
 	}
 
+	targetCacheNamespaces := setupCacheNamespaces(opts.TargetNamespaces...)
```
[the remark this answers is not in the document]
erikgb: Maybe this is a good idea. WDYT, @asmaoune?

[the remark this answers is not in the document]
asmaoune: @erikgb  Yes it is a good idea to add a comment

deploy/charts/trust-manager/templates/rbac-per-namespace.yaml
asmaoune: @erikgb In the existing ClusterRole file, no indentation was used in the rules section. To keep it consistent, I added the namespaced resource rules without indentation (I didn’t want to change the structure of the existing file). However, in the rbac-per-namespace file that I created, I used indentation in the rules section — that’s why the include is indented.

erikgb: I suggest using it consistently, without indentation. Or is there a reason you want to indent the array? Non-indented arrays are used by most of the Kubernetes ecosystem, as it makes the YAML more "compact".

asmaoune: @erikgb There is no reason. I did the changes

deploy/charts/trust-manager/values.yaml
```diff
@@ -239,6 +239,11 @@ app:
     # before installing trust-manager.
     namespace: cert-manager
 
+  # List of target namespaces that trust-manager can write to. By default, trust-manager can write targets in any namespace.
+  # +docs:property
+  # targetNamespaces: ["ns-1", "ns-2"]
+  targetNamespaces: []
```
```suggestion
```

There should be no default value. The default/common behavior is documented above. An empty array default can be very confusing.

erikgb (changes requested):
One more thing to fix, which I also commented in my last review. When this is fixed, I think we are ready to merge.

/approve

StenGruener:
hi all, 

looking on these lines here

```
- apiGroups:
  - ""
  resources:
  - "secrets"
  verbs: ["get","list","watch"]
````

What is the particular need to "get" all secrets from all namespaces even in case authorizedSecrets are applied

PR #762, opened 2025-10-02
Add support for injecting CA from secret for trust manager Webhook

zeusal:
Add support for injecting CA from Secret, allowing cainjector to run with --enable-certificates-data-source=false as recommended by best practices. https://cert-manager.io/docs/concepts/ca-injector/#controlling-which-data-sources-are-enabled


erikgb:
@zeusal, did you also read the note in https://cert-manager.io/docs/installation/best-practice/#memory?

image

zeusal:
> @zeusal, did you also read the note in https://cert-manager.io/docs/installation/best-practice/#memory?

Yes, I did read the note in the best practices guide. The MR is aligned with it: by allowing inject-ca-from-secret, we can keep running cainjector with --enable-certificates-data-source=false, so the CA is injected from a pre-generated secret instead of generating certificates in the pod. This follows the memory and security recommendations from the guide.

The MR doesn’t change the default behavior — it only adds an optional way to provide the CA from a secret, keeping the pod lightweight and compliant with the best practices.

With this change, the chart can natively support both behaviors. It allows avoiding certificate auto-discovery and additional watchers, while still letting the MutationWebhook obtain the CA from a secret by specifying ca-from-secret and setting cert-manager.io/allow-direct-injection: "true" on that secret.

deploy/charts/trust-manager/values.yaml
```diff
@@ -299,6 +299,10 @@ app:
         # The name of cert-manager's Service Account. Only used if `app.webhook.tls.approverPolicy.enabled` is true.
         certManagerServiceAccount: "cert-manager"
 
+      # Added option to inject CA from Secret instead of Certificate.
+      # This allows running cainjector with --enable-certificates-data-source=false, which follows cert-manager best practices by disabling unused data sources.
+      injectCaFrom: "certificate"
```
erikgb: I am not a big fan of this "enum", but your use case makes a lot of sense. Are you able to see a way to avoid this? Could it make sense to unconditionally use the `cert-manager.io/inject-ca-from-secret` annotation on the webhook?

zeusal: First of all, sorry for the delay....Would the intention be to switch the default behaviour to inject-ca-from-secret and drop Certificate-based injection entirely? I kept both options mainly to preserve the existing Certificate-based behaviour while allowing users to run cainjector with --enable-certificates-data-source=false.

zeusal:
```suggestion
      injectCaFromSecret: false
```

Based on your feedback, I simplified this by replacing the `injectCaFrom` enum with a boolean `injectCaFromSecret` option. This keeps both behaviours available while avoiding the enum:

* `false`: use `cert-manager.io/inject-ca-from` (current/default behaviour)
* `true`: use `cert-manager.io/inject-ca-from-secret`

This also allows users to run cainjector with `--enable-certificates-data-source=false`.

erikgb: > Would the intention be to switch the default behaviour to inject-ca-from-secret and drop Certificate-based injection entirely?

Yes, mainly to not complicate the chart even more. We set the name of the secret in https://github.com/cert-manager/trust-manager/blob/main/deploy/charts/trust-manager/templates/certificate.yaml#L34, so this should work the same either way. Please add a comment about why we are doing this to the template.

zeusal: Okay, I'll make the changes, but does the “Secret” need to have “cert-manager.io/allow-direct-injection” set to “true”?  https://controller.cert-manager.io/docs/concepts/ca-injector/?utm_source=chatgpt.com

zeusal:
Duplicate MR, final MR is #1078 with the specified changes writed here.

PR #771, opened 2025-10-09
Don't set the tag in values.yaml, since it is overwritten at chart build time

inteon:
If we keep this unused value in `values.yaml`, it confuses users and developers.

### DIFF
--- a/deploy/charts/trust-manager/values.yaml
+++ b/deploy/charts/trust-manager/values.yaml
@@ -102,9 +102,9 @@ defaultPackageImage:
   repository: quay.io/jetstack/trust-pkg-debian-bookworm
 
   # Override the image tag of the default package image.
-  # If no value is set, the chart's appVersion is used.
+  # Is set at chart build time to the version specified in ./make/00_debian_bookworm_version.mk.
   # +docs:property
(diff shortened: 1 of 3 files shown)


deploy/charts/trust-manager/values.yaml
```diff
@@ -102,9 +102,9 @@ defaultPackageImage:
   repository: quay.io/jetstack/trust-pkg-debian-bookworm
 
   # Override the image tag of the default package image.
-  # If no value is set, the chart's appVersion is used.
+  # Is set at chart build time to the version specified in ./make/00_debian_bookworm_version.mk.
```
erikgb, line 105: Isn't these docs user-facing? Maybe we should avoid references to makefiles in the source repo? Maybe something in the lines of "latest release of the default package image"?

inteon: True, not sure how we could make this clear otherwise.
How will someone looking at the values.yaml file understand where this tag value comes from?

inteon:
I think it is clear in the doc that this does not require an action from their end. Might be a bit dirty to add these developer comments in the values.yaml file, but is the best solution I found.
/unhold

PR #768, opened 2025-10-08
Bump default CAs bundle version to trigger release

erikgb:
As shown in https://artifacthub.io/packages/helm/cert-manager/trust-manager?modal=security-report, which I have also verified locally (see below), we have vulnerabilities in our default CA images. This PR bumps the default CA images which I "think" should trigger a new release of these images automatically.

````
Report Summary

┌──────────────────────────────────────────────────────────────────────────────────┬──────────┬─────────────────┬─────────┐
│                                      Target                                      │   Type   │ Vulnerabilities │ Secrets │
├──────────────────────────────────────────────────────────────────────────────────┼──────────┼─────────────────┼─────────┤
│ quay.io/jetstack/trust-pkg-debian-bookworm:20230311-deb12u1.0@sha256:4e46f31af8- │  alpine  │        0        │    -    │
│ ab4aedb861b997be37e2004ba5313a187d0f0a3cdb4680937a98a3 (alpine 3.21)             │          │                 │         │
├──────────────────────────────────────────────────────────────────────────────────┼──────────┼─────────────────┼─────────┤
│ ko-app/debian-bundle-static                                                      │ gobinary │        2        │    -    │
└──────────────────────────────────────────────────────────────────────────────────┴──────────┴─────────────────┴─────────┘
Legend:
- '-': Not scanned
- '0': Clean (no security findings detected)


ko-app/debian-bundle-static (gobinary)

Total: 2 (UNKNOWN: 0, LOW: 0, MEDIUM: 1, HIGH: 1, CRITICAL: 0)

┌─────────┬────────────────┬──────────┬────────┬───────────────────┬─────────────────┬─────────────────────────────────────────────────────────────┐
│ Library │ Vulnerability  │ Severity │ Status │ Installed Version │  Fixed Version  │                            Title                            │
├─────────┼────────────────┼──────────┼────────┼───────────────────┼─────────────────┼─────────────────────────────────────────────────────────────┤
│ stdlib  │ CVE-2025-47907 │ HIGH     │ fixed  │ v1.24.5           │ 1.23.12, 1.24.6 │ database/sql: Postgres Scan Race Condition                  │
│         │                │          │        │                   │                 │ https://avd.aquasec.com/nvd/cve-2025-47907                  │
│         ├────────────────┼──────────┤        │                   │                 ├─────────────────────────────────────────────────────────────┤
│         │ CVE-2025-47906 │ MEDIUM   │        │                   │                 │ os/exec: Unexpected paths returned from LookPath in os/exec │
[...]

PR #776, opened 2025-10-16
Bump trust package suffix, forcing a new go 1.25.3 build

inteon:
Since 1.25.2 had some unexpected breaking changes, I think we should bump the sufix again to build using go 1.25.3 instead of go 1.25.2.

(see https://github.com/cert-manager/trust-manager/pull/768 for last suffix bump)


maelvls:
I've looked at this PR, I'd be OK

PR #816, opened 2025-12-12
Release improvements

SgtCoDFish:
Adds a way to do a pre-release scan, and workflows which scan trust packages for issues.

Also improves the release docs to hopefully be clearer.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,10 +1,33 @@
 # Releases
 
-## Schedule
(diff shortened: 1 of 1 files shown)


SgtCoDFish:
First run of the new workflow failed (as expected): https://github.com/cert-manager/trust-manager/actions/runs/20171788939

Now to create new releases of the trust package images!

PR #817, opened 2025-12-12
Bump trust package versions to address CVE-2025-61729

SgtCoDFish:
No reason to expect that this is exploitable but we'd like to patch anyway

I manually confirmed that the upstream versions listed (`20230311+deb12u1` and `20210119`) are indeed the latest available versions.


SgtCoDFish:
/approve

(adding manual LGTM label since I think Richard forgot!)

PR #844, opened 2026-01-18
Fix conversion Bundle->ClusterBundle

erikgb:
This change is extracted from https://github.com/cert-manager/trust-manager/pull/702 to make that PR smaller and easier to review.

While working on the user-facing migration, I struggled to get some of the tests to pass. After investigation, I detected some bugs that need to be fixed:

1. We allow multiple source elements with in-line CA certificates in the `Bundle` sources array, while `ClusterBundle` only provides a single field for this. This mismatch (many-to-one) is a bit problematic; I considered several ways to fix this, but ended up with the simplest solution: concatenate all in-line source CA certificates into the `ClusterBundle` `inLineCAs` field when converting. This breaks round-trip conversion, but we only need conversion from `Bundle` to `ClusterBundle`.
2. One of our integration tests is asserting that we can produce the legacy JKS format. And this test breaks if using the PCKS#12 encoder to encode the JKS. To ensure we don't break any users, it is required to keep the legacy JKS encoder until we remove JKS support completely. To choose the correct encoder to use, we need an annotation to support the migrated bundle controller in https://github.com/cert-manager/trust-manager/pull/702

### DIFF
--- a/pkg/apis/trust/v1alpha1/conversion.go
+++ b/pkg/apis/trust/v1alpha1/conversion.go
@@ -17,7 +17,9 @@ limitations under the License.
 package v1alpha1
 
 import (
+	"fmt"
 	"slices"
+	"strings"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	apimachineryconversion "k8s.io/apimachinery/pkg/conversion"
@@ -27,10 +29,19 @@ import (
 	trustv1alpha2 "github.com/cert-manager/trust-manager/pkg/apis/trustmanager/v1alpha2"
 )
 
-const annotationKeyJKSKey = "internal.trust-manager.io/jks-key"
+// AnnotationKeyJKSKey is the controller-internal annotation key used to indicate the key (if any)
+// in the ClusterBundle target specification that should be encoded with the legacy JKS encoder.
+// The JKS additional format is deprecated in the Bundle API, but we don't want to break users
+// when upgrading to a trust-manager to a release with support for ClusterBundle.
+// Support for this annotation using the legacy encoder will be removed together with the removal
+// of the v1alpha1 Bundle API.
+const AnnotationKeyJKSKey = "internal.trust-manager.io/jks-key"
 
 func (src *Bundle) ConvertTo(dstRaw conversion.Hub) error {
-	dst := dstRaw.(*trustv1alpha2.ClusterBundle)
+	dst, ok := dstRaw.(*trustv1alpha2.ClusterBundle)
+	if !ok {
+		return fmt.Errorf("expected a ClusterBundle, but got a %T", dstRaw)
+	}
 	dst.ObjectMeta = src.ObjectMeta
 
 	scheme := runtime.NewScheme()
@@ -53,6 +64,9 @@ func (src *Bundle) ConvertTo(dstRaw conversion.Hub) error {
 	dst.Spec.Sources = slices.DeleteFunc(dst.Spec.Sources, func(bs trustv1alpha2.BundleSource) bool {
 		return bs == trustv1alpha2.BundleSource{}
 	})
+	if len(dst.Spec.Sources) == 0 {
+		dst.Spec.Sources = nil
+	}
 
 	return nil
 }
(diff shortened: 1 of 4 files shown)


pkg/bundle/controller/bundle_controller.go
```diff
@@ -144,9 +146,10 @@ func convertBundleToClusterBundle(bundle *trustapi.Bundle) (*trustmanagerapi.Clu
...
 	clusterBundle := &trustmanagerapi.ClusterBundle{}
-	clusterBundle.APIVersion = "trust-manager.io/v1alpha2"
-	clusterBundle.Kind = "ClusterBundle"
-	clusterBundle.Name = bundle.Name
+	clusterBundle.Name = cb.Name
+	if jksKey, ok := cb.Annotations[trustapi.AnnotationKeyJKSKey]; ok {
+		clusterBundle.Annotations = map[string]string{trustapi.AnnotationKeyJKSKey: jksKey}
+	}
 	clusterBundle.Spec = cb.Spec
 	return clusterBundle, nil
```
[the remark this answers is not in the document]
erikgb: Well, that's not entirely true, Copilot. I am NOT copying ALL fields, and I don't want most of the metadata converted from `Bundle`.

pkg/apis/trust/v1alpha1/conversion.go
```diff
@@ -27,7 +28,7 @@ import (
 	trustv1alpha2 "github.com/cert-manager/trust-manager/pkg/apis/trustmanager/v1alpha2"
 )
 
-const annotationKeyJKSKey = "internal.trust-manager.io/jks-key"
+const AnnotationKeyJKSKey = "internal.trust-manager.io/jks-key"
```
SgtCoDFish, line 38: nitpick: If this is to be made public, I think it would benefit from having a doc comment explaining what it is and when it would be used (or not used)

pkg/apis/trust/v1alpha1/conversion.go
```diff
@@ -27,7 +28,7 @@ import (
...
 
 func (src *Bundle) ConvertTo(dstRaw conversion.Hub) error {
 	dst := dstRaw.(*trustv1alpha2.ClusterBundle)
```
SgtCoDFish: suggestion (unrelated to this PR): This is a little spooky - if the hub type were to ever change (which presumably it would for v1) this would just silently blow up in a surprising way, right? It kinda requires us to either get rid of v1alpha1 first or else to remember this footgun.

Should we add a check for this cast (doesn't have to be this PR)?

```suggestion
	dst, ok := dstRaw.(*trustv1alpha2.ClusterBundle)
	if !ok { ... }
```

pkg/apis/trust/v1alpha1/conversion.go
```diff
@@ -78,7 +82,16 @@ func Convert_v1alpha1_BundleSource_To_v1alpha2_BundleSource(in *BundleSource, ou
...
-		obj.Spec.InLineCAs = in.InLine
+		if obj.Spec.InLineCAs == nil {
+			obj.Spec.InLineCAs = in.InLine
+		} else {
+			// The following logic is not pretty, but is required as we allow multiple inline sources
+			// in the Bundle sources array.
+			// It breaks the round-trippable conversion Bundle->ClusterBundle->Bundle,
+			// but works for converting Bundle->ClusterBundle, and that's what we need for the migration.
+			cas := strings.TrimSuffix(*obj.Spec.InLineCAs, "\n") + "\n" + *in.InLine
+			obj.Spec.InLineCAs = &cas
```
SgtCoDFish, line 103: note: I'm not massively against breaking strict round-trippability here (this should probably be a pretty rare case). Plus, I can't really see a way for this to lead to actual data loss, just a change of format. This seems OK to me!

pkg/apis/trust/v1alpha1/conversion_test.go
```diff
@@ -21,17 +21,52 @@ import (
...
 	utilconversion "github.com/cert-manager/trust-manager/pkg/util/conversion"
+	"github.com/cert-manager/trust-manager/test/dummy"
 )
 
+func TestBundle_ConvertTo(t *testing.T) {
+	tests := map[string]struct {
+		src Bundle
+		exp trustmanagerapi.ClusterBundle
+	}{
+		"multiple inline Bundle sources should be concatenated in ClusterBundle inLineCAs": {
```
SgtCoDFish, line 42: suggestion: This is a good test - could we go a step further and test the round trip back to v1alpha1, just so that the behaviour we're specifying here is testable in the future?

pkg/apis/trust/v1alpha1/conversion.go
erikgb: Makes sense! I've now added docs to the constant.

pkg/apis/trust/v1alpha1/conversion.go
erikgb: Nice catch! Not strictly related to this change, but I've now fixed both unguarded casts in this file.

pkg/apis/trust/v1alpha1/conversion_test.go
erikgb: I have now extended the test with an additional assert of the conversion spoke->hub->spoke.

PR #855, opened 2026-01-24
Eliminate use of naked bool (includeDefaultCAs) in ClusterBundle API

erikgb:
Somehow related to https://github.com/cert-manager/trust-manager/pull/850, naked booleans are discouraged in Kubernetes APIs because they make the API constrained and hard to extend without breaking changes. Ref. [Kube API conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types):

> Think twice about bool fields. Many ideas start as boolean but eventually trend towards a small set of mutually exclusive options. Plan for future expansions by describing the policy options explicitly as a string type alias (e.g. TerminationMessagePolicy).

In this PR, I suggest eliminating the only remaining naked boolean in the new `ClusterBundle` API: `includeDefaultCAs`, into something I feel is better aligned with the API conventions.

Note that this is a breaking change, but ClusterBundle is not yet user-facing, ref. https://github.com/cert-manager/trust-manager/pull/702.


pkg/apis/trustmanager/v1alpha2/types_cluster_bundle.go
```diff
@@ -100,6 +93,28 @@ type BundleSource struct {
...
+	//		container.
+	//
+	//		If no default CA package was configured at startup, specifying this source
+	//		will result in reconciliation failure.
+	//
+	//		The version of the default CA package used for a Bundle is reported in
+	//		status.defaultCAPackageVersion.
+	// - Disabled: No default CAs are used as sources.
+	// +required
+	// +kubebuilder:validation:Enum=System;Disabled
```
SgtCoDFish, line 119: question: (sorry, I've not got time to check right now) Is this check case sensitive? I think the UX isn't great if "system" is rejected and I'd very much like this to be case insensitive!

erikgb: It is case sensitive, so the value `system` would not be accepted. And making it case-insensitive would also be a violation of the [API conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#constants) IMO.

SgtCoDFish: Ah, thank you for linking to the API conventions. If it's standardised there, I'm much less bothered by it and I'll have to stop being lazy and use CamelCase!

PR #865, opened 2026-01-31
Fix events RBAC (new API group)

erikgb:
This PR fixes an oversight in https://github.com/cert-manager/trust-manager/pull/863. When migrating to the new events API group, we also need to update RBAC. This is listed as a breaking change in the controller-runtime release notes: https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.23.0

### DIFF
--- a/deploy/charts/trust-manager/templates/_helpers.tpl
+++ b/deploy/charts/trust-manager/templates/_helpers.tpl
@@ -77,7 +77,7 @@ Namespaced resources rules
   verbs: ["get","list","create","patch","watch","delete"]
 
 - apiGroups:
-  - ""
+  - "events.k8s.io"
   resources:
   - "events"
   verbs: ["create","patch"]


deploy/charts/trust-manager/templates/_helpers.tpl
```diff
@@ -77,7 +77,7 @@ Namespaced resources rules
   verbs: ["get","list","create","patch","watch","delete"]
 
 - apiGroups:
-  - ""
+  - "events.k8s.io"
```
wallrj-cyberark, line 80: @erikgb I *think* the RBAC needs support both the new and the old events API, for the leader election package.
See 
 * https://github.com/kubernetes/kubernetes/issues/82846
 * https://github.com/kubernetes-sigs/controller-runtime/issues/3355

You can see that the kubebuilder project, for example, has RBAC roles for the old and the new events APIs
 * Old for leaderelection: https://github.com/kubernetes-sigs/kubebuilder/blob/master/docs/book/src/getting-started/testdata/project/config/rbac/leader_election_role.yaml#L34-L40
 * New for controller events: https://github.com/kubernetes-sigs/kubebuilder/blob/master/docs/book/src/getting-started/testdata/project/config/rbac/role.yaml#L53-L59

erikgb: You seem to be right here. Fix: https://github.com/cert-manager/trust-manager/pull/889

PR #870, opened 2026-02-06
Helm Chart: Add support for setting relabelling on the ServiceMonitor

tiesmaster:
As discussed [here](https://kubernetes.slack.com/archives/C4NV3DWUC/p1770300885888059), it's convenient to be able to configure metrics relabelling on the deployed ServiceMonitor, similarly like with [cert-manager](https://github.com/cert-manager/cert-manager/blob/26d38b0bc40c9d9a6d01298b500dd972372f35a7/deploy/charts/cert-manager/values.yaml#L670-L682).

This adds the same property in the `values.yaml` as in the helm chart of cert-manager, and wires it up appropriately, and updates the schema, and documentation. I've tried to keep it as close to cert-manager as possible.

If there is anything I've missed, let me know, and I'll update the PR.

### DIFF
--- a/deploy/charts/trust-manager/README.md
+++ b/deploy/charts/trust-manager/README.md
@@ -614,6 +614,27 @@ The timeout for a metrics scrape.
 > ```
 
 Additional labels to add to the ServiceMonitor.
+#### **app.metrics.service.servicemonitor.endpointAdditionalProperties** ~ `object`
+> Default value:
+> ```yaml
+> {}
+> ```
+
+EndpointAdditionalProperties allows setting additional properties on the endpoint such as relabelings, metricRelabelings etc.  
+  
+For example:
+
+```yaml
+endpointAdditionalProperties:
+ relabelings:
+ - action: replace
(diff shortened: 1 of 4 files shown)


test-values.yaml
hawksight, line 1: I don't think we need this file. It's useful to see you have tested it, but adding to the repo as is is too generic.
We don't have helm tests, so we can probably drop this file from the PR. You could include it in the description as a test case should others require it.

tiesmaster: Agreed, I will remove the file from the PR. TBH, I used it for testing purposes, but I forgot to exclude it from the commit. I will remove it in my next update.

tiesmaster:
@hawksight Thanks for the review. I've ran the command, and it updated the `README.md` of the helm chart. I've added that to the initial commit (to keep the history clean, I hope that's ok), and pushed a new update. Please have another look.

erikgb (commented):
/approve

Thanks @tiesmaster 🚀

PR #871, opened 2026-02-06
Introduce KAL optional fields checks

erikgb:
In this PR, I enable the [OptionalFields](https://github.com/kubernetes-sigs/kube-api-linter/blob/main/docs/linters.md#optionalfields), which is one of the linters enabled by default in KAL.

After some discussions with @inteon, I decided to take a step back and reduce the size of the initial PR. Mainly to make it more reviewable, but also since we are in doubt about some of the failing checks. So now I have added a few lint suppressions that we could eventually fix in a follow-up PR.


pkg/apis/trust/v1alpha1/types_bundle.go
```diff
@@ -220,7 +221,7 @@ type SourceObjectKeySelector struct {
 	// includeAllKeys is a flag to include all keys in the object's `data` field to be used. False by default.
 	// This field must not be true when `key` is set.
 	// +optional
-	IncludeAllKeys bool `json:"includeAllKeys,omitempty"`
+	IncludeAllKeys *bool `json:"includeAllKeys,omitempty"`
```
inteon, line 234: Actually @erikgb, the combination `{includeAllKeys: false, key: test}` does seem valid to me and should be allowed as far as I know.

erikgb: Yes, when we change the field to a pointer.

inteon: Sorry, for asking again; why does it have to be a pointer?
Both `{includeAllKeys: false, key: test}` and `{key: test}` are valid and
both `{includeAllKeys: false}` and `{}` are invalid.
So, we don't really have to know whether `includeAllKeys` was provided explicitly, right?

erikgb: One of the goals of the linter is to make the serialized form consistent and clear. Without a pointer, it is impossible to distinguish the following serialized forms from each other:

````json
{}
````

````json
{"includeAllKeys": false}
````

inteon: Ok, so we don't need to be able to tell if `includeAllKeys` is set for validation.

erikgb:
I finally made CI pass. Some changes I like, others I dislike. Not sure what to do. There are also some test changes I prefer to take out to a separate PR - depending on how we move forward with this. WDYT, @inteon?

erikgb:
@inteon, any chance you could take another look at this? Especially if we should disable some of the KAL checks.

Issue #868, opened 2026-02-04
trust-pkg-debian-bookworm image contains Go stdlib vulnerabilities


tdolliver-veza:
# Summary

The `trust-pkg-debian-bookworm` image has Go stdlib vulnerabilities that can be resolved by rebuilding with the current build toolchain, which already uses Go 1.25.6.

## Affected Image

quay.io/jetstack/trust-pkg-debian-bookworm:20230311-deb12u1.3

## Vulnerabilities

| CVE | Severity | Description |
|-----|----------|-------------|
| CVE-2025-61726 | HIGH | net/url: Memory exhaustion in query parameter parsing |
| CVE-2025-61728 | HIGH | archive/zip: Excessive CPU consumption when building archive index |
| CVE-2025-61730 | MEDIUM | TLS 1.3 handshake issue |

All three are fixed in Go 1.25.6.

## Trivy Scan Output

```
Report Summary

┌─────────────────────────────────────────────────────────────────────────────┬──────────┬─────────────────┬─────────┐
│                                   Target                                    │   Type   │ Vulnerabilities │ Secrets │
├─────────────────────────────────────────────────────────────────────────────┼──────────┼─────────────────┼─────────┤
│ quay.io/jetstack/trust-pkg-debian-bookworm:20230311-deb12u1.3 (alpine 3.21) │  alpine  │        0        │    -    │
├─────────────────────────────────────────────────────────────────────────────┼──────────┼─────────────────┼─────────┤
│ ko-app/debian-bundle-static                                                 │ gobinary │        3        │    -    │
└─────────────────────────────────────────────────────────────────────────────┴──────────┴─────────────────┴─────────┘
Legend:
- '-': Not scanned
- '0': Clean (no security findings detected)


ko-app/debian-bundle-static (gobinary)

Total: 3 (UNKNOWN: 0, LOW: 0, MEDIUM: 1, HIGH: 2, CRITICAL: 0)

┌─────────┬────────────────┬──────────┬────────┬───────────────────┬─────────────────┬──────────────────────────────────────────────────────────────┐
│ Library │ Vulnerability  │ Severity │ Status │ Installed Version │  Fixed Version  │                            Title                             │
├─────────┼────────────────┼──────────┼────────┼───────────────────┼─────────────────┼──────────────────────────────────────────────────────────────┤
[...]

PR #875, opened 2026-02-07
Trigger a new build of default trust bundle images

erikgb:
Similar to https://github.com/cert-manager/trust-manager/pull/817. I am echoing what @SgtCoDFish wrote in that PR:

> No reason to expect that this is exploitable but we'd like to patch anyway

> I manually confirmed that the upstream versions listed (20230311+deb12u1 and 20210119) are indeed the latest available versions.

Fixes https://github.com/cert-manager/trust-manager/issues/868

### DIFF
--- a/make/00_debian_bookworm_version.mk
+++ b/make/00_debian_bookworm_version.mk
@@ -17,5 +17,5 @@
 # This file is used to store the latest version of the debian trust package and the DEBIAN_BOOKWORM_BUNDLE_VERSION
 # variable is automatically updated by the `upgrade-debian-bookworm-trust-package-version` target and cron GH action.
 
-DEBIAN_BOOKWORM_BUNDLE_VERSION := 20230311+deb12u1.3
+DEBIAN_BOOKWORM_BUNDLE_VERSION := 20230311+deb12u1.4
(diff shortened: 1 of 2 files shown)


make/00_debian_bookworm_version.mk
```diff
@@ -17,5 +17,5 @@
 # This file is used to store the latest version of the debian trust package and the DEBIAN_BOOKWORM_BUNDLE_VERSION
 # variable is automatically updated by the `upgrade-debian-bookworm-trust-package-version` target and cron GH action.
 
-DEBIAN_BOOKWORM_BUNDLE_VERSION := 20230311+deb12u1.3
+DEBIAN_BOOKWORM_BUNDLE_VERSION := 20230311+deb12u1.4
```
inteon, line 20: general comment: this tag version seems wrong to me, now the .4 is included in the build metadata instead of the actual version number.
This means (as far as I understand) that, when sorting, this version will be equivalent to `20230311+deb12u1.3`.

erikgb: The upstream bundle version is still `20230311+deb12u1`. We just need something extra to patch our own stuff.

PR #895, opened 2026-02-25
Allow specifying a custom type for Secret targets

mrak:
This pull request adds a new field `type` for secret targets.

Example `Bundle`

```yaml
kind: Bundle
metadata:
  name: corp-ca
spec:
  sources:
  - secret:
      key: tls.crt
      name: corp-certs
  target:
    namespaceSelector:
      matchExpressions:
      - key: bundleTarget
        operator: In
        values:
        - corp
    secret:
      key: ca.crt
      type: nginx.org/ca
```

Generated `Secret`

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: corp-ca
  namespace: my-namespace
type: nginx.org/ca
data:
  ca.crt: ++++++++
```

An example use case would be for those using the NGINX Inc (F5) ingress controller which requires secrets of type `nginx.org/ca` with a `ca.crt` key for use in mTLS `Policy` resources.

Since the `type` field only makes sense for secret targets, but the code is shared between configmap and secret targets, some effort was made to add validation and testing to ensure that `type` is not an accepted field for configmap targets.


mrak:
Thanks @erikgb! I can look into this. One question I had was around the apiverion conversions. Would my changes need to modify the v1alpha1 and v1alpha2 versions? Or do I need to make a `v1alpha3` to account for the added `type` field for the secret?

erikgb:
You need to update both Bundle (v1alpha1) and ClusterBundle (v1alpha2). The latter is still not user-facing, but we are working on it.

mrak:
Apologies for the larger PR. I needed to touch many things when I replaced the KeyValueTarget and the TargetTemplate with specialized versions 😅

mrak:
This is ready for another review!

pkg/bundle/internal/target/target.go
```diff
@@ -397,7 +401,14 @@ type Resource struct {
 	types.NamespacedName
 }
 
-func TrustBundleHash(data []byte, additionalFormats *trustapi.AdditionalFormats, target *trustapi.TargetTemplate) string {
+// TargetSpec is an interface that both ConfigMapTarget and SecretTarget satisfy.
+type TargetSpec interface {
+	GetAnnotations() map[string]string
+	GetLabels() map[string]string
+	GetSecretType() corev1.SecretType
```
erikgb, line 408: I am not particularly happy with this method, as it is only relevant for secret targets. It's only used for the hash calculation, if I am not mistaken? Can we find another way?

The secret `type` is an immutable property, so we might want to handle changes to this property in the bundle spec a bit differently, as changing it will require the secret to be recreated.

mrak:
Sorry, I haven't had time to return back to this. We've moved forward with a different pattern internally to create the same Secret resources with a custom type.

I'll close this PR in the meantime. If I return to it or others pick it up it can be reopened!

PR #918, opened 2026-03-29
add bundle metrics

### LINKED ISSUE #588: Add ability to monitor validity period for CAs in bundle
I would like a way to monitor the validity period of CA's that are distributed managed by trust-manager Bundles. This would help provide visibility if there are CAs in the bundle that are close to expiring that may need to be updated, or removed from the bundle.

My initial thought is this could be achieved via custom Prometheus metrics that provides the Not Before and Not After dates for each CA in the bundle. If we choose this route, @erikgb recommended using metrics similar to `cert_not_before`  and `cert_not_after` (inspired by https://github.com/ribbybibby/ssl_exporter).

Another option could be to have trust-manager update the `.status` block of each `Bundle` CR with the dates for the CA , and then users could use that metadata however they need to in their environments. For example, they could pull that into Prometheus using kube-state-metrics using [Custom Resource State Metrics](https://github.com/kubernetes/kube-state-metrics/blob/main/docs/metrics/extend/customresourcestate-metrics.md).

/kind feature

Peac36:
Fixes: https://github.com/cert-manager/trust-manager/issues/588

### Changes
- New metrics: certmanager_ssl_ca_not_before and certmanager_ssl_ca_not_after expose certificate validity dates as Unix timestamps
- Source tracking: Certificates are now tracked with their source (ConfigMap, Secret, inline, or default CA package) for better observability
- Prometheus integration: Metrics are automatically scraped when the trust-manager is deployed with Prometheus

### DIFF
--- a/cmd/trust-manager/app/app.go
+++ b/cmd/trust-manager/app/app.go
@@ -28,12 +28,16 @@ import (
 	"k8s.io/klog/v2"
 	ctrl "sigs.k8s.io/controller-runtime"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
 	"sigs.k8s.io/controller-runtime/pkg/metrics/server"
 	ctrlwebhook "sigs.k8s.io/controller-runtime/pkg/webhook"
 
 	"github.com/cert-manager/trust-manager/cmd/trust-manager/app/options"
 	trustapi "github.com/cert-manager/trust-manager/pkg/apis/trust/v1alpha1"
 	"github.com/cert-manager/trust-manager/pkg/bundle"
+	bundleMetrics "github.com/cert-manager/trust-manager/pkg/bundle/metrics"
+	"github.com/cert-manager/trust-manager/pkg/fspkg"
+	pkgMetrics "github.com/cert-manager/trust-manager/pkg/metrics"
 	"github.com/cert-manager/trust-manager/pkg/webhook"
 )
 
@@ -95,6 +99,23 @@ func NewCommand() *cobra.Command {
 				return fmt.Errorf("failed to create manager: %w", err)
 			}
 
+			var pkg *fspkg.Package
+			if opts.Bundle.DefaultPackageLocation != "" {
+				data, err := fspkg.LoadPackageFromFile(opts.Bundle.DefaultPackageLocation)
+				if err != nil {
+					return fmt.Errorf("must load default package successfully when default package location is set: %w", err)
+				}
+				pkg = &data
+				log.Info("successfully loaded default package from filesystem", "id", pkg.StringID(), "path", opts.Bundle.DefaultPackageLocation)
+			}
+
+			m := pkgMetrics.NewRegistrator(metrics.Registry)
+			m.Add(bundleMetrics.NewBundleCollector(log, opts.Bundle, mgr.GetClient(), pkg))
+
+			if err := mgr.Add(m); err != nil {
+				return fmt.Errorf("failed to add metrics: %w", err)
+			}
+
 			if err := mgr.AddReadyzCheck("webhook", mgr.GetWebhookServer().StartedChecker()); err != nil {
 				return fmt.Errorf("failed to add webhook ready check: %v", err)
 			}
@@ -106,7 +127,7 @@ func NewCommand() *cobra.Command {
 			logf.IntoContext(ctx, log)
 
 			// Add Bundle controller to manager.
-			if err := bundle.SetupWithManager(ctx, mgr, opts.Bundle); err != nil {
+			if err := bundle.SetupWithManager(ctx, mgr, opts.Bundle, pkg); err != nil {
 				return fmt.Errorf("failed to register Bundle controller: %w", err)
 			}
 
--- a/pkg/bundle/internal/source/source.go
+++ b/pkg/bundle/internal/source/source.go
@@ -18,6 +18,8 @@ package source
 
 import (
 	"context"
+	"crypto/sha256"
+	"crypto/x509"
 	"fmt"
 
 	corev1 "k8s.io/api/core/v1"
@@ -38,6 +40,20 @@ type NotFoundError struct{ error }
 type InvalidPEMError struct{ error }
 
 type InvalidSecretError struct{ error }
+type Kind string
+
+const (
+	ConfigMapKind Kind = "ConfigMap"
+	SecretKind    Kind = "Secret"
+	InlineKind    Kind = "InLine"
+	DefaultCAKind Kind = "DefaultCAs"
+)
+
+type CertMetadata struct {
+	Kind Kind
+	Name string
+	Key  string
+}
 
 // BundleData holds the result of a call to BuildBundle. It contains the resulting PEM-encoded
 // certificate data from concatenating all the sources together, binary data for any additional formats and
@@ -46,6 +62,41 @@ type BundleData struct {
 	CertPool *util.CertPool
 
 	DefaultCAPackageStringID string
+
+	meta map[[32]byte]CertMetadata
+}
+
+func NewBundleData(pool *util.CertPool) BundleData {
+	return BundleData{
+		CertPool: pool,
+		meta:     make(map[[32]byte]CertMetadata),
+	}
+}
+
+func (b *BundleData) AddFromPem(pemData []byte, kind Kind, name, key string) error {
+	err := b.CertPool.ParsePemData(pemData, func(cert *x509.Certificate) {
+		if b.CertPool.AddCert(cert) {
+			hash := sha256.Sum256(cert.Raw)
+			if kind == DefaultCAKind || kind == InlineKind {
+				name = fmt.Sprintf("%x", hash)
+			}
+			b.meta[hash] = CertMetadata{
+				Kind: kind,
+				Name: name,
+				Key:  key,
+			}
+		}
+	})
+	return err
+}
+
+func (b *BundleData) GetMetadata(cert *x509.Certificate) (CertMetadata, error) {
+	hash := sha256.Sum256(cert.Raw)
+	meta, ok := b.meta[hash]
+	if !ok {
+		return CertMetadata{}, fmt.Errorf("no metadata found for certificate with hash %x", hash[:])
+	}
+	return meta, nil
 }
 
 type BundleBuilder struct {
@@ -61,12 +112,12 @@ type BundleBuilder struct {
 // BuildBundle retrieves and concatenates all source bundle data for this Bundle object.
 // Each source data is validated and pruned to ensure that all certificates within are valid.
 func (b *BundleBuilder) BuildBundle(ctx context.Context, sources []trustapi.BundleSource) (BundleData, error) {
-	var resolvedBundle BundleData
-	resolvedBundle.CertPool = util.NewCertPool(
+	certPool := util.NewCertPool(
 		util.WithFilteredExpiredCerts(b.FilterExpiredCerts),
 		util.WithFilteredNonCaCerts(b.FilterNonCACerts),
 		util.WithLogger(logf.FromContext(ctx).WithName("cert-pool")),
 	)
+	resolvedBundle := NewBundleData(certPool)
 
 	for _, source := range sources {
 		var certSource bundleSource
@@ -94,7 +145,7 @@ func (b *BundleBuilder) BuildBundle(ctx context.Context, sources []trustapi.Bund
 			panic(fmt.Sprintf("don't know how to process source: %+v", source))
 		}
 
-		if err := certSource.addToCertPool(ctx, resolvedBundle.CertPool); err != nil {
+		if err := certSource.addToBundle(ctx, &resolvedBundle); err != nil {
 			return BundleData{}, err
 		}
 	}
@@ -108,15 +159,15 @@ func (b *BundleBuilder) BuildBundle(ctx context.Context, sources []trustapi.Bund
 }
 
 type bundleSource interface {
-	addToCertPool(context.Context, *util.CertPool) error
+	addToBundle(context.Context, *BundleData) error
 }
 
 type inlineBundleSource struct {
 	pemData string
 }
 
-func (s inlineBundleSource) addToCertPool(_ context.Context, pool *util.CertPool) error {
-	if err := pool.AddCertsFromPEM([]byte(s.pemData)); err != nil {
+func (s inlineBundleSource) addToBundle(_ context.Context, bundle *BundleData) error {
+	if err := bundle.AddFromPem([]byte(s.pemData), InlineKind, "", ""); err != nil {
 		return InvalidPEMError{fmt.Errorf("inline source contains invalid PEM data: %w", err)}
 	}
 	return nil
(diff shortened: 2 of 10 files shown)


erikgb (commented):
@Peac36, thanks for picking this up! What do you think about starting without the "source" metrics labels, as suggested in https://github.com/cert-manager/trust-manager/issues/588#issuecomment-2833399661? As I expected, getting those label values is rippling through various parts of the code. We need to refactor to obtain the required information cleanly. WDYT?

Peac36:
Hi @erikgb - I'm not sure that I understand you. The labels are registered without the `source` prefix. In the code, the variables that are used to store these values are prefixed, but this is for clarity.

erikgb:
> Hi @erikgb - I'm not sure that I understand you. The labels are registered without the `source` prefix. In the code, the variables that are used to store these values are prefixed, but this is for clarity.

@Peac36, my issue with the proposed changes is that they modify code I don't think should be changed by adding certificate metrics. As an example, the `CertPool` struct is now changed to track the source of a cert in the pool. This struct was supposed to represent a mutable counterpart to the CertPool in std Go. I just think we have to think a bit more about if, and eventually how, we want the source metrics labels to be. There could potentially be multiple sources for the same certificate in a trust-manager bundle. I propose to start simple, adding certificate metrics for the certificates in the bundle.

Peac36:
I thought you were speaking about the metric labels .. my mistake. I agree with you that those changes extend the cert pool interface, making it a bit unclear. My initial approach was to introduce an alternative container, similar to `CertPool`, that handles metric logic. However, the biggest challenge I faced was that the `CertPool` is not a simple container but has additional logic integrated into it - [deduplication](https://github.com/cert-manager/trust-manager/blob/051e2d884f29ae4d9b65359871d01fdad461116b/pkg/util/cert_pool.go#L159) ,  [filtering](https://github.com/cert-manager/trust-manager/blob/051e2d884f29ae4d9b65359871d01fdad461116b/pkg/util/cert_pool.go#L148-L155) and [parsing](https://github.com/cert-manager/trust-manager/blob/051e2d884f29ae4d9b65359871d01fdad461116b/pkg/util/cert_pool.go#L123). 
While working on implementing the alternative structure, I realized that either I need to extract all this logic from cert pool or simply clone it into the new structure - both scenarios have the same flaw. It's easy to introduce a change only in one struct and forget to add to another. That's why I chose this way in the end.

As next step, I can create a proposal of how to implement this  - wdyt? I'm open for ideas.

Peac36:
@erikgb  - I just pushed the new variant of the code. I believe this is what you meant in your last comment. The main changes are that sources do not interact with the cert pool but with the bundle. The cert pool now has a new method `ParsePemData` that handle parsing of the cert data from the pem data and delegates the storing to a callback. 

Overall, these changes create much cleaner and straightforward way to work with cert and meta(labels). Thanks for the guidance.  If you don't see any issue with this approach, I'll continue testing and notify you once I'm ready.

cmd/trust-manager/app/app.go
```diff
@@ -95,6 +98,18 @@ func NewCommand() *cobra.Command {
...
 
+			var pkg *fspkg.Package
+			if opts.Bundle.DefaultPackageLocation != "" {
+				data, err := fspkg.LoadPackageFromFile(opts.Bundle.DefaultPackageLocation)
+				if err != nil {
+					return fmt.Errorf("must load default package successfully when default package location is set: %w", err)
+				}
+				pkg = &data
+				log.Info("successfully loaded default package from filesystem", "id", pkg.StringID(), "path", opts.Bundle.DefaultPackageLocation)
+			}
```
erikgb, line 110: Was it required to move this piece of code?

erikgb (commented):
@Peac36, this looks a lot better! I didn't have time to fully review the PR, but I like this approach better! 🚀 I will do a better review later this week.

You should probably take a look at https://github.com/cert-manager/approver-policy/pull/863. We were debugging some issues in approver-policy last week.

/ok-to-test

cmd/trust-manager/app/app.go
Peac36: Yes, I wanted to avoid loading the default package twice - once in the controller and once in the metrics.

 >You should probably take a look at https://github.com/cert-manager/approver-policy/pull/863. We were debugging some issues in approver-policy last week.

Thanks, I'll take a look. Meanwhile, I've addressed some of the issues raised by the cooperator

pkg/bundle/metrics/BundleCollector.go
```diff
@@ -0,0 +1,138 @@
...
+		certificates := bundleData.CertPool.Certificates()
+		metrics := make([]prometheus.Metric, 0, len(certificates)*2)
+
+		for j := range certificates {
+			s := certificates[j]
+			meta, err := bundleData.GetMetadata(s)
+			if err != nil {
+				bc.logger.Error(err, "failed to get metadata")
+				continue
+			}
```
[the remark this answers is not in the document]
Peac36: Agree, I'll set the verbosity level to 5 to avoid spamming

pkg/bundle/internal/source/source.go
```diff
@@ -108,15 +156,15 @@ func (b *BundleBuilder) BuildBundle(ctx context.Context, sources []trustapi.Bund
 }
 
 type bundleSource interface {
-	addToCertPool(context.Context, *util.CertPool) error
+	addToBundle(context.Context, BundleData) error
 }
```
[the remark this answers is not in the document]
Peac36: Done

pkg/bundle/metrics/bundlecollector.go
```diff
@@ -0,0 +1,138 @@
...
+		builder := source.BundleBuilder{
+			Reader:         bc.client,
+			Options:        bc.opts,
+			DefaultPackage: bc.pkg,
+		}
+		bundleData, err := builder.BuildBundle(ctx, bundle.Spec.Sources)
+		if err != nil {
+			bc.logger.Error(err, "failed to build bundle")
+			continue
+		}
```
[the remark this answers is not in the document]
Peac36: @erikgb  - what do you think about this? I agree that caching is necessary here, but I don't think the metric and reconciler should use the shared cache. It makes sense to cache the metrics only and not interfere with the logic in the reconciler

pkg/util/cert_pool.go
```diff
@@ -99,6 +99,12 @@ func NewCertPool(options ...Option) *CertPool {
...
 func (cp *CertPool) AddCertsFromPEM(pemData []byte) error {
+	return cp.ParsePemData(pemData, func(cert *x509.Certificate) {
+		cp.AddCert(cert)
+	})
+}
+
+func (cp *CertPool) ParsePemData(pemData []byte, processCert func(cert *x509.Certificate)) error {
 	if pemData == nil {
 		return fmt.Errorf("certificate data can't be nil")
 	}
```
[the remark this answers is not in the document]
Peac36: Disagree with one. Without meaningful processCert, the whole ParsePemData functionality is pointless. So panicking here seems reasonable

Peac36:
@erikgb  - I've addressed most of the Copilot issues. I'll need your opinion on one of the issues raised. 

>You should probably take a look at https://github.com/cert-manager/approver-policy/pull/863. We were debugging some issues in approver-policy last week.

This is also fixed.

pkg/bundle/metrics/bundlecollector_test.go
```diff
@@ -0,0 +1,393 @@
...
+}
+
+func getDescName(desc *prometheus.Desc) string {
+	descStr := desc.String()
+	parts := strings.Split(descStr, "\"")
+	if len(parts) >= 2 {
+		return strings.TrimSuffix(parts[1], "\"")
+	}
+	return ""
+}
```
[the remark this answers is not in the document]
Peac36: Don't get this one

pkg/bundle/metrics/bundlecollector.go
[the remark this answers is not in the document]
SgtCoDFish: I think this is a really important thing to get right and likely this needs to change.

> Consider updating these metrics from the existing Bundle reconciler (or maintaining a cached snapshot with bounded refresh/TTL) so scrapes are O(1) and don’t perform expensive rebuild work.

I think this is the way to go. The existing reconciler can publish metrics as it does its tasks, and as long as publishing metrics cannot fail and break the reconcile I think that's reasonable and good.

pkg/util/cert_pool.go
```diff
@@ -99,6 +99,12 @@ func NewCertPool(options ...Option) *CertPool {
 //
 // Additionally, if the input PEM bundle contains no non-expired certificates, an error is returned.
 func (cp *CertPool) AddCertsFromPEM(pemData []byte) error {
+	return cp.ParsePemData(pemData, func(cert *x509.Certificate) {
+		cp.AddCert(cert)
+	})
+}
+
+func (cp *CertPool) ParsePemData(pemData []byte, processCert func(cert *x509.Certificate)) error {
```
SgtCoDFish, line 125: nitpick: PEM should be all caps when appearing in function names

```suggestion
func (cp *CertPool) ParsePEMData(pemData []byte, processCert func(cert *x509.Certificate)) error {
```

pkg/bundle/internal/source/source.go
```diff
@@ -46,6 +62,41 @@ type BundleData struct {
...
+}
+
+func NewBundleData(pool *util.CertPool) BundleData {
+	return BundleData{
+		CertPool: pool,
+		meta:     make(map[[32]byte]CertMetadata),
+	}
+}
+
+func (b *BundleData) AddFromPem(pemData []byte, kind Kind, name, key string) error {
```
SgtCoDFish, line 76: nitpick: 
```suggestion
func (b *BundleData) AddFromPEM(pemData []byte, kind Kind, name, key string) error {
```

pkg/bundle/metrics/bundlecollector.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+func (bc BundleCollector) Describe(desc chan<- *prometheus.Desc) {
+	desc <- bc.caSSLNotBeforeMetric
+	desc <- bc.caSSLNotAfterMetric
+}
+
+func (bc BundleCollector) Collect(ch chan<- prometheus.Metric) {
+	bundles := v1alpha1.BundleList{}
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
```
SgtCoDFish, line 66: nitpick: is there any way to pass a context into this function to avoid context.Background()?

pkg/bundle/metrics/bundlecollector.go
```diff
@@ -0,0 +1,138 @@
...
+
+	"github.com/cert-manager/trust-manager/pkg/apis/trust/v1alpha1"
+	"github.com/cert-manager/trust-manager/pkg/bundle/controller"
+	"github.com/cert-manager/trust-manager/pkg/bundle/internal/source"
+	"github.com/cert-manager/trust-manager/pkg/fspkg"
+)
+
+var (
+	caSSLNotBefore = prometheus.NewDesc("certmanager_ssl_ca_not_before", "A Unix timestamp of the date when the CA validity begins", []string{"serial_number", "subject_common_name", "bundle_name", "kind", "name", "key"}, nil)
+	caSSLNotAfter  = prometheus.NewDesc("certmanager_ssl_ca_not_after", "A Unix timestamp of the date when the CA validity ends", []string{"serial_number", "subject_common_name", "bundle_name", "kind", "name", "key"}, nil)
```
SgtCoDFish, line 36: ai review comment: What do you think to this?

Potential Prometheus Cardinality Explosion ⚠️

  Each certificate creates 2 metric series with labels including serial_number (unique per cert):

  caSSLNotBefore = prometheus.NewDesc("certmanager_ssl_ca_not_before", ...,
      []string{"serial_number", "subject_common_name", "bundle_name", "kind", "name", "key"}, nil)

  Problem: With many bundles and certificates:
  - 100 bundles × 50 certs each = 10,000 metric series
  - Each cert has a unique serial number = high cardinality
  - Could cause Prometheus performance issues or out-of-memory errors

  Recommendations:
  - Document the expected cardinality and potential Prometheus impact
  - Consider adding configuration to limit which bundles are monitored
  - Optionally reduce labels (e.g., make serial_number optional via config)
  - Add alerts for metric cardinality

pkg/bundle/metrics/bundlecollector.go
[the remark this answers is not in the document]
Peac36: My motive to be against the bundle reconciler pushing metric is that it will require a shared data container between the reconciler and the metric process, which will need synchronization to avoid any potential race conditions. My goal was to implement the metrics in a way that wouldn't affect in any way the main purpose of the controller.

pkg/bundle/metrics/bundlecollector.go
erikgb: I agree to this. Why do we want unique (or almost unique) values as labels? I suggest removing at least the serial number and subject CN. But labels like name and key are also potential cardinality explosions.

pkg/bundle/metrics/bundlecollector.go
erikgb: Doesn't seem so, ref. https://github.com/prometheus/client_golang/issues/1538.

pkg/bundle/metrics/bundlecollector.go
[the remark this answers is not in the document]
erikgb: I agree with the problem Copilot raised, and it can be solved by producing metrics as part of the Bundle reconciliation. But to land on a good solution, we should first decide on the goal for the metrics, including labels, ref. the "cardinality explosion" discussion above. A collector is very useful when the metrics can be constructed from API resources, but it might not be the case here, as I first thought.  Maybe a more traditional approach of using a Prom registry is better here?

pkg/bundle/metrics/bundlecollector.go
Peac36: I based those on what was discussed and agreed on in the issue - https://github.com/cert-manager/trust-manager/issues/588#issuecomment-2831280721

I'm wondering if removing serial_number and subject_common_name will be enough. The system has no limit on the number of bundle sources. Given that each source can contain multiple PEMs, and each PEM can contain multiple certificates, we could still end up with a significant memory load.

>I would like a way to monitor the validity period of CA's that are distributed managed by trust-manager Bundles. This would help provide visibility if there are CAs in the bundle that are close to expiring that may need to be updated, or removed from the bundle.

Based on the issue description, the goal of these metrics is to provide observability over the certificates in a bundle that are closest to expiry. Instead of exposing the expiry timestamps for all certificates in a bundle, we could expose only the data for the certificate that expires soonest, using bundle_name, kind, name, and key as labels. 

Another idea worth considering is exposing the count of certificates that will expire in the next 30 days per bundle. Giving the operator an overall picture of what needs to be processed in the next few days. 

What do you think? 

---

Peac36: @SgtCoDFish , @erikgb  - what do you think about my comment above?

Peac36: @SgtCoDFish , @erikgb  - kind reminder about this task. I'll have some free time in the next few days. So if you can answer me, I'll proceed with the necessary rework.

PR #946, opened 2026-04-23
feat: add keepCertHistory for automatic CA cert retention on rotation

### LINKED ISSUE #560: Support rotated certificate sources
Right now, there is a manual step involved when rotating certificates, of copying the certs. I think this is a feature gap, especially given that the sister project, cert-manager, does rotation automatically, and so there's not much more that needs to be done to implement completely automatic certificate management in a Kubernetes cluster.

What I'm thinking is allowing users to configure something like this in their `Bundle` configuration:

```yaml
  sources:
  - secret:
      name: "some-cert-manager-issued-ca-cert"
      key: "tls.crt"
      keepCertHistory: true
      certHistoryLimit: 3
```

When trust-manager sees `keepCertHistory: true`, it will copy the cert itself into either a list of certs in the status of the `Bundle` CRD, or into a new CRD that exists for the purpose of storing historical certs. And any time it notices that the cert has changed, it will copy the new cert into the list (or create a new instance of the historical cert CRD), keeping the old cert around. The trust bundle will then contain both the current cert, and all previous certs that the operator has seen. The operator can clean old certs up, primarily this would be done based on the expiry date in the cert itself, but to prevent resource leaks (eg, in the case of misconfiguration of rotation), `certHistoryLimit` will specify the maximum number of old certs to keep, with certs being removed in an oldest first fashion if that number is ever reached. This should have a sensible non infinite default.

So, this will automate the copy step described in the docs, and provide a completely automated means for root CA rotation using cert-manager and trust-manager together.

rajatvig:
Closes https://github.com/cert-manager/trust-manager/issues/560

Adds `keepCertHistory` and `certHistoryLimit` fields to Bundle source selectors. When enabled, trust-manager retains previous CA certificates in the trust bundle after rotation until they expire or the history limit is reached, enabling seamless CA rotation.

Changes:
- New API fields on `SourceObjectKeySelector` (v1alpha1 + v1alpha2) and `SourceCertHistory` status type
- History logic: fingerprint-based rotation detection, expired cert pruning, limit enforcement
- Webhook validation for `certHistoryLimit` and `includeAllKeys` incompatibility
- Unit, integration, and envtest tests

```release-note
Added `keepCertHistory` option for Bundle sources. When enabled, trust-manager retains previous CA certificates in the trust bundle after rotation until they expire or a configurable `certHistoryLimit` is reached.
```


bernardgut:
This is actually a requirement for seamless cert rotation. Hope it gets merged soon

houshengbo:
Can any maintainer provide reviews and comments to this PR? Thank you.

rajatvig:
I plan to work on the automated feedback this week.

erikgb:
I think this large change needs a design to be considered. There are probably corner cases to consider.

Issue #908, opened 2026-03-10
v0.22.0: image template does not work when installing trust-manager and cert-manager from one umbrella chart


venc0r:
I install cert-manager and trust-manager together from an umbrella chart:

```
dependencies:
- name: cert-manager
  version: v1.19.4
  repository: https://charts.jetstack.io
- name: trust-manager
  version: v0.22.0
  repository: https://charts.jetstack.io
```

The `image`-template of trust manager changed with v0.22.0 and differs from the cert-manager but share the name. Cert-manager `image`-template wins and this results in a broken images:
```
❯ helm template --release-name cm . | grep 'image: '
          image: "quay.io/jetstack/cert-manager-cainjector:v1.19.4"
          image: "quay.io/jetstack/cert-manager-controller:v1.19.4"
          image: "quay.io/jetstack/cert-manager-webhook:v1.19.4"
        image: ":quay.io"
        image: ":quay.io"
          image: "quay.io/jetstack/cert-manager-startupapicheck:v1.19.4"
```

there are 2 workarrounds for this, use cert-manager v1.20.0-beta.0:
```
❯ helm template --release-name cm . | grep 'image: '
          image: "quay.io/jetstack/cert-manager-cainjector:v1.20.0-beta.0"
          image: "quay.io/jetstack/cert-manager-controller:v1.20.0-beta.0"
          image: "quay.io/jetstack/cert-manager-webhook:v1.20.0-beta.0"
        image: "quay.io/jetstack/trust-pkg-debian-bookworm:20230311-deb12u1.5"
        image: "quay.io/jetstack/trust-manager:v0.22.0"
          image: "quay.io/jetstack/cert-manager-startupapicheck:v1.20.0-beta.0"
```

or overwrite the image:
```
...
trust-manager:
  image:
    repository: quay.io/jetstack/trust-manager
    tag: v0.22.0
  defaultPackageImage:
    repository: quay.io/jetstack/trust-pkg-debian-bookworm
    tag: 20230311-deb12u1.5
```

```
❯ helm template --release-name cm . | grep 'image: '
          image: "quay.io/jetstack/cert-manager-cainjector:v1.19.4"
          image: "quay.io/jetstack/cert-manager-controller:v1.19.4"
          image: "quay.io/jetstack/cert-manager-webhook:v1.19.4"
        image: "quay.io/jetstack/trust-pkg-debian-bookworm:20230311-deb12u1.5"
[...]

PR #950, opened 2026-04-27
fix: rename image helper to avoid umbrella chart conflicts

FelixPhipps:
Part of a group of PRs aiming to fix the issues reported here:

https://github.com/cert-manager/google-cas-issuer/issues/487

https://github.com/cert-manager/trust-manager/issues/908


deploy/charts/trust-manager/templates/_helpers.tpl
```diff
@@ -35,66 +35,12 @@ IMPORTANT: This function is standardized across all charts in the cert-manager G
 Any changes to this function should also be made in cert-manager, trust-manager, approver-policy, ...
 See https://github.com/cert-manager/cert-manager/issues/6329 for a list of linked PRs.
 */}}
-{{- define "image" -}}
```
inteon, line 38: This change does not seem correct to me, this is much more than just a rename.

inteon (approved):

PR #961, opened 2026-05-04
Add duration for the webhook tls certificate

### LINKED ISSUE #955: Feature Request: Add option to set duration for the webhook certificate
Hi,

Would be possible to add an option that allows to set the duration of the webhook certificate generated by cert-manager?

Thank you.

mvrk69:
Add option to set the duration of the Certificate created by cert-manager for trust-manager webhook

Closes https://github.com/cert-manager/trust-manager/issues/955

### DIFF
--- a/deploy/charts/trust-manager/templates/certificate.yaml
+++ b/deploy/charts/trust-manager/templates/certificate.yaml
@@ -39,6 +39,7 @@ spec:
   privateKey:
     rotationPolicy: Always
   revisionHistoryLimit: 1
+  duration: {{ .Values.app.webhook.tls.certificate.duration }}
   issuerRef:
     name: {{ include "trust-manager.name" . }}
     kind: Issuer
--- a/deploy/charts/trust-manager/templates/certificate.yaml
+++ b/deploy/charts/trust-manager/templates/certificate.yaml
@@ -39,7 +39,9 @@ spec:
   privateKey:
     rotationPolicy: Always
   revisionHistoryLimit: 1
+  {{- if .Values.app.webhook.tls.certificate.duration }}
   duration: {{ .Values.app.webhook.tls.certificate.duration }}
+  {{- end }}
   issuerRef:
     name: {{ include "trust-manager.name" . }}
     kind: Issuer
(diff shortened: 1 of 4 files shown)


deploy/charts/trust-manager/values.yaml
```diff
@@ -354,8 +354,9 @@ app:
         # The name of cert-manager's Service Account. Only used if `app.webhook.tls.approverPolicy.enabled` is true.
         certManagerServiceAccount: "cert-manager"
 
-      # Add labels/annotations to secrets created by Certificate resources when using cert-manager provisioned TLS certificate.
+      # Configure certificate duration and add labels/annotations to secrets created by Certificate resources when using cert-manager provisioned TLS certificate.
       certificate:
+        duration: 2160h
```
erikgb: I don't think we should have a default here, but rely on the cert-manager default. Which currently states

> If unset, this defaults to 90 days

In the [API docs](https://cert-manager.io/docs/reference/api-docs/#cert-manager.io/v1.Certificate).

deploy/charts/trust-manager/templates/certificate.yaml
```diff
@@ -39,6 +39,7 @@ spec:
   privateKey:
     rotationPolicy: Always
   revisionHistoryLimit: 1
+  duration: {{ .Values.app.webhook.tls.certificate.duration }}
```
erikgb, line 43: Please make this output conditionally, ref. the no default comment

erikgb (commented):
@mvrk69, this request makes sense to me. I just did an initial review and added a couple of high-level comments.

deploy/charts/trust-manager/templates/certificate.yaml
mvrk69: Done.

deploy/charts/trust-manager/values.yaml
mvrk69: Sounds good. Thank you for the review.
Change done.

erikgb (commented):
/lgtm
/approve

Thanks!

Issue #971, opened 2026-05-15
Sync continiously re-writes ConfigMap


8de2fdb0:
Using helm Chart version  `0.22.0`.

Adding `target.metadata.labels` has the effect of rewriting the configmap every few seconds.

```yaml
apiVersion: trust.cert-manager.io/v1alpha1
kind: Bundle
metadata:
  creationTimestamp: "2026-04-18T07:48:12Z"
  generation: 1
  name: cluster-internal-ca.crt
  resourceVersion: "39542935"
  uid: 1b77b85b-95ec-4d68-814a-dcf589ad92a3
spec:
  sources:
  - secret:
      key: ca.crt
      name: talos-staging0-cluster-local-pki-key
  target:
    configMap:
      key: ca.crt
      metadata:
        labels:
          cluster-layer: foundation-platform-init
          infrastructure: "true"
          managed-by: talos-staging0-cluster
status:
  conditions:
  - lastTransitionTime: "2026-04-18T07:48:19Z"
    message: Successfully synced Bundle to all namespaces
    observedGeneration: 1
    reason: Synced
    status: "True"
    type: Synced
```

As soon i removed those labels it stopped.

Is that expected? There was no other controller touching these ConfigMaps.

erikgb:
@8de2fdb0, thanks for reporting this issue. The behavior you are observing is NOT expected. I can take a look, but would you mind trying our latest release (0.22.1) and supplying any other relevant information?

erikgb:
@8de2fdb0, I am unable to reproduce your issue. I just opened up a PR to add tests asserting no reconcile loops: https://github.com/cert-manager/trust-manager/pull/972. This makes me believe there is a unique problem in your cluster. Please try to provide us with as much relevant information as you can!

nhavens:
I think we may be hitting this in production with trust-manager v0.22.1.

We added two target metadata labels to a Bundle target:

```yaml
target:
  configMap:
    key: default-trust-bundle.pem
    metadata:
      labels:
        app.kubernetes.io/component: trust-manager
        metadata.kube.drw/managed_by: up
```

After rollout, trust-manager repeatedly reconciled and patched the generated ConfigMaps. In two clusters we saw hundreds of `Successfully synced Bundle...` events in short bursts, followed by etcd `mvcc: database space exceeded` patch failures.

[...]

PR #972, opened 2026-05-17
Add tests for stable resourceVersion when adding labels/annotations to target

erikgb:
This work was done when investigating https://github.com/cert-manager/trust-manager/issues/971. The tests show that it is not an obvious issue with our reconciler, but it might be worth keeping the tests.


inteon (commented):
/approve
/lgtm

PR #980, opened 2026-05-26
fix(chart): honor serviceAccount.name override in ServiceAccount manifest

lexfrei:
## What this PR does

Fixes a chart bug where `serviceaccount.yaml` hard-codes `metadata.name` to `trust-manager.name`, while the Deployment template binds via the `trust-manager.serviceAccountName` helper. The helper honors `.Values.serviceAccount.name`, the SA template does not — so any user that sets a custom `serviceAccount.name` with `serviceAccount.create: true` ends up with a name mismatch.

### Reproduction

```yaml
# values.yaml
serviceAccount:
  create: true
  name: my-trust-sa
```

After `helm install`:

- `ServiceAccount/trust-manager` is created (from the chart-name default).
- `Deployment/trust-manager` references `serviceAccountName: my-trust-sa`.
- Pods stay `Pending` on `ServiceAccount "my-trust-sa" not found`.

### Fix

Switch `serviceaccount.yaml:6` from `{{ include "trust-manager.name" . }}` to `{{ include "trust-manager.serviceAccountName" . }}` so the SA manifest and the Deployment binding resolve through the same helper.

### Test

Adds the first helm-unittest coverage under `deploy/charts/trust-manager/tests/` (the `verify-helm-unittest` make target was already wired up in `make/_shared/helm/helm.mk`, it just had no test files yet). Three scenarios:

1. `create=true`, no name override — SA name and Deployment SA reference both default to `trust-manager`.
2. `create=true`, `name=my-sa` — both resolve to `my-sa`. Would fail without the fix.
3. `create=false`, `name=external-sa` — SA manifest suppressed, Deployment binds to `external-sa`.

Verified scenario 2 fails on unfixed `main` with `Expected: my-sa / Actual: trust-manager` and passes after the fix.

### Provenance

Surfaced in a [cozystack/cozystack#2744](https://github.com/cozystack/cozystack/pull/2744) review of a downstream vendoring of `trust-manager` v0.22.1. Downstream consumers that override `serviceAccount.name` (for example, for namespace-scoped naming policies) hit this. Installs without the override are not affected because both sides fall back to the chart-name default.

### DIFF
--- a/deploy/charts/trust-manager/templates/serviceaccount.yaml
+++ b/deploy/charts/trust-manager/templates/serviceaccount.yaml
@@ -3,7 +3,7 @@ apiVersion: v1
 kind: ServiceAccount
 automountServiceAccountToken: {{ .Values.serviceAccount.automountServiceAccountToken }}
 metadata:
-  name: {{ include "trust-manager.name" . }}
+  name: {{ include "trust-manager.serviceAccountName" . }}
   namespace: {{ include "trust-manager.namespace" . }}
   labels:
     {{- include "trust-manager.labels" . | nindent 4 }}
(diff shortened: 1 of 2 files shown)


erikgb (commented):
Thanks @lexfrei! And a special thanks for adding tests around this!

/lgtm
/approve

PR #998, opened 2026-06-10
Make new trust package based on Debian Trixie the new default

erikgb:
Following up on https://github.com/cert-manager/trust-manager/pull/985, we should make the new trixie-based trust package image the new default.


SgtCoDFish (approved):
/lgtm
/approve

Can't see a reason not to do this!

PR #999, opened 2026-06-10
Trigger new releases of older trust-packages

erikgb:
Both are reported as vulnerable due to an outdated Go version (ref.). https://github.com/cert-manager/trust-manager/actions/runs/27282425445


SgtCoDFish (approved):
/lgtm
/approve

Simple enough, thanks!

Issue #886, opened 2026-02-14
Allow creating `ClusterRole` aggregations


brookelew:
Create [user-facing roles](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#user-facing-roles) with the

```
rbac.authorization.k8s.io/aggregate-to-admin: "true"
rbac.authorization.k8s.io/aggregate-to-edit: "true"
rbac.authorization.k8s.io/aggregate-to-view: "true"
```

authorizations, similar to what  `cert-manager` has.

/kind feature

PR #1009, opened 2026-06-16
feat(helm): aggregate Bundle read access into the cluster-reader ClusterRole

yugstar:
## What

Adds a user-facing aggregated ClusterRole for trust-manager `Bundle` resources, as requested in #886, mirroring cert-manager's precedent (`global.rbac.aggregateClusterRoles`).

When `global.rbac.create` is true and the new `global.rbac.aggregateClusterRoles` flag is true (default `true`), one ClusterRole is created:

- `trust-manager-cluster-view`: `get/list/watch` on `bundles`, aggregated to `cluster-reader`.

## Why read goes to cluster-reader (not view)

`Bundle` is cluster-scoped. cert-manager aggregates its own cluster-scoped resource (ClusterIssuers) read access into `cluster-reader` via `aggregate-to-cluster-reader`, rather than into the namespaced `view` role, so I followed that exact pattern. Because the resource is cluster-scoped, this grant takes effect for subjects bound via a ClusterRoleBinding to `cluster-reader`. A namespaced RoleBinding does not confer Bundle access.

## On write access

An earlier revision of this PR also aggregated write into `edit`/`admin`. Per review that has been dropped. A namespaced RoleBinding to `edit`/`admin` cannot grant access to a cluster-scoped resource, so aggregating write into those roles is the wrong tool while `Bundle` is cluster-scoped. Once a namespace-scoped Bundle exists (after the ClusterBundle migration), edit aggregation can be added on that resource. This PR therefore references #886 rather than closing it.

## Opt-in / safety

- Default on to match cert-manager and give the out-of-the-box UX #886 asks for.
- Set `global.rbac.aggregateClusterRoles: false` to opt out.

## Verification

- `helm template` with the flag on, off, and with `global.rbac.create=false`.
- New helm-unittest suite `tests/clusterrole_aggregate_test.yaml` passes.
- `values.schema.json` and `README.md` regenerated via `make generate-helm-schema generate-helm-docs`. `make verify-helm-values` and `verify-helm-lint` pass.

Refs #886

### DIFF
--- a/deploy/charts/trust-manager/README.md
+++ b/deploy/charts/trust-manager/README.md
@@ -15,6 +15,13 @@
 > ```
 
 Create required ClusterRoles, Roles, ClusterRoleBindings and RoleBindings for trust-manager.
+#### **global.rbac.aggregateClusterRoles** ~ `bool`
+> Default value:
+> ```yaml
+> true
+> ```
+
+Aggregate read access to Bundles into the "cluster-reader" ClusterRole, mirroring how cert-manager aggregates its cluster-scoped ClusterIssuers. Bundle is a cluster-scoped resource, so this only takes effect for subjects bound via a ClusterRoleBinding. For more information, see [User-facing roles](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#user-facing-roles).
 ### CRDs
 
 #### **crds.enabled** ~ `bool`
--- a/deploy/charts/trust-manager/templates/clusterrole-aggregate.yaml
+++ b/deploy/charts/trust-manager/templates/clusterrole-aggregate.yaml
@@ -1,0 +1,25 @@
+{{- if .Values.global.rbac.create }}
+{{- if .Values.global.rbac.aggregateClusterRoles }}
+# Bundle is a cluster-scoped resource, so its read access is aggregated into
+# the "cluster-reader" ClusterRole (mirroring how cert-manager aggregates its
+# cluster-scoped ClusterIssuers), rather than into the namespaced
+# view/edit/admin roles.
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  labels:
+    {{- include "trust-manager.labels" . | nindent 4 }}
+    rbac.authorization.k8s.io/aggregate-to-cluster-reader: "true"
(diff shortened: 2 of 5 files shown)


erikgb:
@yugstar, thanks for your PR. I don't think we should grant any write access to cluster-scoped resources through aggregated cluster roles. The Kubernetes aggregated user-facing roles are typically used in namespaces (in a role binding): https://kubernetes.io/docs/reference/access-authn-authz/rbac/#user-facing-roles.

After migrating Bundle to ClusterBundle, it will make room for a new namespace-scoped Bundle resource. When that is available, we can aggregate edit permissions to the new (namespace-scoped) Bundle.

yugstar:
Agreed on dropping the write aggregation. A namespaced RoleBinding cannot grant access to a cluster-scoped resource anyway.

For the read side I see two options. One is to keep a read-only ClusterRole aggregated into cluster-reader, the same way cert-manager does it for ClusterIssuer with its cluster-view role. The other is to hold this change until the namespace-scoped Bundle lands and then do read and edit together.

Which do you prefer?

erikgb:
Granting view permission to cluster-reader is fine. 😊

yugstar:
Thanks! Updated the PR to keep only the read-only cluster-reader role and dropped the write aggregation. I also changed it to Refs #886 instead of Fixes, so the issue stays open for the edit aggregation on the namespace-scoped Bundle later.

yugstar:
On the Copilot review: cluster-reader is OpenShift-specific as it notes, but switching to view, edit and admin would not fit a cluster-scoped resource like Bundle. cert-manager's own chart makes the same split: its cluster-scoped ClusterIssuers aggregate into cluster-reader only, while namespaced resources like Certificates and Issuers aggregate into view, edit, admin and cluster-reader (deploy/charts/cert-manager/templates/rbac.yaml). This PR follows that pattern and keeps Bundle read-only into cluster-reader, matching the direction you suggested.

I also dropped the stray "built-in" wording from the template comment and added test assertions for the rule's apiGroups, resources and rule scope.

Happy to adjust anything else you would like @erikgb, otherwise this should be ready.

erikgb (commented):
Thanks a lot!

/lgtm
/approve

PR #1016, opened 2026-06-21
Make CertPool support pluggable filters

erikgb:
This PR was initiated by a request from Khwaja Faraz on Slack: https://kubernetes.slack.com/archives/C4NV3DWUC/p1781970306601169.

It seems like he is using trust-manager as an SDK and custom filters for certificates in the CertPool. In this PR, I've made an attempt to make the filtering pluggable. I've made the change non-breaking, which is a big plus, as others might be importing trust-manager as a Go module.

### DIFF
--- a/pkg/util/cert_pool.go
+++ b/pkg/util/cert_pool.go
@@ -34,24 +34,42 @@ import (
 type CertPool struct {
 	certificates map[[32]byte]*x509.Certificate
 
-	filterExpired bool
-
-	filterNonCACerts bool
+	filters []CertPoolFilterFunc
 
 	logger logr.Logger
 }
+
+// CertPoolFilterFunc is invoked for each certificate before it's added to the pool.
+// Return true to accept the certificate or false to reject it.
+type CertPoolFilterFunc func(*x509.Certificate) bool
 
 type Option func(*CertPool)
 
 func WithFilteredExpiredCerts(filterExpired bool) Option {
 	return func(cp *CertPool) {
-		cp.filterExpired = filterExpired
+		if filterExpired {
+			cp.AppendFilter(func(cert *x509.Certificate) bool {
+				if time.Now().After(cert.NotAfter) {
+					cp.Logger().Info("ignoring expired certificate", "certificate", cert.Subject)
+					return false
+				}
+				return true
+			})
+		}
 	}
 }
 
 func WithFilteredNonCaCerts(filterNonCACerts bool) Option {
 	return func(cp *CertPool) {
-		cp.filterNonCACerts = filterNonCACerts
+		if filterNonCACerts {
+			cp.AppendFilter(func(cert *x509.Certificate) bool {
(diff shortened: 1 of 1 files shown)


farazkhawaja (approved):

farazkhawaja:
Thanks @erikgb

pkg/util/cert_pool.go
```diff
@@ -34,24 +34,42 @@ import (
...
 	logger logr.Logger
 }
 
+// CertPoolFilterFunc is invoked for each certificate before it's added to the pool.
+// Return true to accept the certificate or false to reject it.
+type CertPoolFilterFunc func(*x509.Certificate) bool
+
 type Option func(*CertPool)
 
 func WithFilteredExpiredCerts(filterExpired bool) Option {
```
SgtCoDFish, line 48: comment: huh, it didn't really occur to me before but having an option like this take a boolean really doesn't make much sense! There's literally no reason to ever write `WithFilteredExpiredCerts(false)`!

Probably not worth breaking to fix that, it's just a sharp edge.

pkg/util/cert_pool.go
```diff
@@ -34,24 +34,42 @@ import (
...
 	return func(cp *CertPool) {
-		cp.filterExpired = filterExpired
+		if filterExpired {
+			cp.AppendFilter(func(cert *x509.Certificate) bool {
```
SgtCoDFish, line 51: nitpick: This does allow someone to add the same filter multiple times, which the previous impl didn't allow. If someone adds a filter in a loop (mistakenly) the filter list could grow and consume a lot of memory.

We _could_ just ignore that for simplicity, I don't think it's a blocker

SgtCoDFish (commented):
/lgtm
/approve
/hold

Added a nitpick which can be ignored if you want.

(This review was fully human, no AI this time 👴 !)

pkg/util/cert_pool.go
erikgb: No, this is not right, but I suggest avoiding making this PR break for users. We can always fix it in a follow-up PR.

pkg/util/cert_pool.go
erikgb: Agreed that it would be dumb to add the same filter multiple times. But do you have any concrete suggestions on how to improve this, @SgtCoDFish?

SgtCoDFish: I suppose something simple is to keep `filterExpired` and `filterNonCACerts`, leaving the new filters slice just for "extra" filters. Exposing AppendFilter means there's always a risk of unbounded growth in the slice unless we limit the number somehow - but we can mitigate the risk for the common cases.

But this is definitely not a blocker - it's hardening and it's a very minor risk anyway. I'll unhold - no reason to keep this unmerged IMO

SgtCoDFish:
/unhold

No reason not to merge now!

Issue #837, opened 2026-01-12
Add startupapicheck for trust-manager


SgtCoDFish:
See https://github.com/cert-manager/approver-policy/issues/782 - basically this issue is to do that but for trust-manager.

Issue #742, opened 2025-09-09
Add option to disable webhook in Helm chart


quentin-besse-orange:
Hello,

**Is your feature request related to a problem? Please describe.**
Currently, in Helm we don’t have the possibility to completely disable the webhook.  
It seems that this option is missing.

 **Describe the solution you'd like**
 
Add a parameter in `values.yaml`:
 
 ```yaml
 webhook:
	 enabled: true
 ```

**Advantages**
- Provides more flexibility for users who don’t need the webhook.
- Simplifies deployment in environments where webhooks are restricted or not allowed.

/kind feature

erikgb:
@quentin-besse-orange, the long-term goal is to remove the (validating) webhook completely. But right now, it serves an important role in validating trust-manager resources. How do you expect trust-manager to work without important validation enabled? It is currently an integral part of trust-manager and can/should not be disabled.

quentin-besse-orange:
By disabling the webhook:
- There is no resource validation step
- The Trust Manager pod starts without any issues

What do you plan to use instead of the webhooks ?

erikgb:
> What do you plan to use instead of the webhooks ?

CEL expressions in CRDs: https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#validation-rules. We are already working on a migration from `Bundle` to `ClusterBundle`, which would also make the API more "self-validating".

> There is no resource validation step

This would not be acceptable, as it would require the controller to have a lot more advanced error-handling. Admission control is important for controllers.

quentin-besse-orange:
Do you have a roadmap available? I’d be happy to contribute.

quentin-besse-orange:
I wonder if the goal is to replace **all** webhooks with CEL.  
If so, I don’t see how some cases could be handled. Currently, the validations require comparing:

```go
type BundleSpec struct {
	Sources []BundleSource `json:"sources"`
	Target  BundleTarget  `json:"target,omitempty"`
}
```

based on their `name`, in order to check:

- “cannot define the same ConfigMap as both source and target”
- “cannot define the same Secret as both source and target”
    

I’d be interested to see how this could be achieved using CEL alone.

Otherwise, if we keep webhooks in addition to CEL: is it possible to enable them conditionally, similar to how [Crossplane](https://github.com/crossplane/crossplane/blob/main/cluster/charts/crossplane/values.yaml#L86) does?

erikgb:
[...]

PR #1045, opened 2026-07-25
feat(helm): make webhook failurePolicy configurable

### LINKED ISSUE #881: Helm install fails when extraObjects contains Bundles
I'm using the latest Helm chart v0.20.3. Using Helm version v3.17.1.

I've added 2 Bundle resources to the `extraObjects` field in the Helm values.yaml and when I try to deploy now, I get an error:
```
level=WARN msg="upgrade failed" name=trust-manager error="Internal error occurred: failed calling webhook \"trust.cert-manager.io\": failed to call webhook: Post \"https://trust-manager.mynamespace.svc:443/validate-trust-cert-man
ager-io-v1alpha1-bundle?timeout=5s\": dial tcp 10.233.23.225:443: connect: connection refused && Internal error occurred: failed calling webhook \"trust.cert-manager.io\": failed to call webhook: Post \"https://trust-mana
ger.mynamespace.svc:443/validate-trust-cert-manager-io-v1alpha1-bundle?timeout=5s\": dial tcp 10.233.23.225:443: connect: connection refused"
```

Even though there's an error, the trust-manager pod does come up healthy afterward. Running the Helm command again deploys the Bundles successfully.

I expected this to work because I do something similar in the cert-manager Helm chart, where I use the `extraObjects` block to deploy a `ClusterIssuer` and `Certificate` resource.

Note: I already deploy the Bundle CRD separately before my Helm install command so that's not an issue.

Here's my customized Helm values:
```
crds:
  enabled: false
  keep: true

namespace: "mynamespace"

nameOverride: "trust-manager"

app:
  trust:
    namespace: "mynamespace"

secretTargets:
  enabled: true
  # If set to true, grant read/write permission to all secrets across the cluster. Use with caution!
  # If set, ignores the authorizedSecrets list.
  authorizedSecretsAll: false
  # A list of secret names which trust-manager will be permitted to read and write across all namespaces.
  # These are the only allowable Secrets that can be used as targets. If the list is empty (and authorizedSecretsAll is false),
  # trust-manager can't write to secrets and can only read secrets in the trust namespace for use as sources.
  authorizedSecrets:
    - trust-bundle-all
    - custom-ca-secret


extraObjects:
    # Bundle that contains all certs
  - apiVersion: trust.cert-manager.io/v1alpha1
    kind: Bundle
    metadata:
      name: trust-bundle-all
    spec:
      sources:
      - useDefaultCAs: true
      - secret:
          name: "extra-trusted-certs"
          key: "certificates"
      - secret:
          name: "mongo-cert-root-ca"
          key: "tls.crt"
      - secret:
          name: "my-tls-secret"
          key: "tls.crt"
      - secret:
          name: "s3-tls-secret"
          key: "tls.crt"
      target:
        secret:
          key: "trust-bundle-all.crt"
          metadata:
            annotations:
              reloader.stakater.com/search: "true"
        additionalFormats:
          jks:
            key: "trust-bundle-all.jks"
        namespaceSelector:
          matchLabels: 
            my.org/main-namespace: 'true'
  - apiVersion: trust.cert-manager.io/v1alpha1
    kind: Bundle
    metadata:
      name: custom-ca-secret
    spec:
      sources:
      - useDefaultCAs: true
      - secret:
          name: "custom-ca-raw-certs"
          includeAllKeys: true
      target:
        secret:
          key: "ca.crt"
          metadata:
            annotations:
              reloader.stakater.com/search: "true"
        additionalFormats:
          jks:
            key: "keycloak-spi.truststore.jks"
```

jojinkb:
Fixes #881

## Problem

Deploying `Bundle`s via `extraObjects` in the same Helm release as trust-manager itself can fail the install/upgrade:

```
Internal error occurred: failed calling webhook "trust.cert-manager.io": failed to call webhook:
Post "https://trust-manager.mynamespace.svc:443/validate-trust-cert-manager-io-v1alpha1-bundle?timeout=5s":
dial tcp 10.233.23.225:443: connect: connection refused
```

Helm applies the `Bundle`s in the same manifest batch as the `Deployment` (no readiness ordering), so whenever the cluster-scoped `ValidatingWebhookConfiguration` already exists — every `helm upgrade`, or a re-install where it lingered — while the webhook `Service` has no ready endpoints, the `Bundle` CREATE is rejected and the whole release fails. Re-running the same command succeeds once the pod is ready, exactly as reported in #881.

Note this is *not* solved by the startupapicheck proposal (#837, PRs #900/#958): a post-install hook Job runs only **after** the main manifests — including `extraObjects` — have already been applied, so it can't sequence resources within the same release. Adding `helm.sh/hook` annotations to the `extraObjects` is also not a good workaround, since hook resources are not managed by the release (they leak on uninstall) and only sequence correctly with `--wait`.

## Solution

Expose the webhook failure policy as `app.webhook.failurePolicy` (default: `Fail` — no behavior change), instead of hardcoding `Fail` in the template.

Users who deploy `Bundle`s in the same release (e.g. via `extraObjects`) can set `Ignore`, allowing `Bundle`s to be admitted while the webhook is briefly unavailable:

- webhook validation still applies whenever the webhook is reachable (unlike disabling the webhook, which was rejected in #742);
- the CRD structural schema is always enforced by the API server;
- this is consistent with the long-term direction of moving validation into CEL and eventually removing the webhook (#742 discussion).

The `extraObjects` value documentation now also calls out the caveat and points at the new value.

If you'd rather address this differently (e.g. docs-only, or waiting for the ClusterBundle/CEL migration), happy to adjust.

## Changes

- `values.yaml`: new `app.webhook.failurePolicy` value (default `Fail`) + caveat note on `extraObjects`
- `templates/webhook.yaml`: render the value instead of the hardcoded `Fail`
- `tests/webhook_test.yaml`: new helm-unittest suite covering the default (`Fail`) and the override (`Ignore`); the override test fails without this change (schema rejects the unknown property)
- `values.schema.json` / `README.md`: regenerated with `make generate-helm-schema generate-helm-docs`

## Testing

- `make verify-helm-unittest` — 6 suites / 20 tests pass (new suite fails before the fix)
- `make verify-helm-values` — no errors
- `make verify-helm-lint` — 0 charts failed
- `make verify-helm-kubeconform` — passes
- `helm template` sanity checks: defaults render `failurePolicy: Fail` (unchanged); with `app.webhook.failurePolicy=Ignore` plus the reporter's `Bundle` `extraObjects`, the webhook renders `Ignore` and the `Bundle`s render as before

### DIFF
--- a/deploy/charts/trust-manager/README.md
+++ b/deploy/charts/trust-manager/README.md
@@ -572,6 +572,18 @@ Port that the webhook listens on.
 > ```
 
 Timeout of webhook HTTP request.
+#### **app.webhook.failurePolicy** ~ `string`
+> Default value:
+> ```yaml
+> Fail
+> ```
+
+The failure policy of the Bundle validating webhook. Accepted values are Fail and Ignore.  
+  
+Setting this to "Ignore" allows Bundles to be admitted while the webhook is unavailable, e.g. when Bundles are installed in the same Helm release as trust-manager itself (using  
+`extraObjects`) and the webhook endpoints aren't ready yet. Bundles admitted while the  
+webhook is unavailable skip webhook validation; the CRD schema is still enforced by the  
+Kubernetes API server.
 #### **app.webhook.service.type** ~ `string`
 > Default value:
 > ```yaml
@@ -779,7 +791,8 @@ NOTE: These annotations won't be added to the CRDs.
 > []
 > ```
 
-Extra manifests to be deployed. This is useful for deploying additional resources that are not part of the chart.  
+Extra manifests to be deployed. This is useful for deploying additional resources that are not part of the chart. NOTE: Bundles deployed here are validated by the trust-manager webhook, which may not be ready when the objects are applied, causing the Helm install or upgrade to fail. Consider setting  
+`app.webhook.failurePolicy` to "Ignore" or deploying Bundles separately from this chart.  
 For example:
 
 ```yaml
--- a/deploy/charts/trust-manager/templates/webhook.yaml
+++ b/deploy/charts/trust-manager/templates/webhook.yaml
@@ -114,7 +114,7 @@ webhooks:
           - bundles
     admissionReviewVersions: ["v1"]
     timeoutSeconds: {{ .Values.app.webhook.timeoutSeconds }}
-    failurePolicy: Fail
+    failurePolicy: {{ .Values.app.webhook.failurePolicy }}
     sideEffects: None
     clientConfig:
 {{ if .Values.app.webhook.tls.helmCert.enabled }}
--- a/deploy/charts/trust-manager/tests/webhook_test.yaml
+++ b/deploy/charts/trust-manager/tests/webhook_test.yaml
@@ -1,0 +1,28 @@
+suite: ValidatingWebhookConfiguration failure policy
+templates:
+  - templates/webhook.yaml
+release:
+  name: tm
+  namespace: trust
+tests:
+  - it: defaults to failurePolicy Fail
+    documentSelector:
+      path: kind
+      value: ValidatingWebhookConfiguration
+    asserts:
+      - equal:
+          path: webhooks[0].failurePolicy
+          value: Fail
+
+  - it: renders the configured failurePolicy
+    set:
+      app:
+        webhook:
+          failurePolicy: Ignore
+    documentSelector:
+      path: kind
+      value: ValidatingWebhookConfiguration
+    asserts:
+      - equal:
+          path: webhooks[0].failurePolicy
+          value: Ignore
--- a/deploy/charts/trust-manager/values.schema.json
+++ b/deploy/charts/trust-manager/values.schema.json
@@ -386,6 +386,9 @@
     "helm-values.app.webhook": {
       "additionalProperties": false,
       "properties": {
+        "failurePolicy": {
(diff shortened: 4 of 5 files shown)


erikgb:
Thanks for the PR, @jojinkb, but I need to discuss this with my co-maintainers. It never seems right to set the webhook `failurePolicy` to `Ignore`, as you risk getting faulty Bundle resources in your cluster. But at the same time, I have some understanding for your use case, even if I would personally NEVER use the `extraResources` value to provision trust-manager resources. Sadly, Helm is completely borked around CRD handling.

erikgb:
@jojinkb, this PR was discussed in today's cert-manager weekly development meeting, and we decided to reject the proposed change. There is no apparent reason to make the webhook failurePolicy configurable, as it is a footgun that could allow invalid resources to enter the cluster. The `extraObjects` value should not be used for CRs from CRDs in the chart. A similar change has already been rejected in cert-manager for the same reasons. The usual way to address your challenge is to orchestrate it at a higher level. Use FluxCD `dependsOn` or ArgoCD sync waves.

Sorry! Thanks for contributing to trust-manager!

/close

shastaxc:
Should this issue be closed for this reason? https://github.com/cert-manager/trust-manager/issues/881

erikgb:
> Should this issue be closed for this reason? #881

Good catch! Thanks, @shastaxc! I will close the issue with a similar comment.

PR #1078, opened 2026-08-26
Inject CA from Certificate Secret directly instead of via Certificate

zeusal:
Add support for injecting CA from Secret, allowing cainjector to run with --enable-certificates-data-source=false as recommended by best practices. https://cert-manager.io/docs/concepts/ca-injector/#controlling-which-data-sources-are-enabled.

Since there is already a conditional that determines whether to use the secret from the Helm chart or the secret generated by the `certificate` object, the approach in the latter case is to use that secret from the `certificate` object directly.

The annotations are added directly to the `values` because otherwise we would have to pre-buil the annotations from the `certificate.yaml` file template to allow for the annotation: cert-manager.io/allow-direct-injection: "true"

Discuss in MR: https://github.com/cert-manager/trust-manager/pull/762


zeusal:
@erikgb Can you review this MR?

deploy/charts/trust-manager/values.yaml
```diff
@@ -384,7 +384,9 @@ app:
       certificate:
         # Certificate duration, if unset/empty the default from cert-manager will be used. Example configuration to set 1 year: duration: 8766h
         duration: ""
-        secretTemplate: {}
+        secretTemplate:
+          annotations:
+            cert-manager.io/allow-direct-injection: "true"
```
erikgb: Maybe we should add this annotation to the secretTemplate inside the Certificate template to make it easier to use the secretTemplate Helm value? If a user has set this value, trust-manager will no longer work, as the Certificate secret will be missing the required annotation to inject directly from the secret.

zeusal: That's true, I am thinking two options:

Merge into secretTemplate:
`{{- $secretTemplate := dict }}
{{- $_ := set $secretTemplate "annotations" (dict "cert-manager.io/allow-direct-injection" "true") }}
{{- with .Values.app.webhook.tls.certificate.secretTemplate }}
{{- $secretTemplate = mergeOverwrite $secretTemplate . }}
{{- end }}
spec:
  secretName: {{ include "trust-manager.name" . }}-tls
  secretTemplate:
    {{- toYaml $secretTemplate | nindent 4 }}

Insert annotations:
`secretTemplate:
  annotations:
    cert-manager.io/allow-direct-injection: "true"
    {{- with .Values.app.webhook.tls.certificate.secretTemplate.annotations }}
    {{- toYaml . | nindent 4 }}
    {{- end }}
  {{- with .Values.app.webhook.tls.certificate.secretTemplate.labels }}
  labels:
    {{- toYaml . | nindent 4 }}
  {{- end }}

I prefer the first option because it is easier for the end user.

erikgb: I agree! Let's go for the merge approach!

deploy/charts/trust-manager/templates/webhook.yaml
```diff
@@ -94,7 +94,7 @@ metadata:
   {{- if or (not .Values.app.webhook.tls.helmCert.enabled) .Values.commonAnnotations }}
   annotations:
     {{- if not .Values.app.webhook.tls.helmCert.enabled }}
-    cert-manager.io/inject-ca-from: "{{ include "trust-manager.namespace" . }}/{{ include "trust-manager.name" . }}"
+    cert-manager.io/inject-ca-from-secret: "{{ include "trust-manager.namespace" . }}/{{ include "trust-manager.name" . }}-tls"
```
erikgb, line 101: Could you add a short (template) comment about WHY we are doing this?

deploy/charts/trust-manager/templates/certificate.yaml
```diff
@@ -15,6 +15,12 @@ spec:
   selfSigned: {}
 
 ---
+{{- $secretTemplate := dict }}
+{{- $_ := set $secretTemplate "annotations" (dict "cert-manager.io/allow-direct-injection" "true") }}
+
+{{- with .Values.app.webhook.tls.certificate.secretTemplate }}
+{{- $secretTemplate = mergeOverwrite $secretTemplate . }}
+{{- end }}
```
erikgb: Maybe also a short comment about WHY here?

erikgb (commented):
Some minor nits, but this look pretty close now.

deploy/charts/trust-manager/values.yaml
zeusal: I have applied this changes with helper block

deploy/charts/trust-manager/templates/webhook.yaml
zeusal: Added comment about why

deploy/charts/trust-manager/templates/certificate.yaml
zeusal: Added comment into helper function about why

deploy/charts/trust-manager/templates/_helpers.tpl
```diff
@@ -160,3 +160,18 @@ Namespaced resources rules
...
+can obtain its CA bundle from the TLS Secret regardless of whether the
+certificate is managed by cert-manager or generated by Helm.
+*/}}
+{{- define "trust-manager.webhook.secretTemplate" -}}
+{{- $secretTemplate := dict "annotations" (dict "cert-manager.io/allow-direct-injection" "true") -}}
+{{- with .Values.app.webhook.tls.certificate.secretTemplate -}}
+{{- $secretTemplate = mergeOverwrite $secretTemplate . -}}
+{{- end -}}
+{{- toYaml $secretTemplate -}}
+{{- end -}}
```
erikgb:
```suggestion
{{- end -}}

```

Nit: missing an empty line at the end of the file.

deploy/charts/trust-manager/templates/_helpers.tpl
```diff
@@ -160,3 +160,18 @@ Namespaced resources rules
...
 {{- end }}
 {{- end -}}
+
+{{/*
+Build webhook certificate secretTemplate.
+Ensures direct CA injection is always allowed.
+Add the CA injection annotations automatically so the webhook configuration
+can obtain its CA bundle from the TLS Secret regardless of whether the
+certificate is managed by cert-manager or generated by Helm.
+*/}}
```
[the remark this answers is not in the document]
erikgb: This simplification makes sense to me.

deploy/charts/trust-manager/templates/webhook.yaml
```diff
@@ -94,7 +94,11 @@ metadata:
   {{- if or (not .Values.app.webhook.tls.helmCert.enabled) .Values.commonAnnotations }}
   annotations:
     {{- if not .Values.app.webhook.tls.helmCert.enabled }}
-    cert-manager.io/inject-ca-from: "{{ include "trust-manager.namespace" . }}/{{ include "trust-manager.name" . }}"
+    {{- /*
+    Use the TLS Secret as the CA source so cainjector can populate the
+    webhook caBundle independently of how the certificate was created.
+    */}}
+    cert-manager.io/inject-ca-from-secret: "{{ include "trust-manager.namespace" . }}/{{ include "trust-manager.name" . }}-tls"
     {{- end }}
```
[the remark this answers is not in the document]
erikgb: This is a good suggestion, and we always need better testing. But not a blocker for me. WDYT, @zeusal?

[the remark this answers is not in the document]
zeusal: Give me a little time, because I am working on creating these tests.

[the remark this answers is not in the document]
zeusal: Tests added

zeusal:
@erikgb  You can check it again; I think everything's okay now.

deploy/charts/trust-manager/templates/_helpers.tpl
```diff
@@ -160,3 +160,18 @@ Namespaced resources rules
   {{- end }}
 {{- end }}
 {{- end -}}
+
+{{/*
+Build webhook certificate secretTemplate.
+Ensures direct CA injection is always allowed.
+Adds the CA injection annotation automatically so the webhook configuration
+can obtain its CA bundle from the TLS Secret when the webhook certificate is
+managed by cert-manager.
```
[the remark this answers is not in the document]
erikgb: I think the text works well as it is.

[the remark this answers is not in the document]
zeusal: I'm so sorry, I have accepted the change, but I can revert back fastly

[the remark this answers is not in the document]
erikgb: No worries! Just keep it as it is now.

erikgb (commented):
Thanks @zeusal! 🚀 I am planning a new release soonish. Stay tuned to verify that the features works in a new cluster.

/lgtm
/approve

PR #1070, opened 2026-08-21
Adopt the shared kube-api-linter makefile module

wallrj:
Replaces the local Kube API Linter setup added in #850 with the shared `kube-api-linter` module from cert-manager/makefile-modules#541 (now merged):

- `make/kube-api-lint.mk` and `.custom-gcl.yml` are deleted; `klone.yaml` klones the `kube-api-linter` module from makefile-modules `main` (all modules now at the same commit). The curated `.golangci-kal.yml` config is kept as `kube_api_linter_config`.
- No API changes: #1080 already bumped kube-api-linter to the July snapshot the module defaults to, and suppressed the 3 new `optionalfields` findings, so this is now a pure build-system change.

Verified locally after rebasing on #1080: `make verify-kube-api-lint` reports 0 issues.

/hold

*with claude fable-5*

PR #1080, opened 2026-08-28
Bump KAL to the latest pseudo-version

erikgb:
This PR is a preparation for https://github.com/cert-manager/trust-manager/pull/1070. Instead of fixing the new lint violations, I propose suppressing them. The new violations are reported on the soon-to-be-deprecated Bundle API, and changing a lot of fields to non-pointers will only create unnecessary noise and PR conflicts. We have also previously added suppressions for the same lint violations in the Bundle API.

The PR also contains a minor unrelated change to the .gitignore file. I can do this in a separate PR, but it would be easiest for me to include it here. 🌻

### DIFF
--- a/.custom-gcl.yml
+++ b/.custom-gcl.yml
@@ -2,4 +2,4 @@ version: v2.13.1
 name: golangci-kube-api-linter
 plugins:
 - module: 'sigs.k8s.io/kube-api-linter'
-  version: 'v0.0.0-20260114104534-18147eee9c49' # Pin to a commit while there's no tag
+  version: 'v0.0.0-20260716143926-092fe0c72997' # Pin to a commit while there's no tag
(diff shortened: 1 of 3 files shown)


wallrj:
Looks good. This matches the suppressions already on `main` (same `//nolint:kubeapilinter // We think pointers work best here` style on the v1alpha1 target fields), and suppressing rather than churning the soon-to-be-deprecated Bundle API makes sense. The pinned pseudo-version `v0.0.0-20260716143926-092fe0c72997` is the same snapshot the shared makefile-modules `kube-api-linter` module now defaults to, so #1070 can drop its fix commit and rebase cleanly on this.

Verify job is green, which also confirms the `nolint` directives work against the newer snapshot.

/approve
/lgtm
/label tide/merge-method-squash

*with claude fable-5*

PR #784, opened 2025-10-30
Rename Makefile debian trust package variables and files

inteon:
- standardize variable naming (always use bullseye or bookworm in variable names)
- remove old/ unused variables

PR #1085, opened 2026-09-01
Remove Debian Bullseye (EOL) trust bundle

erikgb:
Debian Bullseye has reached EOL, and I don't think we will publish any more updates to our trust bundle based on Debian Bullseye. https://www.debian.org/News/2026/20260831

I want to do this now, as we are about to prepare a new release of trust-manager and need to bump the default trust bundle images, which are reported as vulnerable by scanners.


wallrj:
I reviewed this at 2f098e0. The removal looks complete and consistent: `git grep -i bullseye` on the PR head returns nothing, and no remaining file consumes the deleted `RELEASE_OCI_PACKAGE_DEBIAN_BULLSEYE_*` outputs or `00_debian_bullseye_version.mk` variables. LGTM.

While verifying that, I found three pre-existing leftovers of the bullseye era that this PR brushes against but doesn't finish. None of them blocks this PR, but since this is the mopping-up PR they may be worth folding in (or noting for a follow-up):

1. **`release.yaml` consumes outputs that nothing emits.** [`.github/workflows/release.yaml` lines 42–43](https://github.com/cert-manager/trust-manager/blob/2f098e06275b61f62c7456ccc9df385444085543/.github/workflows/release.yaml#L42-L43) (and the release-notes lines [61–62](https://github.com/cert-manager/trust-manager/blob/2f098e06275b61f62c7456ccc9df385444085543/.github/workflows/release.yaml#L61-L62)) still read `RELEASE_OCI_PACKAGE_DEBIAN_IMAGE`/`_TAG`, but the `release` target only emits `_BOOKWORM_`/`_TRIXIE_` suffixed names ([`make/02_mod.mk` lines 53–58](https://github.com/cert-manager/trust-manager/blob/2f098e06275b61f62c7456ccc9df385444085543/make/02_mod.mk#L53-L58)). On the next tagged release those lines in the generated notes will be blank, and the bookworm/trixie image+tag that *are* emitted never appear in the notes. This was already dangling before this PR (the merge-base only emits suffixed names too), but with bullseye gone it's permanently dead.

2. **Duplicate trixie prerequisite in the smoke test.** [`make/test-smoke.mk` line 44](https://github.com/cert-manager/trust-manager/blob/2f098e06275b61f62c7456ccc9df385444085543/make/test-smoke.mk#L44) lists `oci-load-package_debian_trixie` twice — #1000 (b76deca) replaced the bullseye entry with a second trixie entry instead of deleting it. Harmless (make dedups prerequisites), but confusing.

3. **`RELEASE.md` names a workflow that doesn't exist.** [`RELEASE.md` line 60](https://github.com/cert-manager/trust-manager/blob/2f098e06275b61f62c7456ccc9df385444085543/RELEASE.md#L60) points at `.github/workflows/trust-package-upgrade-debian-<version>.yaml`, but upgrades run from the single matrixed [`trust-package-upgrade-debian.yaml`](https://github.com/cert-manager/trust-manager/blob/2f098e06275b61f62c7456ccc9df385444085543/.github/workflows/trust-package-upgrade-debian.yaml). (The per-version `trust-package-release-debian-<version>.yaml` files it also mentions do exist.)

*with claude fable-5*

wallrj:
/lgtm
/approve
/hold

Holding in case you want to fold in the suggestions above, @erikgb — this is good to merge as-is, so feel free to `/unhold` if you'd rather leave them for a follow-up.

*with claude fable-5*

erikgb:
Thanks for the extended review, @wallrj. I should have fixed all of this now. PTAL!

wallrj:
/lgtm
/approve
/unhold

Thanks @erikgb — I checked the follow-up at 1fc21d0 and all three points are addressed.

On removing the `RELEASE_OCI_PACKAGE_DEBIAN_*` outputs rather than replacing them with the bookworm/trixie variants: I did the archaeology and deletion is right. Those lines only fed the draft release notes, and they have printed blank values since v0.20.3, when #784 renamed the make outputs to suffixed names without updating `release.yaml`. The published notes never mention the trust package image, and the chart records the authoritative default (`defaultPackageImage.name: trust-pkg-debian-trixie`, tag baked from `make/00_debian_trixie_version.mk` at chart build time), so nothing that works today is lost.

*with claude fable-5*

PR #1086, opened 2026-09-01
Bump default CAs bundle version to trigger release

erikgb:
All our default bundle images are reported as vulnerable by vulnerability scanners. This PR is the same drill as in https://github.com/cert-manager/trust-manager/pull/768 (example) to trigger a new release of the bundle images. This is done now in preparation for a new trust-manager release.

### DIFF
--- a/make/00_debian_bookworm_version.mk
+++ b/make/00_debian_bookworm_version.mk
@@ -17,5 +17,5 @@
 # This file is used to store the latest version of the debian trust package and the DEBIAN_BOOKWORM_BUNDLE_VERSION
 # variable is automatically updated by the `upgrade-debian-bookworm-trust-package-version` target and cron GH action.
 
-DEBIAN_BOOKWORM_BUNDLE_VERSION := 20250419~deb12u1.0
+DEBIAN_BOOKWORM_BUNDLE_VERSION := 20250419~deb12u1.1
 DEBIAN_BOOKWORM_BUNDLE_SOURCE_IMAGE=docker.io/library/debian:12-slim
--- a/make/00_debian_trixie_version.mk
+++ b/make/00_debian_trixie_version.mk
@@ -17,5 +17,5 @@
 # This file is used to store the latest version of the debian trust package and the DEBIAN_TRIXIE_BUNDLE_VERSION
 # variable is automatically updated by the `upgrade-debian-trixie-trust-package-version` target and cron GH action.
 
-DEBIAN_TRIXIE_BUNDLE_VERSION := 20250419.1
+DEBIAN_TRIXIE_BUNDLE_VERSION := 20250419.2
 DEBIAN_TRIXIE_BUNDLE_SOURCE_IMAGE=docker.io/library/debian:13-slim


wallrj (approved):
The version bumps themselves are correct. I checked:

- the latest published tags are `20250419-deb12u1.0` ([trust-pkg-debian-bookworm](https://quay.io/repository/jetstack/trust-pkg-debian-bookworm?tab=tags)) and `20250419.1` ([trust-pkg-debian-trixie](https://quay.io/repository/jetstack/trust-pkg-debian-trixie?tab=tags)), so `.1` and `.2` are the right next suffixes and neither tag exists yet, meaning `oci-maybe-push` will really build and push;
- the base ca-certificates versions (`20250419~deb12u1` and `20250419`) are still current in bookworm-security and trixie, so the exact-mode `apt-get install` in the release build will succeed;
- the daily cron runs in `latest` mode and reuses the current suffix when the base version is unchanged, so it won't revert this bump.

However, while tracing the bump through the release machinery I found a problem: **the new suffix ends up in the image tag, but not in the contents of the image**.

The release build runs [`debian-trust-package-fetch.sh` in `exact` mode](https://github.com/cert-manager/trust-manager/blob/10a35e3bac14617788b91956738a44efbd2a6654/make/debian-trust-package.mk#L30-L32). The version that gets written into `cert-manager-package-debian.json` is `installed_version` plus `version_suffix`, and [`version_suffix` is only copied from the target version when `ACTION == "latest"`](https://github.com/cert-manager/trust-manager/blob/10a35e3bac14617788b91956738a44efbd2a6654/make/debian-trust-package-fetch.sh#L136-L142) — in `exact` mode it is always `.0`. Meanwhile the [image tag comes straight from the `.mk` variable](https://github.com/cert-manager/trust-manager/blob/10a35e3bac14617788b91956738a44efbd2a6654/make/00_mod.mk#L52) this PR bumps. So the two disagree:

```mermaid
flowchart TD
    A["make/00_debian_trixie_version.mk<br/>DEBIAN_TRIXIE_BUNDLE_VERSION := 20250419.2"] --> B["image tag<br/>trust-pkg-debian-trixie:20250419.2"]
    A --> C["debian-trust-package-fetch.sh exact<br/>apt-get install ca-certificates=20250419"]
    C --> D["version_suffix stays '.0'<br/>only reused when ACTION == latest"]
    D --> E["embedded JSON<br/>version: 20250419.0"]
    B --> F["one image, two different versions"]
    E --> F
    F --> G["Bundle status.defaultCAPackageVersion<br/>cert-manager-debian-trixie-20250419.0-hash<br/>identical before and after the upgrade"]
```

This is not hypothetical — the image published from #768 already shows it:

```console
$ crane export quay.io/jetstack/trust-pkg-debian-trixie:20250419.1 - \
    | tar -xO --wildcards '*cert-manager-package-debian.json' | jq .version
"20250419.0"
```

Why it matters: the embedded version is the one users can see. [`Package.Version` exists "to distinguish updated bundles from older counterparts"](https://github.com/cert-manager/trust-manager/blob/10a35e3bac14617788b91956738a44efbd2a6654/pkg/fspkg/package.go#L43-L44) and feeds [`Package.StringID`](https://github.com/cert-manager/trust-manager/blob/10a35e3bac14617788b91956738a44efbd2a6654/pkg/fspkg/package.go#L47-L51), which is reported as `defaultCAPackageVersion` in Bundle status. After an operator rolls out the rebuilt `.2` image to pick up the fixed Go binaries, their Bundle status will still say `...20250419.0-<hash>`, so there is no way to confirm from the API that the remediation actually landed — the whole point of a suffix bump.

The fix looks small: in `exact` mode apt is told to install exactly the target base version, so `installed_version` is guaranteed to equal it, and the script can safely reuse the target suffix there too — for example, drop the `ACTION == "latest"` half of the condition on [line 139](https://github.com/cert-manager/trust-manager/blob/10a35e3bac14617788b91956738a44efbd2a6654/make/debian-trust-package-fetch.sh#L139).

Separately: this is the second manual suffix-bump PR (after #768), and there were six more on the old bookworm base (up to `20230311-deb12u1.7`). The nightly `trust-package-security-scan` workflow already detects exactly the condition that triggers these PRs, and the `upgrade-debian-*-trust-package-version` targets already know how to edit the version files and open a PR. Wiring the scan failure to bump the suffix and open the PR automatically would close the vulnerability window without waiting for someone to notice a red scan. Happy to raise issues or PRs for either of these.

Approving because the bump itself is right and gets the rebuilt images out; holding in case you want to discuss or address the above first.

/hold

*with claude fable-5*

wallrj:
I've opened #1087 to fix the embedded-version problem described in my review above.

Note that the merge order matters: `oci-maybe-push` only builds and pushes a tag that doesn't already exist, so if this PR merges first, the `.1`/`.2` images will be published with `"version": "...0"` embedded and won't be rebuilt when #1087 lands. Merging #1087 first means the images released by this PR get the correct embedded versions.

*with claude fable-5*

wallrj:
#1087 has merged, so the release builds triggered by this PR will now embed the bumped suffix in the package (they run from `main`, so no rebase needed). Lifting the hold.

/unhold
/lgtm

*with claude fable-5*

wallrj:
Confirming this worked end to end now that #1087 merged first. Both release workflows succeeded ([trixie](https://github.com/cert-manager/trust-manager/actions/runs/33718461000), [bookworm](https://github.com/cert-manager/trust-manager/actions/runs/33718460997)) and the published images embed the bumped suffix:

```console
$ crane export quay.io/jetstack/trust-pkg-debian-trixie:20250419.2 - \
    | tar -xO --wildcards '*cert-manager-package-debian.json' | jq .version
"20250419.2"
$ crane export quay.io/jetstack/trust-pkg-debian-bookworm:20250419-deb12u1.1 - \
    | tar -xO --wildcards '*cert-manager-package-debian.json' | jq .version
"20250419~deb12u1.1"
```

(The previous `trust-pkg-debian-trixie:20250419.1` embedded `"20250419.0"`.)

*with claude fable-5*

PR #1100, opened 2026-09-11
Don't run trust-package-upgrade-debian in forks

erikgb:
After resyncing my fork of trust-manager, I get daily emails about failing jobs. This PR restricts the upgrade workflow to run only in the upstream repo, using a pattern already in use in this repo.


SgtCoDFish (approved):
/lgtm
/approve

Makes sense to me!

PR #1110, opened 2026-09-19
chart: document that authorizedSecrets cannot grant Secret creation

kta1kri:
## What

Corrects the chart documentation for `secretTargets`. No rendered RBAC changes — `_helpers.tpl` is
identical to `main` apart from a Helm comment that renders to nothing.

## Why

Two statements in the chart are inaccurate today.

**1. `authorizedSecrets` does not bound writing; it does not permit creation at all.**

The rule emitted for the name-list path is:

```yaml
- apiGroups: [""]
  resources: ["secrets"]
  verbs: ["create","patch","delete"]
  resourceNames: {{ .Values.secretTargets.authorizedSecrets }}
```

Kubernetes RBAC ignores `resourceNames` for `create`, because the API server does not know the
object's name at authorization time
([RBAC reference](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#referring-to-resources)).
The `create` entry therefore never matches, and trust-manager can only patch and delete Secrets
that already exist under one of the listed names. The current description — "permitted to read and
write" / "the only allowable Secrets that can be used as targets" — reads as if creation were both
permitted and name-bounded; neither is true.

**2. `secretTargets.enabled` on its own grants nothing.**

The Secret rules are inside `{{- if .Values.secretTargets.authorizedSecretsAll }} … {{- else if
$.Values.secretTargets.authorizedSecrets }}`, so with the defaults (`authorizedSecretsAll: false`,
`authorizedSecrets: []`) no Secret rule is emitted. The read-all warning belongs on the two values
that actually trigger it.

## Note on the earlier revision

The first version of this branch split the rule so `create` sat in its own entry without
`resourceNames`. As @Copilot pointed out, that expands the effective grant to "create a Secret
under any name" rather than documenting the existing one, so it has been reverted.

## Open question for maintainers

If an operator sets `authorizedSecrets` and the target Secret does not exist yet, trust-manager
cannot create it. Is that the intended contract, or should the chart grant `create` and accept
that RBAC cannot bound it by name? Happy to follow up either way.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


kta1kri:
Thanks for the review — the first point is right and I have reverted it.

Splitting the rule so that `create` sat outside `resourceNames` did expand the effective grant to
"create a Secret under any name", which is not something a documentation PR should do silently.
`templates/_helpers.tpl` is now byte-identical to `main` apart from a `{{- /* */ -}}` comment,
so no rendered RBAC changes.

What is left is the wording, which I have made narrower:

- `secretTargets.authorizedSecrets`: patch and delete are restricted to the listed names; **create
  is not granted at all on this path**, because Kubernetes ignores `resourceNames` for `create`, so
  that entry never matches. A target Secret has to exist already, or `authorizedSecretsAll` is
  needed — which grants create for every name.
- `secretTargets.enabled`: qualified per your second point. On its own it emits no Secret rule, so
  the read-all grant only appears once `authorizedSecrets` or `authorizedSecretsAll` is set.

That leaves a real question I did not want to answer on your behalf: if an operator sets
`authorizedSecrets` and the target Secret does not exist yet, trust-manager cannot create it. Is
that the intended contract (targets must be pre-created), or should the chart grant `create` and
accept that RBAC cannot bound it by name? I am happy to follow up with whichever you prefer.

PR #1111, opened 2026-09-20
fix: merge commonLabels as a map to avoid duplicate label keys

pgmpofu:
## Summary

`.Values.commonLabels` is concatenated onto the chart's own labels as raw YAML text (`{{ toYaml .Values.commonLabels }}`) rather than merged, so any key that collides with a chart-owned key produces two mapping entries for the same key and breaks rendering:

```
yaml: unmarshal errors:
  line 14: mapping key "app.kubernetes.io/managed-by" already defined at line 13
```

Setting any of `app`, `app.kubernetes.io/name`, `app.kubernetes.io/instance`, `app.kubernetes.io/version`, `app.kubernetes.io/managed-by` or `helm.sh/chart` via `commonLabels` currently breaks the render. `app.podLabels` and `servicemonitor.labels` have the same defect, since they are concatenated after the helper.

This is the trust-manager half of cert-manager/cert-manager#9348. The same fix landed in google-cas-issuer as cert-manager/google-cas-issuer#562, and this keeps the semantics identical so `commonLabels` behaves the same in both charts.

## Fix

Labels are merged as a map, so a key can only be emitted once. **On conflict `commonLabels` wins**, which matches the de facto behaviour of the concatenated form (`commonLabels` was emitted last, so it already won wherever the decoder was lenient rather than erroring).

`app` is deliberately kept out of the shared helper, because whether it may be overridden depends on whether a selector reads it:

| Resource | `app` | Why |
| --- | --- | --- |
| Pod template | forced | The Deployment, PodDisruptionBudget, metrics Service and webhook Service selectors all match on it |
| Metrics Service | forced, plus `app.kubernetes.io/component` | The ServiceMonitor selects the Service on **both**; overriding either would silently stop metrics being scraped |
| Everything else | overridable | Merged at lower precedence, exactly as `commonLabels` could override it before |

`mergeOverwrite` is used rather than `merge` because `merge` skips empty source values, so an intentionally empty `podLabels` value would lose to `commonLabels`.

With default values the rendered label set is unchanged — only key ordering differs, since `toYaml` sorts.

## Test plan

- [x] `make verify-helm-unittest` — 29 pass (22 existing + 7 new in `tests/labels_test.yaml`). The new tests fail against the unfixed templates with the `mapping key already defined` error above.
- [x] `make verify-helm-values` (`helm-tool lint`), `verify-helm-lint`, `verify-helm-kubeconform`, `verify-pod-security-standards`, `verify-kube-api-lint` — all pass
- [x] Rendered the chart with PDB and ServiceMonitor enabled across 8 scenarios (`commonLabels.app`, chart-key overrides, component hijack, `podLabels.app`, podLabels/commonLabels collisions, YAML 1.1 boolean-ish keys such as `on`/`no`, `servicemonitor.labels` collisions), asserting no duplicate keys and that **every selector still matches its target's labels**. 6 of the 8 fail on `main`.
- [x] Default-values label sets are identical to `main`

🤖 Generated with [Claude Code](https://claude.com/claude-code)


pgmpofu:
Some verification evidence for reviewers, run against this branch with `helm template` and a strict YAML decoder (one that rejects duplicate mapping keys, like the client Helm and the API server use). All of it is reproducible from the chart directory.

### 1. The bug, and that it is fixed

On `main`, a single colliding key corrupts multiple documents. Rendering with `--set app.metrics.service.servicemonitor.enabled=true --set podDisruptionBudget.enabled=true` and then strict-decoding:

| `--set` | documents that fail to decode on `main` | on this branch |
| --- | --- | --- |
| `commonLabels.app=x` | 6 — Deployment, both Services, PodDisruptionBudget, ValidatingWebhookConfiguration, ServiceMonitor | 0 |
| `commonLabels."app\.kubernetes\.io/component"=x` | 1 — metrics Service | 0 |
| `commonLabels.prometheus=x` | 1 — ServiceMonitor | 0 |
| `app.podLabels.app=x` | 1 — Deployment | 0 |

The failure is `yaml: unmarshal errors: mapping key "app" already defined`, i.e. the same class of error as the original report.

### 2. No behaviour change with default values

Rather than diffing the rendered text (which changes, because `toYaml` sorts keys and quotes differently), I decoded both renders and compared the label map and every selector per resource:

```
17 resources, 0 differ
```

Same result for `--set commonLabels.team=platform`, `--set app.podLabels.extra=1`, and `--set app.metrics.service.servicemonitor.labels.foo=bar`.

### 3. Selectors still match under hostile values

This is the part worth a careful look, since the chart has five selectors that read labels this PR touches. For defaults and for each of `commonLabels.app=hijack`, `commonLabels.prometheus=hijack`, `app.podLabels.app=hijack`:

```
Deployment selector      -> pod template : matches
Service/trust-manager    -> pod template : matches
Service/..-metrics       -> pod template : matches
PodDisruptionBudget      -> pod template : matches
ServiceMonitor           -> Service/..-metrics : matches
ServiceMonitor "prometheus" label          : default (not hijacked)
```

### 4. Tests

`make verify-helm-unittest` passes — 8 suites, 29 tests, including the 7 suites that already existed. Copying `tests/labels_test.yaml` onto unmodified `main` gives 7 failed / 7 errored / 2 passed, with real `mapping key ... already defined` errors, so the new tests are genuine regression guards rather than tests that pass either way.

`verify-helm-values`, `verify-helm-lint` and `verify-helm-kubeconform` also pass.

### One deliberate detail worth confirming

On `main` the ServiceMonitor emitted `prometheus:` and then `servicemonitor.labels`, so with last-wins decoding `servicemonitor.labels` could override the `prometheus` label. This branch preserves that ordering. It is easy to get backwards when converting the concatenation to a merge, and getting it backwards would silently stop a Prometheus instance selecting the ServiceMonitor for anyone relying on it — so flagging it as intentional rather than accidental.

This is a port of the approach merged in cert-manager/google-cas-issuer#562, tracked for the remaining charts in cert-manager/cert-manager#9348.

pgmpofu:
Pushed 4118f8f — I found one resource this PR had missed.

`clusterrole-aggregate.yaml` emits `rbac.authorization.k8s.io/aggregate-to-cluster-reader` immediately after the labels helper, so it still had the same duplicate-key defect:

```
$ helm template tm deploy/charts/trust-manager \
    --set global.rbac.aggregateClusterRoles=true \
    --set 'commonLabels.rbac\.authorization\.k8s\.io/aggregate-to-cluster-reader=x'
# ClusterRole/trust-manager-cluster-view:
#   duplicate key 'rbac.authorization.k8s.io/aggregate-to-cluster-reader'
```

It is worth treating as an override rather than a default: the built-in `cluster-reader` ClusterRole aggregates on that label, so allowing `commonLabels` to change it would silently drop Bundle read access from that role rather than fail loudly. Other `commonLabels` keys still apply to the resource normally, and there is a test for each half of that.

I also added the two missing entries to the `commonLabels` documentation — the aggregation label and the ServiceMonitor's `prometheus` label, which this PR already protects but did not mention.

### How I found it

Rather than reading the templates again, I enumerated every label key the chart emits by default, then rendered once per key with that key set in `commonLabels` and strict-decoded the result:

```
app                                                     ok
app.kubernetes.io/component                             ok
app.kubernetes.io/instance                              ok
app.kubernetes.io/managed-by                            ok
app.kubernetes.io/name                                  ok
app.kubernetes.io/version                               ok
helm.sh/chart                                           ok
prometheus                                              ok
rbac.authorization.k8s.io/aggregate-to-cluster-reader   BROKEN   <-- fixed by 4118f8f
```

After the fix, all of them pass. That sweep is mechanical, so it should be a reliable check for the remaining charts in cert-manager/cert-manager#9348 too — the aggregation label is exactly the kind of thing that is easy to miss by grepping for `app:`.

### Re-verified after the change

- All 13 emitted label keys: no duplicate mapping keys.
- Default render still matches `main` exactly — 17 resources, 0 differences in labels or selectors, with the ServiceMonitor, PodDisruptionBudget and aggregated ClusterRole all enabled.
- The new tests fail against the previous commit on this branch (`duplicate key ... already defined`) and pass after it.
- `verify-helm-values`, `verify-helm-unittest` (8 suites, 31 tests), `verify-helm-lint`, `verify-helm-kubeconform` all pass; regenerating the README and schema produces no further diff.

